PR #30, opened 2020-06-18
Ability to define events as produced/consumed

tsurdilo:

- [x] Specification
- [x] Examples
- [x] Schema
- [x] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] Website
- [ ] SDK
- [ ] TCK
- [ ] Other

This PR adresses Issue: https://github.com/cncf/wg-serverless-workflow/issues/29

It allows users to define an event definition as "produced" or "consumed" with a new "kind" property of the event definition.

If kind is "produce" the "source" property of the event definition is not required. Implementations
must set this property of the CE when the event is produced.

### DIFF
--- a/specification/README.md
+++ b/specification/README.md
@@ -268,9 +268,10 @@ Since function definitions are reusable, their data input parameters are defined
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
 | name | Unique event name | string | yes |
-| source | CloudEvent source | string | yes |
+| source | CloudEvent source | string | yes if kind is set to "consume", otherwise no |
 | type | CloudEvent type | string | yes |
 | correlationToken | Context attribute name of the CloudEvent which value is to be used for event correlation | string | no |
+| kind | Defines the events as either being consumed or produced by the workflow. Default is consumed. | enum | no |
 | [metadata](#Workflow-Metadata) | Metadata information | object | no |
 
 <details><summary><strong>Click to view example definition</strong></summary>
@@ -289,7 +290,8 @@ Since function definitions are reusable, their data input parameters are defined
    "name": "ApplicationSubmitted",
    "type": "org.application.submit",
    "source": "applicationssource",
-   "correlationToken": "applicantId"
+   "correlationToken": "applicantId",
+   "kind": "consumed"
 }
 ```
 
@@ -301,6 +303,7 @@ name: ApplicationSubmitted
 type: org.application.submit
 source: applicationssource
 correlationToken: applicantId
+kind: consumed
 ```
 
 </td>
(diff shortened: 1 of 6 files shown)


tsurdilo:
@manuelstein can you please take a look? Thanks.

tsurdilo:
fixed typos
consume->consumed
produce->produced

specification/README.md
```diff
@@ -366,28 +367,60 @@ Since function definitions are reusable, their data input parameters are defined
...
 </details>
 
 Defines events that can be consumed or produced during workflow execution.
-Consumed events can trigger actions to be executed. Events can also be produced during workflow
-execution to be consumed by clients.
 
-As serverless workflow definitions are vendor neutral, so should be the events definitions that they consume and produce.
-As such event format within serverless workflows uses the [CloudEvents](https://github.com/cloudevents/spec) specification to describe events.
+Consumed events can trigger creations of workflow instances, or continue execution of existing workflow instances.
+Workflows can produce events during their execution. These events can be then consumed by clients (or other workflows).
```
manuelstein, line 410: Please clarify if an eventdef can serve for both consumed and produced event types or if that case requires a second eventdef.

specification/schema/serverless-workflow-schema.json
```diff
@@ -182,6 +182,12 @@
           "type": "string",
           "description": "CloudEvent type"
         },
+        "kind": {
+          "type" : "string",
+          "enum": ["consumed", "produced"],
```
manuelstein, line 187: I was wondering - if it's a binary choice (either consumed or produced), then the presence or absence of the source field alone could determine the kind (that would require a documented specification but would not require the addition of a 'kind' field). Unless the kind can be both, in which case the source field is required yet it is also an event that could be produced.

tsurdilo:
@manuelstein yes i think the cases where the same event type is produced and consumed might be very small in which case defining them separately (one consumed and one produced) would be ok.

I understand your point regarding absence of the source param to determine produced / consumed. It would certainly make the model shorter, but IMO less deterministic, as it could clash with user mistakes during modeling for example. For example if I am to write lets say a visualization tool or a linter for this scenario I would not know if user forgot to add the source or its on purpose in order to make it a "produced" one. WDYT?

specification/README.md
tsurdilo: as defined currently it cannot. and as you mentioned i think the use cases where this happens seem very small in which case i think it would be ok. I will update the spec doc to mention this clearly.

specification/schema/serverless-workflow-schema.json
tsurdilo: added comment on this below.

tsurdilo:
rebased

tsurdilo:
rebased

tsurdilo:
rebased

tsurdilo:
@manuelstein mind reviewing again after latest comments? thanks.

tsurdilo:
rebased

manuelstein:
> @manuelstein yes i think the cases where the same event type is produced and consumed might be very small in which case defining them separately (one consumed and one produced) would be ok.
> 
> I understand your point regarding absence of the source param to determine produced / consumed. It would certainly make the model shorter, but IMO less deterministic, as it could clash with user mistakes during modeling for example. For example if I am to write lets say a visualization tool or a linter for this scenario I would not know if user forgot to add the source or its on purpose in order to make it a "produced" one. WDYT?

Sorry, just added my thumbs up to your comment. Fair enough, so LGTM

PR #34, opened 2020-06-21
Adding cron-based scheduled start definition

tsurdilo:

- [x] Specification
- [x] Examples
- [x] Schema
- [x] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] Website
- [ ] SDK
- [ ] TCK
- [ ] Other

Addresses issue https://github.com/cncf/wg-serverless-workflow/issues/31

Adds cron-based scheduled start definition


tsurdilo:
@singhkays Hi, can you please take a look? Thanks!

specification/schema/serverless-workflow-schema.json
```diff
@@ -1575,33 +1575,66 @@
...
+            "triggered"
           ],
           "description": "Kind of start definition"
         },
         "schedule": {
           "description": "If kind is scheduled, define when the starting state is or becomes active",
           "$ref": "#/definitions/schedule"
+        },
+        "trigger": {
+          "description": "If kind is 'triggered'', define the repeating intervals at which to start workflow instances",
```
ricardozanini, line 1588:
```suggestion
          "description": "If kind is 'triggered', define the repeating intervals at which to start workflow instances",
```

tsurdilo:
@ricardozanini thanks for review!!!!! ill update asap!

specification/README.md
```diff
@@ -2852,8 +2852,9 @@ If the defined callback event has not been received during this time period, the
 
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
-| kind | End kind ("default", "scheduled") | enum | yes |
+| kind | End kind ("default", "scheduled", "triggered") | enum | yes |
```
singhkays, line 2855: Triggered sounds to me like a manual invocation more than "scheduled"/cron :) Left a comment suggesting using "recurring" or overloading "scheduled". I think historically "cron" and "scheduled" have been used interchangeably which might make them separating here confusing. 

An example question I think users might ask - _"Scheduled doesn't mean cron in serverless workflow?"_

specification/README.md
```diff
@@ -2916,7 +2952,19 @@ There are two cases to discuss when dealing with scheduled start states:
...
+Defining a trigger for the start definition allows you to create process instances periodically based on defined repeating intervals.
+Trigger definitions can handle absolute time intervals (not calculated in respect to some particular point in time).
+For example let's say we have a workflow which performs data batch processing which has to be done periodically. In this case we could use a cron definition
+
+``` text
+0 0/5 * * * ?
+```
+
+to define that this workflow should be triggered every 5 minutes, starting at full hour. 
+[See here](#http://www.quartz-scheduler.org/documentation/quartz-2.3.0/tutorials/tutorial-lesson-06.html) to get more information on defining cron expressions.
```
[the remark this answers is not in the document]
singhkays: +1

specification/README.md
tsurdilo: Good point. Let me see if we can do the overload as you mentioned, having either interval or cron property under "scheduled" obj def. WDYT?

tsurdilo:
Thanks for the review @singhkays !

specification/README.md
```diff
@@ -2866,46 +2867,81 @@ If the defined callback event has not been received during this time period, the
...
       ],
       "description": "Kind of start definition"
     },
     "schedule": {
-      "description": "If kind is scheduled, define when the starting state is or becomes active",
+      "description": "If kind is 'scheduled', define when the starting state is or becomes active",
       "$ref": "#/definitions/schedule"
+    },
+    "trigger": {
+      "description": "If kind is 'triggered'', define the repeating intervals at which to start workflow instances",
```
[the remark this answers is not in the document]
tsurdilo: fixed

specification/README.md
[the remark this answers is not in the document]
tsurdilo: fixed

specification/README.md
```diff
@@ -2962,6 +3010,38 @@ a specific interval, the "waiting" for event "X" should only be started when the
...
+  "required": [
+    "interval"
+  ]
+}
+```
+
+</details>
+
+The interval property uses a [cron expression](#http://www.quartz-scheduler.org/documentation/quartz-2.3.0/tutorials/tutorial-lesson-06.html)) 
+to describe a repeating interval upon which the state becomes active and a new workflow instance is created.
```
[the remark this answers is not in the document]
tsurdilo: fixed

specification/schema/serverless-workflow-schema.json
tsurdilo: fixed

specification/README.md
tsurdilo: fixed

tsurdilo:
@singhkays @ricardozanini I made some updates to this PR after our discussions on the issue and also after thinking about this some more :) 

@singhkays 
I have used your overloading idea which I thought was very good. Please look at the updates to see if they fit your needs better now.
Basically the start definition now you can do either (for scheduled):

```json
"start": {
              "kind": "scheduled",
              "schedule": {
                "interval": "2020-03-20T09:00:00Z/2020-03-20T15:00:00Z"
              }
          }
```

or

```json
"start": {
            "kind": "scheduled",
            "schedule": {
                "cron": "0 0/15 * * * ?"
            }
        }
```

I have also lifted the restriction of event based starting states not being able to use the cron definition, however added a note
for runtime implmentors on what should happen in order to trigger workflow instances in this scenario
WDYT?

singhkays:
@tsurdilo Overloading looks good 👍 

Re: below
> Defining a cron-based scheduled starts for the runtime implementations would mean that there needs to be an event service which issues 
the needed events at the defined times to trigger workflow instance creation.

I'm trying to understand a couple of things:
1. What would happen if the event was sent at a time other than cron specified time (e.g. 0 0/15 * * * ?)? 
2. To write a workflow in the example I described above, would I use Event state or Operation state as the starting state to allow ad-hoc execution in addition to cron scheduled execution?

tsurdilo:
@singhkays 

It is really up to the runtime engine for 1. and 2. to allow/not allow instance creation outside the cron/interval definitions IMO for both Event and Operation (+ other non-event-triggered starting) states.

Do you think that we should add a flag for this behavior where the default should be not to be able to create instances outside what is defined but can be set to allow it? I think this decision should be something the person modeling a workflow should set as its most likely mandated by business requirements, rather than depending on some behavior of the runtime (which can differ from one to another)?

singhkays:
Interesting.. I can see a use-case of not allowing a said cron workflow outside of scheduled time. 

I'm not too sure if the default should be to block or allow. In my experience, choosing a default is always tricky in cases where users don't read the docs and end up shooting themselves in their foot :) Could the default option be an item that could be voted on by the larger community so we can get more eyes on it?

tsurdilo:
@singhkays sure, we could bring this up in our next meeting. @manuelstein if you don't mind can we add this as an agenda item ? thanks.

tsurdilo:
@singhkays we could also not have a default and force a decision to be made with a boolean param for example. I just thought having a default could be useful but I see from your point that yes it can be tricky.

singhkays:
@tsurdilo Can you show me an example of how a boolean param would look in the spec?

tsurdilo:
@singhkays 

Let's say we call this param "directInvoke" with possible values "allow" and "deny" we could have for example:

1) Don't set "allow" or "deny" as default in the json schema, thus making this param required:

```json
 "start": {
   "kind": "scheduled",
   "schedule": {
   "cron": "0 0/15 * * * ?",
   "directInvoke": "allow"
   }
}
```

2) Set "allow" as default value of this param (thus making it not required in json schema, and its use would be to overwrite the default with "deny"):

```json
 "start": {
   "kind": "scheduled",
   "schedule": {
   "cron": "0 0/15 * * * ?",
   "directInvoke": "deny"
   }
}
```

WDYT? (Names of param and its values are of course tbd :) )

tsurdilo:
rebased

singhkays:
I like the first option. It forces the user to chose an option. Also if they didn't read the documentation, the schema forces them to read upon the usage of this option. 

For the naming, need to think this through. Something that is anti of a "scheduled execution" might be more clearer. For e.g. adhocExecution but that's a terrible example 😆

tsurdilo:
@singhkays 

agree with the first option as well. 
for the name, yeah let me know what you'd like to see used. Ad-hoc is typically used in workflows/processes to describe unstructured processes where theirs flow is not pre-defined, something we are planning to add to spec in the future so I'd like to reserve that word for that if possible :)

singhkays:
Reserving ad-hoc for future features sounds good. Can't think of another name atm so open if others have suggestions. If not, directInvoke will do.

tsurdilo:
rebased

tsurdilo:
@singhkays added the directInvoke param. Yes let's make a decision on this is you want in our next meet. I think time zone can be figured out by the runtime impls.

singhkays:
@tsurdilo Do we need a parameter for time zone? How would a runtime implementation take in the input from the user?

tsurdilo:
rebased

tsurdilo:
@singhkays my understanding is that time zone can be figured out by the runtime impls. Let me know if you agree or not. If you think we need a specific param we could add, but I assume if thats the case it could be also needed as input to other definitions. Would setting a time zone as a metadata input on the workflow def level work for you ?

singhkays:
Metadata approach might work but it might be better to be more explicit as I'm not sure another place in the spec where a timezone is needed. I was looking into Argo workflow example comparisons and came across this as one of the options for cron workflows in Argo docs.

tsurdilo:
@singhkays good point. agree with you and added the timezone property. Let me know what you think. Thanks.

specification/README.md
```diff
@@ -2568,6 +2569,9 @@ Once a workflow instance is created, the start state schedule can be ignored for
 The cron property uses a [cron expression](http://crontab.org/) 
 to describe a repeating interval upon which the state becomes active and a new workflow instance is created.
 
+The timezone is used to define a time zone name to evaluate the cron expression against. If not specified it should default to the local
+machine time zone. See [here](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) for a list of timezon names.
```
singhkays, line 2573: "timezon" typo

singhkays:
👍  lgtm

specification/README.md
tsurdilo: fixed. thanks.

tsurdilo:
rebased

tsurdilo:
@singhkays in our meet today I think everyone was OK with the changes made in this PR. Can you give it a "final GO" and I will merge it :) Thanks!

singhkays:
@tsurdilo lgtm 👍

PR #50, opened 2020-06-29
Allow actions to be used instead of states in ForEach state and Paral…

tsurdilo:
…lel State branches

- [x] Specification
- [ ] Examples
- [x] Schema
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] Website
- [ ] SDK
- [ ] TCK
- [ ] Other

Issue: https://github.com/cncf/wg-serverless-workflow/issues/39

Allows ForEach state and branches in Parallel state to chose to define either states or actions to be executed.

### DIFF
--- a/specification/README.md
+++ b/specification/README.md
@@ -1523,7 +1523,7 @@ states:
 
 Each branch receives the same copy of the Parallel state's data input.
 
-A branch can define either actions to be executed within that branch or states.
+A branch can define either states or actions to be executed.
 
 States within each branch are only allowed to transition to states defined in the same branch.
 Transitions to other branches or workflow states are not allowed.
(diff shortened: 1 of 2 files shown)


falko:
This raises the question of why the language needs both states and actions. Wouldn't one of them be enough?

tsurdilo:
@falko I think there are use causes for both. There is an issue: https://github.com/cncf/wg-serverless-workflow/issues/40 which is like "part two" of this work to simplify the markup. It is similar scenario where you can either express the work that needs to be done via actions (list of operation states), or more complex scenario where control flow logic is needed. 

Allowing actions is just a simplification for users (they do not have to define list of operation states). Both cases can be seen as equivalent.

PR #75, opened 2020-07-14
Add ability to invoke services in actions via event

tsurdilo:

- [x] Specification
- [x] Examples
- [x] Schema
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] Website
- [ ] SDK
- [ ] TCK
- [ ] Other

Some services cannot be invoked via a source uri but only via events. This pr adds the ability to reference a trigger and result event in action definitions to cover this scenario.

Also for event correlation rules, we only so far allow a single string-based "token" which references a name of a CE context attribute. Adding ability to add more than one correlation rule as well as be able to correlate on context attribute value of multiple events to be the same, as well as set a specific value of a context attribute.


tsurdilo:
rebased

PR #159, opened 2020-10-23
Add mechanism to control non-determinism in retries

### LINKED ISSUE #155: Add mechanism to control non-determinism in retries
On the [retry definition](https://github.com/serverlessworkflow/specification/blob/master/specification.md#retry-definition) there is an option to increase the time period to be waited between attempts ("multiplier"). Many exponential backoff implementations also make use of a (usually small) bounded random amount of time that is added to the delay.

It would be nice if this bounded random amount of time (sometimes referred to as "jitter") was part of the spec, it can be very important to certain workloads and IMO it's better that this is enforced by the specification than to be left up to the implementation.

This parameter could be of type integer and be measured in milliseconds. The delay for the `n-th` attempt could be defined as: `interval * exp(multiplier, n-1) + rand(0, jitter)`, though I believe the exact formula could be left up to the implementation. The key point here is enforcing a mechanism to provide bounded randomness.

**Why is this needed**:

This is important in scenarios where the transient failure is a result of a race condition (for example) between two actions which could be, by bad luck, causing the failure and waiting the same amount of time before retrying and causing the failure again, only to wait for the same amount of time again (in this scenario, their "interval" and "multiplier" would be the same).

LCaparelli:
Signed-off-by: Lucas Caparelli <lucas.caparelli112@gmail.com>

- [X] Specification
- [X] Schema
- [ ] Examples
- [ ] Usecases
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

Add mechanism to control non-determinism in retries. Fix #155 

I was not sure if the spec and the schema is everything I needed to change. Sorry if other things needed modification as well, just let me know and I'll add them :-)

### DIFF
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -339,6 +339,10 @@
           "minimum": 0,
           "minLength": 0,
           "description": "Maximum number of retry attempts. Value of 0 means no retries are performed"
+        },
+        "jitter": {
+          "type": "string",
+          "description": "Maximum random amount of time added or subtracted from the delay between each retry (ISO 8601 repeatable format)"
         }
       },
       "required": [
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -341,8 +341,10 @@
           "description": "Maximum number of retry attempts. Value of 0 means no retries are performed"
         },
         "jitter": {
-          "type": "string",
-          "description": "Maximum amount of random time added or subtracted from the delay between each retry (ISO 8601 duration format)"
+          "type": ["number","string"],
+          "minimum": 0.0,
+          "maximum": 1.0,
+          "description": "If float type, maximum amount of random time added or subtracted from the delay between each retry relative to total delay (between 0.0 and 1.0). If string type, absolute maximum amount of random time added or subtracted from the delay between each retry (ISO 8601 duration format)"
         }
       },
       "required": [
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -342,7 +342,7 @@
         },
         "jitter": {
           "type": "string",
-          "description": "Maximum random amount of time added or subtracted from the delay between each retry (ISO 8601 repeatable format)"
+          "description": "Maximum amount of random time added or subtracted from the delay between each retry (ISO 8601 duration format)"
         }
       },
       "required": [
--- a/specification.md
+++ b/specification.md
@@ -1332,7 +1332,7 @@ transition:
 | interval | Interval value for retry (ISO 8601 repeatable format). For example: "R5/PT15M" (Starting from now repeat 5 times with 15 minute intervals)| string | no |
 | multiplier | Multiplier value by which interval increases during each attempt (ISO 8601 time format). For example: "PT3S" meaning the second attempt interval is increased by 3 seconds, the third interval by 6 seconds and so on | string | no |
 | maxAttempts | Maximum number of retry attempts. Value of 0 means no retries are performed | string or integer | no |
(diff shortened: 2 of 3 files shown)


schema/workflow.json
```diff
@@ -339,6 +339,10 @@
           "minimum": 0,
           "minLength": 0,
           "description": "Maximum number of retry attempts. Value of 0 means no retries are performed"
+        },
+        "jitter": {
+          "type": "string",
+          "description": "Maximum random amount of time added or subtracted from the delay between each retry (ISO 8601 repeatable format)"
```
tsurdilo, line 345: should not be repeating intervals, but time duration format

specification.md
```diff
@@ -1332,6 +1332,7 @@ transition:
 | interval | Interval value for retry (ISO 8601 repeatable format). For example: "R5/PT15M" (Starting from now repeat 5 times with 15 minute intervals)| string | no |
 | multiplier | Multiplier value by which interval increases during each attempt (ISO 8601 time format). For example: "PT3S" meaning the second attempt interval is increased by 3 seconds, the third interval by 6 seconds and so on | string | no |
 | maxAttempts | Maximum number of retry attempts. Value of 0 means no retries are performed | string or integer | no |
+| jitter | Maximum random amount of time added or subtracted from the delay between each retry (ISO 8601 repeatable format) | string | no |
```
tsurdilo, line 1335: same comment here regarding "repeating", should be time duration

schema/workflow.json
tsurdilo, line 345: I wonder if we can reword this a little, saying this parameter represents a "random amount" might mean that its like a random number, different for each instance of this particular workflow, but it's not.

tsurdilo (changes requested):

schema/workflow.json
LCaparelli: Well, the amount of time that gets added is the random part, so maybe we could move the "random" word a bit further down, like:

```suggestion
          "description": "Maximum amount of random time added or subtracted from the delay between each retry (ISO 8601 repeatable format)"
```

I think this way it's harder to interpret that the maximum amount is random. Wdyt?

tsurdilo: +1

tsurdilo (approved):

tsurdilo:
@LCaparelli looks good to me. thanks! We will go over it in our meet on Monday to get others opinions as well.

tsurdilo:
@LCaparelli looks good to me w/ the changes discussed. We can merge after updates.

LCaparelli:
@tsurdilo done, thanks for the review!

Should I be concerned with the failing action? Looks like the docker build is failing due to a failure in `npm`. Could the changes have caused that?

tsurdilo:
@LCaparelli not sure why this "greeting" action is failing honestly. I'll just remove it.

tsurdilo:
@LCaparelli removed that action. Mind amending this pr? For example just add your name to: https://github.com/serverlessworkflow/specification/blob/master/community/contributors.md

ricardozanini (approved):

tsurdilo:
Waiting on @LCaparelli to update jitter param to be either string or integer value as discussed and then for @jorgenj to take a look at approve. Thanks guys!

LCaparelli:
@tsurdilo @jorgenj done, please let me know your thoughts :-)

jorgenj (approved):
LGTM!

PR #167, opened 2020-10-28
Specify how conditions ordering is controlled

### LINKED ISSUE #156: Should the order of conditions in a Switch State be honored by implementations?
**What is the question**:
Are implementations expected to honor the order of Switch State conditions? Seems logical, but it's not explicitly stated, so I thought it would be worth asking.

In a switch with 2 conditions (A and B), it's possible that both condition A and B are true, so order matters as either A or B are eligible for execution and could produce different results depending on which gets ran.

I believe this is important not only to data-based conditions, but to event-based conditions as well: it's possible that event A and B have already happened or arrive simultaneously when this switch state is reached and the order on which they are consumed matters.

LCaparelli:
Fix #156 

Signed-off-by: Lucas Caparelli <lucas.caparelli112@gmail.com>

- [X] Specification
- [ ] Schema
- [ ] Examples
- [ ] Usecases
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -1648,6 +1648,11 @@ There are two types of conditions for switch states:
 
 These are exclusive, meaning that a switch state can define one or the other condition type, but not both.
 
+At times multiple defined conditions can be evaluated to `true` by runtime implementations.
+Conditions defined first take precedence over conditions defined later. This is backed by the fact that arrays/sequences
+are orderd in both JSON and YAML. For example, let's say there are two `true` conditions: A and B, defined in that order.
+Because A was defined first, its transition will be executed, not B's.
+
 In case of data-based conditions definition, switch state controls workflow transitions based on the states data.
 If no defined conditions can be matched, the state transitions is taken based on the `default` property.
 This property can be either a `transition` to another workflow state, or an `end` definition meaning a workflow end.


specification.md
```diff
@@ -1631,6 +1631,15 @@ There are two types of conditions for switch states:
...
 These are exclusive, meaning that a switch state can define one or the other condition type, but not both.
 
+The order on which conditions are defined must be respected by implementations unless priorities are speficied. 
+If priorities are specified, the most prioritary among satisfied conditions must have its transition executed.
+
+For example, let's say there are two conditions, A and B, defined in that order. If both A and B are satisfied
+then the A's transition must be executed. This behavior is akin to swtich statements in many programming languages.
+On the other hand, if A had a priority of 10 and B's priority was 20 then B's transition would be executed instead.
+
+If the `priority` parameter of one of the defined conditions is specified, all conditions must also specify it.
```
ricardozanini, line 1641: Not sure if this is true. All other nodes could have `integer.MAX` defined by default. Would improve UX, I don't want to start adding a new field in all my states just because I need two or more states to have prio.

LCaparelli: This is per state though, all conditions in the same state must specify priority. I can clarify that, thanks for pointing it out!

But do you think this is still undesirable within a single Switch State?

Int.MAX would effectively make all undefined priorities to take precedence over those with a priority. Wouldn't int.min be more reasonable?

Come to think of it, this parameter could probably use some boundaries. [0..100] for example, wdyt?

tsurdilo: I don't think we need to go as far as defining boundaries. The only restriction I would place is "positive integer including 0", so I would definitely add "minimum": 0 to the definition in json schema.

specification.md
```diff
@@ -1631,6 +1631,15 @@ There are two types of conditions for switch states:
 
 These are exclusive, meaning that a switch state can define one or the other condition type, but not both.
 
+The order on which conditions are defined must be respected by implementations unless priorities are speficied. 
+If priorities are specified, the most prioritary among satisfied conditions must have its transition executed.
```
tsurdilo, line 1635: Is "prioritary" a word? :) 
I would change this a little bit, for example, set up the issue up-front, something like:

At times multiple defined conditions can be evaluated to "true" by runtime implementations. This can lead to ambiguity on which condition and thus its transition the workflow control-flow logic should take next.
The specification has two ways of dealing with this ambiguity on the markup level:
1. Talk about the default order-based approach
2. Talk about the explicit approach using the priority parameter
3. Talk about how 1) and 2) are mutually exclusive, meaning its either one or the other (per switch state)

Hope this helps

specification.md
```diff
@@ -1631,6 +1631,15 @@ There are two types of conditions for switch states:
...
+
+For example, let's say there are two conditions, A and B, defined in that order. If both A and B are satisfied
+then the A's transition must be executed. This behavior is akin to swtich statements in many programming languages.
```
tsurdilo, line 1638: I would not put anything about "programming languages" here. However I would add that In the cases "priority" is used, and not all conditions define a priority OR certain conditions in this case have the SAME priority defined in the markup, the runtime engines are expected to notify the user about this (as this is definitely a logical workflow error)

tsurdilo (changes requested):

tsurdilo:
@LCaparelli overall looks really good. Thanks!
Left some comments. Don't be afraid to just write as you would talk to any of us when explaining the issue (personal). 
You could even add some examples and describe the expected outcome. 

Don't forget that there is also a "default" parameter for conditions, meaning that if none of them are evaluated as true, the transition taken will be the "default" one. Idk if this is covered later on in the section, but you may add that in there if needed.

specification.md
ricardozanini: To me, priority 0 takes over 1. So, if all other states or actors in this context have a greater number they will have lower priority.

Like @tsurdilo said minimum 0 in the schema will do the job and every other state that does not have this attribute, it's assumed it will take the lower priority.

specification.md
ricardozanini: @tsurdilo I would say that if `priority` is not set, the order-based will take place for the remaining states.

specification.md
tsurdilo: I think bpmn works like that (lower priority meaning higher rank) but we don't have to follow that. I think from just the language point of view or way we think, higher priority means higher rank. So we can just go with that for our lang def. What would be the counter-argument to this?

specification.md
tsurdilo: that would be confusing imo. in most cases I do believe that users will go with the default order-based ranking. We should imo not mix the two but make them mutually exclusive ways of defining priority.

LCaparelli: > Is "prioritary" a word? :)

I found some usages of it when googling, but I wasn't sure hehe. I'll change the wording as suggest and hopefully it will be simpler/clearer. :-)

> @tsurdilo I would say that if priority is not set, the order-based will take place for the remaining states.

I considered this approach, but we had previously talked about requiring all conditions to define priority (if one is defined) in #167. I see the value for UX in not requiring this, though with some complexity added for implementations.

I don't have a strong opinion regarding which way to go and would be fine with either, I'll send in changes once we decide which approach we're going to choose:

  a. require all conditions in a single state to have priorities if one of them has
  b. conditions with undefined priorities are checked first (in the order they appear) and then the ones which have priorities (ordered according to that)

specification.md
LCaparelli: To me lower numbers for higher priority seems a bit counter-intuitive, but historically there are plenty of examples choosing either way, so I don't think there would be a problem with any choice.

@tsurdilo any thoughts about this?

tsurdilo: yeah like i said to me its more logical that higher priority value means higher rank (gets picked first).

specification.md
tsurdilo: definitely 1)

ricardozanini: Well, I'm convinced. Let's make them mutually exclusive. If someone needs to control this, so they control each node.

LCaparelli:
@tsurdilo @ricardozanini  I tried making language more welcoming, clarifying a few things and making the actual parameter definition less ambiguous. Let me know your thoughts :-)

specification.md
```diff
@@ -1631,6 +1631,25 @@ There are two types of conditions for switch states:
...
+At times multiple defined conditions can be evaluated to `true` by runtime implementations.
+This can lead to ambiguity as to which condition should have its transition executed next.
+The specification has two ways of dealing with this ambiguity on the markup level:
+
+  1. order-based: conditions defined first take precedence over conditions defined later.
+  This is backed by the fact that arrays/sequences are orderd in both JSON and YAML.
+  2. priority-based: conditions with higher priority value take precedence over conditions with lower priority.
+  This value is taken from a condition's `priority` parameter.
+
+For example, let's say there are two conditions, A and B, defined in that order. If both A and B are satisfied
```
tsurdilo, line 1643: This sentence is a little confusing to me. Could we maybe make it more clear by stating which of the two ways 1. and 2. the example deals with?

For example what I'm thinking:

When using the order-based approach, let's say we have two data / event based conditions A, and B ....
On the other hand when using the priority-based approach and having the same example of two conditions A and B, ........

tsurdilo (changes requested):

tsurdilo:
@LCaparelli very nice update - thank you! Just left small comments hope its ok.

specification.md
```diff
@@ -1631,6 +1631,25 @@ There are two types of conditions for switch states:
...
+then the A's transition must be executed. On the other hand, if A had a priority of 10 and B's priority was 20
+then B's transition would be executed instead.
+
+If the `priority` parameter of one of the defined conditions is specified, all conditions in the same Switch State
+must also specify it. If the user fails to do so, the implementation must notify them of this.
+
+If two conditions in the same Switch State define the same value for the `priority` parameter, this is considered
+a logical error and implementations must notify users of this.
```
tsurdilo, line 1651: Change "implementations" to "runtime implementations" to be clear please. Same above (there are two places)

LCaparelli:
@tsurdilo I didn't use the exact same words as your suggestion because I wanted to break the example down a bit more, I think it's much clearer now. Wdyt?

tsurdilo (approved):

tsurdilo:
@LCaparelli very nice. thanks!

ricardozanini (approved):

tsurdilo:
@LCaparelli hi :) we havent merged this yet as there are still some questions that came up during our last meetings. 
If you don't mind please get in touch with @manuelstein directly or if you have a chance join one of the next meetings and we can discuss those questions there. Thanks!

LCaparelli:
> 
> 
> @LCaparelli hi :) we havent merged this yet as there are still some questions that came up during our last meetings.
> If you don't mind please get in touch with @manuelstein directly or if you have a chance join one of the next meetings and we can discuss those questions there. Thanks!

Hey! Sorry, I couldn't attend the past couple of meetings, but I'll be there at the next one. We can discuss further then or over here if @manuelstein would prefer that :-)

LCaparelli:
@tsurdilo @manuelstein, terribly sorry, I got sucked into an urgent customer call which took place at the same time as the meeting and I couldn't join. Unfortunately the meeting is during my commercial hours, so sometimes I'm unable to join. :-(

manuelstein:
No worries. This has been delayed because I have questions about the use case where this would be applied. We're encoding serverless workflows in either json or yaml. Either way, the order of lists/arrays is preserved. I don't see why conditions should be assigned a priority attribute to them. Can you maybe explain the use case a bit more?

I think if the use case is clearer, also the description should explain what happens when two conditions have the same priority.

Currently, if any item has a priority, all the other must define a priority, too.

I think priority attributes introduce a lot of complexity to maintain a workflow definition, so I think this needs a really good use case. Otherwise it could still be defined as an extension.

LCaparelli:
@manuelstein the initial proposal was to rely on the fact that arrays are ordered in both json and yaml, as you have mentioned. However, @tsurdilo brought up his concern that some business automation runtimes might not represent this as an ordered list [at the original issue](https://github.com/serverlessworkflow/specification/issues/156#issuecomment-717330823).

I am not aware of an example use case for this myself, but I agree that having the priorities as an extension might be a more adequate middle ground, removing complexity from most use cases (which I imagine would have no trouble simply relying on the ordering specified in markup) while still supporting these edge cases.

Do you also agree @tsurdilo?

tsurdilo:
@LCaparelli @manuelstein yes that was added because I asked @LCaparelli to do it :) 
Having priority which is optional (with the restrictions that we defined that if used has to be used across all conditions)
allows us to not have to later on update spec  in case users are executing this on runtime where depending on order might not be possible. In addition, the use of priority might be more descriptive for some users that might actually want to express it regardless of the ordering.

LCaparelli:
Thought about this some more and I'm not sure of how much complexity this really adds @manuelstein, considering they're optional as @tsurdilo pointed out. For most cases, users would be able to use this without even being aware that setting priorities is an option and it would work just fine using the specified ordering.

The way I see it, since priorities are optional, they add functionality for those who need it without really affecting those who don't.

> I think if the use case is clearer, also the description should explain what happens when two conditions have the same priority.

Sorry, forgot to address this. This is actually explained in this PR changes:

```
If two conditions in the same Switch State define the same value for the priority parameter, this is considered a logical error and runtime implementations must notify users of this.
```

LCaparelli:
@tsurdilo as agreed upon during the meeting I made the necessary changes to no longer enforce a priority-based approach and to rely solely on the order on which conditions were defined on the markup.

Let me know your thoughts :-)

tsurdilo (approved):

tsurdilo:
@LCaparelli looks good. thanks for making the updates (again :)) and sorry you had to do it cause of me

Issue #125, opened 2020-08-17
Update workflow error handling section


tsurdilo:
**What would you like to be added**:
Current "workflow error handling" section does not clearly define how errors / retries / timeouts etc can be handled.
Examples do not include some error checking / timeout / retry examples either. 
We should update the spec to address both points.
**Why is this needed**:

tsurdilo:
Handled by PR https://github.com/serverlessworkflow/specification/pull/175

tsurdilo:
closed by commit https://github.com/serverlessworkflow/specification/commit/07e0b86da75564d36749ff4df876f7d4e898e5db

PR #175, opened 2020-11-06
Workflow error handling updates

tsurdilo:

- [x ] Specification
- [ x] Schema
- [ x] Examples
- [ ] Usecases
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

This PR updates the workflow error handling for the specification:

allows for defining domain-specific errors
combines error and retry handling which was separate before
significantly reduces error handling complexity
removes error data filter

retry definition strategies can be defined in a top-level workflow "retries" array now.
on states onErrors error definitions you can reference a particular strategy to be used for the specific error defined

### DIFF
--- a/examples/examples.md
+++ b/examples/examples.md
@@ -963,19 +963,19 @@ The data output of the workflow contains the information of the exception caught
     },
     "onErrors": [
        {
-         "name": "Missing order id",
+         "error": "Missing order id",
          "transition": {
            "nextState": "MissingId"
          }
        },
        {
-         "name": "Missing order item",
+         "error": "Missing order item",
          "transition": {
            "nextState": "MissingItem"
          }
        },
        {
-        "expression": "Missing order quantity",
+        "error": "Missing order quantity",
         "transition": {
           "nextState": "MissingQuantity"
         }
@@ -1045,13 +1045,13 @@ states:
   transition:
     nextState: ApplyOrder
   onErrors:
-  - name: Missing order id
+  - error: Missing order id
     transition:
       nextState: MissingId
-  - name: Missing order item
+  - error: Missing order item
     transition:
       nextState: MissingItem
-  - expression: Missing order quantity
+  - error: Missing order quantity
     transition:
       nextState: MissingQuantity
 - name: MissingId
@@ -1157,7 +1157,7 @@ In the case job submission raises a runtime error, we transition to a SubFlow st
       ],
       "onErrors": [
       {
-        "name": "*",
+        "error": "*",
         "transition": {
           "nextState": "SubmitError"
         }
@@ -1304,7 +1304,7 @@ states:
     actionDataFilter:
       dataResultsPath: "{{ $.jobuid }}"
   onErrors:
-  - name: "*"
+  - error: "*"
     transition:
       nextState: SubmitError
   stateDataFilter:
--- a/examples/examples.md
+++ b/examples/examples.md
@@ -891,10 +891,9 @@ states:
 #### Description
 
 In this example we show off the states error handling capability. The workflow data input that's passed in contains
-missing order information that causes the function in the "ProvisionOrder" state to throw a runtime exception. With the "onError" expression we
-can transition the workflow to different error handling states depending on the error thrown. Each type of error
-in this example is handled by simple delay states, each including an error data filter which sets the exception info as their
-data output. If no error is caught the workflow can transition to the "ApplyOrder" state.
+missing order information that causes the function in the "ProvisionOrder" state to throw a runtime exception. With the "onErrors" definition we
+can transition the workflow to different error handling states. Each type of error
+in this example is handled by simple delay states. If no errors are encountered the workflow can transition to the "ApplyOrder" state.
 
 Workflow data is assumed to me:
 
@@ -956,32 +955,32 @@ The data output of the workflow contains the information of the exception caught
           }
        }
     ],
-    "onError": [
-       {
-         "expression": "{{ $.exceptions[?(@.name == 'MissingOrderIdException')] }}",
-         "transition": {
-           "nextState": "MissingId"
-         }
-       },
-       {
-         "expression": "{{ $.exceptions[?(@.name == 'MissingOrderItemException')] }}",
-         "transition": {
-           "nextState": "MissingItem"
-         }
-       },
-       {
-        "expression": "{{ $.exceptions[?(@.name == 'MissingOrderQuantityException')] }}",
-        "transition": {
-          "nextState": "MissingQuantity"
-        }
-       }
-    ],
     "stateDataFilter": {
        "dataOutputPath": "{{ $.exceptions }}"
     },
     "transition": {
        "nextState":"ApplyOrder"
-    }
+    },
+    "onErrors": [
+       {
+         "name": "Missing order id",
+         "transition": {
+           "nextState": "MissingId"
+         }
+       },
+       {
+         "name": "Missing order item",
+         "transition": {
+           "nextState": "MissingItem"
+         }
+       },
+       {
+        "expression": "Missing order quantity",
+        "transition": {
+          "nextState": "MissingQuantity"
+        }
+       }
+    ]
 },
 {
    "name": "MissingId",
@@ -1041,20 +1040,20 @@ states:
       refName: provisionOrderFunction
       parameters:
         order: "{{ $.order }}"
-  onError:
-  - expression: "{{ $.exceptions[?(@.name == 'MissingOrderIdException')] }}"
-    transition:
-      nextState: MissingId
-  - expression: "{{ $.exceptions[?(@.name == 'MissingOrderItemException')] }}"
-    transition:
-      nextState: MissingItem
-  - expression: "{{ $.exceptions[?(@.name == 'MissingOrderQuantityException')] }}"
-    transition:
-      nextState: MissingQuantity
   stateDataFilter:
     dataOutputPath: "{{ $.exceptions }}"
   transition:
     nextState: ApplyOrder
+  onErrors:
+  - name: Missing order id
+    transition:
+      nextState: MissingId
+  - name: Missing order item
+    transition:
+      nextState: MissingItem
+  - expression: Missing order quantity
+    transition:
+      nextState: MissingQuantity
 - name: MissingId
   type: subflow
   workflowId: handleMissingIdExceptionWorkflow
@@ -1156,12 +1155,9 @@ In the case job submission raises a runtime error, we transition to a SubFlow st
           }
       }
       ],
-      "onError": [
+      "onErrors": [
       {
-        "expression": "{{ $.exceptions[0] }}",
-        "errorDataFilter": {
-          "dataOutputPath": "{{ $.exception }}"
-        },
+        "name": "*",
         "transition": {
           "nextState": "SubmitError"
         }
@@ -1307,10 +1303,8 @@ states:
         name: "{{ $.job.name }}"
     actionDataFilter:
       dataResultsPath: "{{ $.jobuid }}"
-  onError:
-  - expression: "{{ $.exceptions[0] }}"
-    errorDataFilter:
-      dataOutputPath: "{{ $.exception }}"
+  onErrors:
+  - name: "*"
     transition:
       nextState: SubmitError
   stateDataFilter:
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -247,7 +247,7 @@
           "minLength": 1
         },
         "transition": {
-          "description": "Transition to next state to handle the error. If retryRef is define, this transition is taken only if retries were unsuccessful.",
+          "description": "Transition to next state to handle the error. If retryRef is defined, this transition is taken only if retries were unsuccessful.",
           "$ref": "#/definitions/transition"
         },
         "end": {
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -228,7 +228,7 @@
     "error": {
       "type": "object",
       "properties": {
-        "name": {
+        "error": {
           "type": "string",
           "description": "Domain-specific error name, or '*' to indicate all possible errors",
           "minLength": 1
@@ -238,18 +238,28 @@
           "description": "Error code. Can be used in addition to the name to help runtimes resolve to technical errors/exceptions",
           "minLength": 1
         },
-        "retry": {
-          "description": "Retry Definition",
-          "$ref": "#/definitions/retry"
-        },
         "transition": {
-          "description": "Transition to next state. If retry is defined, it is taken if defined retries were not successful",
+          "description": "Transition to next state to handle the error.",
           "$ref": "#/definitions/transition"
+        },
+        "end": {
+          "description": "End workflow execution in case of this error",
+          "$ref": "#/definitions/end"
         }
       },
-      "required": [
-        "name",
-        "transition"
+      "oneOf": [
+        {
+          "required": [
+            "error",
+            "transition"
+          ]
+        },
+        {
+          "required": [
+            "error",
+            "end"
+          ]
+        }
       ]
     },
     "onevents": {
@@ -329,9 +339,19 @@
       "type": "object",
       "description": "Retry Definition",
       "properties": {
+        "error": {
+          "type": "string",
+          "description": "Domain-specific error name, or '*' to indicate all possible errors",
+          "minLength": 1
+        },
+        "code": {
+          "type": "string",
+          "description": "Error code. Can be used in addition to the error to help runtimes resolve to technical errors/exceptions",
+          "minLength": 1
+        },
         "delay": {
           "type": "string",
-          "description": "Time delay between retry attempts (ISO 8601 time format)"
+          "description": "Time delay between retry attempts (ISO 8601 duration format)"
         },
         "multiplier": {
           "type": "string",
@@ -348,10 +368,31 @@
           "minimum": 0.0,
           "maximum": 1.0,
           "description": "If float type, maximum amount of random time added or subtracted from the delay between each retry relative to total delay (between 0.0 and 1.0). If string type, absolute maximum amount of random time added or subtracted from the delay between each retry (ISO 8601 duration format)"
+        },
+        "transition": {
+          "description": "Transition to next state if retries were unsuccessful",
+          "$ref": "#/definitions/transition"
+        },
+        "end": {
+          "description": "End workflow execution definition if retries were unsuccessful",
+          "$ref": "#/definitions/end"
         }
       },
-      "required": [
-        "maxAttempts"
+      "oneOf": [
+        {
+          "required": [
+            "error",
+            "maxAttempts",
+            "transition"
+          ]
+        },
+        {
+          "required": [
+            "error",
+            "maxAttempts",
+            "end"
+          ]
+        }
       ]
     },
     "functionref": {
@@ -477,6 +518,14 @@
             "$ref": "#/definitions/error"
           }
         },
+        "retries": {
+          "type": "array",
+          "description": "States retry definitions",
+          "items": {
+            "type": "object",
+            "$ref": "#/definitions/retry"
+          }
+        },
         "transition": {
           "description": "Next transition of the workflow after the time delay",
           "$ref": "#/definitions/transition"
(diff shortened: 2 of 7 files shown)


tsurdilo:
Closes issue https://github.com/serverlessworkflow/specification/issues/125

schema/workflow.json
```diff
@@ -228,20 +228,27 @@
...
         },
-        "errorDataFilter": {
-          "$ref": "#/definitions/errordatafilter"
+        "retry": {
+          "description": "Retry Definition",
+          "$ref": "#/definitions/retry"
         },
         "transition": {
-          "description": "Next transition of the workflow when expression is matched",
+          "description": "Transition to next state. If retry is defined, it is taken if defined retries were not successful",
```
ricardozanini, line 246: I can't understand this sentence, do you mind rephrasing it?

specification.md
```diff
@@ -3657,30 +3605,46 @@ decide to strictly enforce it.
...
+should be reported by runtime implementations and halt workflow execution, 
+
+Within workflow definitions, errors defined are `domain specific`, meaning they are defined within 
+the actual business domain, rather than their technical (programming-language-specific) description.
+
+For example we can define errors such as "Order not found", or "Item not in inventory", rather than having to
+use terms such as "java.lang.IllegalAccessError", or "response.status == 404", which 
+might make little to no sense to our specific problem domain, as well as may not be portable across various runtime implementations.
+
+In addition to the domain specific error name, users have the option to also an an optional error code
```
ricardozanini, line 3627:
```suggestion
In addition to the domain specific error name, users have the option to also add an optional error code
```

specification.md
```diff
@@ -3657,30 +3605,46 @@ decide to strictly enforce it.
...
+
+Runtime implementations must be able to map the error domain specific name (and the optional error code)
+to concrete technical errors that arise during workflow execution. 
+
+#### Defining Error Handling
+
+Errors are defined via the states `onErrors` property. It is an array of ['error'](#Error-Definition) definitions.
+
+Each error definition should have a unique `name` property. There can be only one error definition
+which has the `name`property set to the wildcard character `*`.
```
ricardozanini, line 3638:
```suggestion
which has the `name` property set to the wildcard character `*`.
```

specification.md
```diff
@@ -3829,31 +3767,38 @@ You can combine retry and onError definitions to define powerful error handling
...
-"afterAnyOtherErrorState" state.
+In this example, if workflow execution encounters the "Service call timeout" error,
+the states control-flow logic should be retried up to 4 times, with a 1 minute delay between
+each retry attempt.
+If the maximum amount of unsuccessful retries is reached, the workflow should transition to the next state
+as defined by the `transition` property. If one of the performed retries is successful,
+the states `transition` property should be taken, and the one defined in the error definition should be 
+ignored.
+
+In order to issue an retry, the current state execution should be halted first, meaning
```
ricardozanini, line 3791:
```suggestion
In order to issue a retry, the current state execution should be halted first, meaning 
```

schema/workflow.json
```diff
@@ -228,20 +228,27 @@
...
+        },
+        "code": {
           "type": "string",
-          "description": "JsonPath expression. Evaluated against error data. Is true if returns non empty result"
+          "description": "Error code. Can be used in addition to the name to help runtimes resolve to technical errors/exceptions",
+          "minLength": 1
         },
-        "errorDataFilter": {
-          "$ref": "#/definitions/errordatafilter"
+        "retry": {
```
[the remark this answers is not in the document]
tsurdilo: ok good point. i think there are good points for both approaches but i do see the value of decoupling. i will update here in a minute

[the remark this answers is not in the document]
tsurdilo: one thing i do want to mention is regarding "all other error are assumed to be transient/retryable"
please note that all errors , caught or retried **have** to be defined. 
we will not have a "catch-all" retry. we cannot issue retries on errors that the users do not think of (are not defined). so that part I will not do in this pr

tsurdilo:
@jorgenj made some big updates - please check again. i have separated the onErrors and retries definitions as per your request
I think this is pretty clean now.

tsurdilo:
@jorgenj i have one more idea...if separating onError/retries is too complex we could combine them in a following way:

1) allow users to define retry definitions w/o specifying a specific error attached to it
2) reference the retry definitions in onErrors for a specific error, that way the error itself is defined only in a single place

Thats just another idea if you guys think makes more sense

tsurdilo:
@jorgenj and I talked about this and have a good solution. So this pr is still work in progress :)

tsurdilo:
@jorgenj @ricardozanini @manuelstein 
Did a final update - the functionality is done now. Please review. Thanks!

specification.md
```diff
@@ -371,6 +376,11 @@ Here is an example of using external resource for event definitions:
 
 Referenced resource must conform to the specifications [Workflow Events JSON Schema](schema/events.json). 
 
+The `retries` property can be either an in-line [retry](#Retry-Definition) definition array, or an URI reference to
+a resource containing an array of [retry](#Retry-Definition) definition. 
+Referenced resource can be used by multiple workflow definitions. For more information about 
+using and referencies retry definitions see the [Workflow Error Handling](#Workflow-Error-Handling) section.
```
jorgenj, line 382: I think you meant referencing here
```suggestion
using and referencing retry definitions see the [Workflow Error Handling](#Workflow-Error-Handling) section.
```

specification.md
```diff
@@ -1323,12 +1323,30 @@ transition:
...
+
+For more information, see the [Workflow Error Handling](#Workflow-Error-Handling) sections.
+
 #### Retry Definition
 
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
-| expression | JsonPath expression. Should be evaluated against state data. True if the expression result is not empty. | string | yes |
-| interval | Interval value for retry (ISO 8601 repeatable format). For example: "R5/PT15M" (Starting from now repeat 5 times with 15 minute intervals)| string | no |
+| name | Unique retry strategy name | string | no |
```
jorgenj, line 1348: Shouldn't `name` be required?

specification.md
```diff
@@ -1369,30 +1387,28 @@ jitter: PT0.001S
 
 </details>
 
-Defines the state retry policy. 
-The `expression` property is a JsonPath expression that is evaluated against state data.
-This assures that both execution errors as well as actions error results can be used during evaluation.
-The expression must evaluate return a result in order to be true.
+Defines the states retry policy (strategy). This is an explicity definition and can be reused across multiple
```
jorgenj, line 1390:
```suggestion
Defines the states retry policy (strategy). This is an explicit definition and can be reused across multiple 
```

specification.md
```diff
@@ -1369,30 +1387,28 @@ jitter: PT0.001S
...
 {
-  "expression": "...",
-  "interval": "R4/PT1M",
+  "name": "Timeout Errors Strategy",
+  "delay": "PT1M",
   "multiplier": "PT2M",
   "maxAttempts": 4
 }
 ```
+which means that we will retry up to 4 times after waiting 1, 3 (1 + 2), 5 (1 + 2 + 2), and 7 (1 + 2 + 2 + 2) minutes.
```
jorgenj, line 1409:
```suggestion
which means that we will retry up to 4 times after waiting with increasing delay between attempts; in this example 1, 3 (1 + 2), 5 (1 + 2 + 2), and 7 (1 + 2 + 2 + 2) minutes between retries.  
```

specification.md
```diff
@@ -1998,19 +2014,18 @@ The workflow id defined cannot be the same id of the workflow there the branch i
...
 
 By default, exceptions that are not handled within branches stop branch execution and are propagated 
-to the Parallel state.
+to the Parallel state and should be handled with its `onErrors` definition.
 
-If the parallel states branch defines actions, all exceptions are propagated to the parallel state 
-and can be handled with the parallel states `onError` definition.
+If the parallel states branch defines actions, all exceptions that arise from executing these actions
+ are propagated to the parallel state 
+and can be handled with the parallel states `onErrors` definition.
```
jorgenj, line 2021: I think we should call out that the very first error to be caught should be used to determine control-flow for the workflow execution, that the workflow runtime should not be expected to gather and somehow act upon all exceptions thrown by all parallel branches.

It might be left unsaid perhaps, but I think this leaves runtime implementations to pro-actively cancel and/or ignore parallel branches once the first error bubbles out which could have dire consequences for workflow execution correctness (imagine the case where actions a & b are executed in parallel.  Action a fails with an error that should result in a retry of the current state, if b is still executing when the retry is attempted, then this could cause serious harm having both old and new instances of action b executing at the same time.  I'm not sure if this level of detail belongs in the spec here, or is left to the runtimes to ensure correctness of their workflow executions.  Thoughts?

specification.md
```diff
@@ -3657,30 +3624,46 @@ decide to strictly enforce it.
...
 
-Each state can explicitly catch runtime errors via its `onError` property. This property includes one or more
-definitions matching the error object properties and defining a transition to a workflow state representing the
-workflow execution flow in case of that particular error.
+For example, we can define errors such as "Order not found", or "Item not in inventory", rather than having to
+use terms such as "java.lang.IllegalAccessError", or "response.status == 404", which 
+might make little to no sense to our specific problem domain, as well as may not be portable across various runtime implementations.
+
+In addition to the domain specific error name, users have the option to also an optional error code
+to help runtime implementations with mapping defined errors to concrete underlying technical ones.
```
jorgenj, line 3647: I don't think it makes sense to allow specifying `code` if the name is `*`.  Should we call that out here that if name==`*`, then `code` must be left undefined?

specification.md
```diff
@@ -3713,31 +3696,23 @@ Let's take a look an an example "onError" definition inside a state:
 <td valign="top">
 
 ```yaml
-onError:
-- expression: "{{ $.errors[?(@.name == 'FunctionExecutionError')] }}"
+onErrors:
+- error: Item not in inventory
```
jorgenj, line 3700: I'm imagining that this is an error that might come out of an open-api based function.  It's not clear to me how the response from a RESTful API will get translated to an error message string here.  Is there even a standard format for error messages from open-api?

btw; this is more for my own understanding, not really a comment on this PR.

specification.md
```diff
@@ -3796,29 +3776,59 @@ You can combine retry and onError definitions to define powerful error handling
...
+
+</td>
+<td valign="top">
+
+```yaml
+retries:
+- name: Service Call Timeout Retry Strategy
+  delay: PT1M
+  maxAttempts: 4
+  transition:
```
jorgenj, line 3800: Also typo here to list `transition` parameter here, right?

specification.md
```diff
@@ -3796,29 +3776,59 @@ You can combine retry and onError definitions to define powerful error handling
...
-    "interval": "PT1M"
-  }
-],
-"onError": [
-  {
-    "expression": "{{ $.errors[?(@.name == 'FunctionExecutionError')] }}'",
+    "name": "Service Call Timeout Retry Strategy",
+    "delay": "PT1M",
+    "maxAttempts": 4,
     "transition": {
```
jorgenj, line 3784: This is a typo to have `transition` defined here, right?

jorgenj (approved):

specification.md
tsurdilo: it does not have to do with openapi. runtime impls have to allow for a mapping between explicit/domain-specific error names + code prop defined in workflows to actual errors.

specification.md
tsurdilo: agreed. i will add this a little bit lower in the "Defining Errors" section and also will try to update the schema to catch this case during validation

specification.md
tsurdilo: ok i think that was implied but putting this info in here would make sense. will update

specification.md
tsurdilo: yeah i forgot to remove after cut/paste sry

specification.md
tsurdilo: yep :)

specification.md
tsurdilo: yes . it is in the schema. will update

tsurdilo:
@jorgenj thanks for the review - very nice comments. updated.

examples/examples.md
```diff
@@ -891,10 +891,9 @@ states:
 #### Description
 
 In this example we show off the states error handling capability. The workflow data input that's passed in contains
-missing order information that causes the function in the "ProvisionOrder" state to throw a runtime exception. With the "onError" expression we
-can transition the workflow to different error handling states depending on the error thrown. Each type of error
-in this example is handled by simple delay states, each including an error data filter which sets the exception info as their
-data output. If no error is caught the workflow can transition to the "ApplyOrder" state.
+missing order information that causes the function in the "ProvisionOrder" state to throw a runtime exception. With the "onErrors" definition we
+can transition the workflow to different error handling states. Each type of error
+in this example is handled by simple delay states. If no errors are encountered the workflow can transition to the "ApplyOrder" state.
```
manuelstein, line 897: Since our default actions are OpenAPI-compatible, should we make a statement how the errors used in this example "Missing order item", "Missing order quantity" would relate to a Swagger specification of the action?

specification.md
```diff
@@ -1274,14 +1281,12 @@ to the trigger/produced event.
...
 
-Error definitions define runtime errors that can occur during workflow execution and how to handle them. For more information,
-see the [Workflow Error Handling section](#Workflow-Error-Handling).
-
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
-| expression | JsonPath expression. Should be evaluated against error data. True if expression result is not empty | string | yes |
-| [errorDataFilter](#error-data-filter) | Error data filter definition | object | yes |
-| [transition](#Transitions) | Next transition of the workflow when `expression` is true | object | yes |
+| error | Domain-specific error name, or '*' to indicate all possible errors | string | yes |
```
manuelstein, line 1286: In case code is defined, can the error be left blank?
"404" would be such an error code when the resource could not be found.

specification.md
```diff
@@ -1274,14 +1281,12 @@ to the trigger/produced event.
...
+| code | Error code. Can be used in addition to the name to help runtimes resolve to technical errors/exceptions. Should not be defined if error is set to '*' | string | no |
```
manuelstein, line 1287: I think a code, e.g. `code: 500`, can always be used on its own. People probably distinguish error cases by the repsonses, so e.g. my service could reply 500 Server Internal Error and attach an error object (as e.g. in this [Swagger component object example](https://swagger.io/specification/#components-object)). If I want to catch all the errors, should I specify `code: 500` AND `error: *` or just `code: 500`. My intuition tells me that I should omit `error` if I want to catch all of them.

specification.md
```diff
@@ -1296,12 +1301,9 @@ see the [Workflow Error Handling section](#Workflow-Error-Handling).
 
 ```json
 {
-   "expression": "{{ $.exceptions }}",
-   "errorDataFilter": {
-     "dataOutputPath": "{{ $.exceptions[0] }}"
-   },
+   "error": "Item not in inventory",
```
manuelstein, line 1304: I think we should guide the reader to the use of Swagger. Of course, the error can be domain-specific, i.e. for a different action kind, we'd allow arbitrary strings. But it would be cool if we could use capital camel case error names such as "ItemNotInInventory" to hine to the use of [objects in Swagger responses](https://swagger.io/specification/#components-object)

examples/examples.md
```diff
@@ -956,32 +955,32 @@ The data output of the workflow contains the information of the exception caught
...
+    "stateDataFilter": {
+       "dataOutputPath": "{{ $.exceptions }}"
+    },
+    "transition": {
+       "nextState":"ApplyOrder"
+    },
+    "onErrors": [
        {
-         "expression": "{{ $.exceptions[?(@.name == 'MissingOrderIdException')] }}",
+         "error": "Missing order id",
```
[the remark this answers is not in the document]
tsurdilo: the mapping of domain specific errors + optional error codes are up to runtimes to provide. 
given different runtimes for sw this cannot be defined within our specification.
this is a common thing that many of the process runtime engines already are doing, not something we are 
making up

examples/examples.md
tsurdilo: they do not relate to swagger. even tho swagger defines how to invoke a service, the underlying runtime impl that actually executes it from the swagger def can be in many different languages and as such the errors that it can produce could vary . runtimes must provide the capabilities for this mapping and not our spec or swagger

specification.md
tsurdilo: i would say no as the error defines the domain specific name. even if on the runtime side only choses to use the error code for the mapping to concrete runtime errors, this name is still going to be used for the reference to the retries top-level definitions

specification.md
tsurdilo: i think if you want to catch all 500 errors, you can still define lets say "error": "Internal Server Errors" and have code lets say 500 or whatever the runtime mapping requires.
The error property is the actual portable one so it should be required

specification.md
tsurdilo: we cant tie error handling to swagger. as mentioned they are not related. the openapi definition will still get executed by a runtime which might use different libs and programming languages to invoke services/operations on the swagger defs. as such the errors those runtimes create might have nothing to do with the actual definition in openapi.

so the mapping has to be be provided by the runtimes in this case.

tsurdilo: typically on the runtime side this is done by wrapping the underlying exception into a custom one which includes the error code or the defined error name. Then it can be related to the whats specified in the workflow definition

ricardozanini (approved):

examples/examples.md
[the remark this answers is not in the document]
manuelstein: but the way we define actions, we don't interact with language-specific runtimes, but with an OpenAPI spec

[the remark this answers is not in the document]
manuelstein: okay, maybe "Missing order id" is misleading here. This would be something I'd expect as a response from the OpenAPI-based action. If it's runtime-specific, then maybe "OutOfMemory" is a more fitting example?

examples/examples.md
manuelstein: Agree that runtimes can cause additional errors that are not covered by the action definition. But how portable is a workflow spec when its error handling is runtime-specific?

specification.md
manuelstein: AFAICS, there are 3 layers where we can get errors from: 

1. the domain specific naming defined through the API (e.g. my function throws a "OutOfBusinessHoursException"), which would be defined as a response in the Swagger spec, e.g. like a [names component object](https://swagger.io/specification/#components-object)
2. protocol errors of the action; the use of swagger suggests HTTP errors, which are response codes
3. runtime errors, such as transport specific errors (e.g. host not found) or system-related errors (out of memory, stack overflow)

manuelstein:
Since we won't find a solution to separate different layers of errors (domain, protocol, runtime), we can merge this as intermediary solution despite its lack of portability.

LGTM

tsurdilo:
@manuelstein thanks for the review

Maybe there is a little bit of confusion here:  the error handling we have here is not runtime specific at all. 
It is done with portability in mind and has many advantages over some of the "leading" workflow solutions on major cloud platforms right now which are indeed not portable :)  

In our spec errors are:
1) defined with a domain-specific name (portable)
2) enhanced with a domain-specific code (portable)

The only requirement we put on the runtime implementations is to map those domain-specific values in workflow markup 
to specific errors of their specific runtime impl. *This* makes the workflow lang 100% portable IMO. 

Note that errors cannot just arise of execution of actions (service invocations). Errors can come from many different places during workflow execution, and they will be raised by the runtime implementations. As such if we went any other way (without the requirement for a mapping) we would end up with a non-portable solution.

I can explain this further and also with examples if needed in one of our meetings if you think it would help.

Would really like to see an example of why you think this is not portable - please bring it up and I'll definitely take a look and we can discuss to improve.

examples/examples.md
tsurdilo: any runtime impl which choses to use SW lang for workflows will need to allow users to create a mapping btw. how errors are expressed in the workflow lang (domain-specific) to the internal errors. 

Most of the time this will be done by the runtimes catching an exception and wrapping it into a custom exception which will include a certain "code" that represents that type of exception. Workflow implementors should have this list of codes available and can use them.

The non-portability factor that I think you are seeing is that those codes might not be portable between runtimes, and there you are correct. When moving from one runtime to another two things could happen:
1) The new runtime provides the mapping to the codes expressed in the workflows 
2) Workflow implementors might have to adjust the codes in their workflows (not the best solution, but ok)

Either way the domain-specific error name stays the same. So yeah I think you do have a point so lets in further updates think maybe of better ways to handle this is there are ideas.

tsurdilo:
thanks for review everyone! I will go ahead and merge this now and we can make further improvements for next release version.

PR #194, opened 2020-12-01
Adding workflow 'visibility' parameter

tsurdilo:

- [ x] Specification
- [x ] Schema
- [ ] Examples
- [ ] Usecases
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

Adds workflow scope. A visibility setting that can be either "public" or "private"

### DIFF
--- a/roadmap/README.md
+++ b/roadmap/README.md
@@ -81,6 +81,7 @@ _Status description:_
 | --- | --- |  --- |
 | ✔️| Adding Workflow Compensation capabilities (cmp [Compensating Transaction](https://docs.microsoft.com/en-us/azure/architecture/patterns/compensating-transaction), [SAGA pattern](https://microservices.io/patterns/data/saga.html)) | [spec doc](../specification.md) |
 | ✔️| Adding comparison examples with Google Cloud Workflow language| [examples doc](../examples/examples-google-cloud-workflows.md) |
+| ✔️ | Adding workflow scope (visibility setting) | [spec doc](../specification.md) |
 | 🚩 | JSONPatch transformations | [issue](https://github.com/serverlessworkflow/specification/issues/149) |
 | 🚩 | Workflow invocation bindings |  |
 | 🚩 | CE Subscriptions & Discovery |  |
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -30,6 +30,15 @@
           "description": "Serverless Workflow schema version",
           "minLength": 1
         },
+        "scope": {
+          "type": "string",
+          "enum": [
+            "public",
+            "private"
+          ],
+          "default": "public",
+          "description": "Defines the scope of this workflow. Can be either 'public' or 'private'. Default is 'public'"
+        },
         "dataInputSchema": {
           "type": "string",
           "format": "uri",
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -30,14 +30,14 @@
           "description": "Serverless Workflow schema version",
           "minLength": 1
         },
-        "scope": {
+        "visibility": {
           "type": "string",
           "enum": [
             "public",
             "private"
           ],
           "default": "public",
-          "description": "Defines the scope of this workflow. Can be either 'public' or 'private'. Default is 'public'"
+          "description": "Defines the visibility of this workflow. Can be either 'public' or 'private'. Default is 'public'"
         },
         "dataInputSchema": {
           "type": "string",
--- a/specification.md
+++ b/specification.md
@@ -245,7 +245,7 @@ which would set the workflow version to `1.0.0`.
 | description | Workflow description | string | no |
 | version | Workflow version | string | no |
 | schemaVersion | Workflow schema version | string | no |
-| scope | Defines the scope of this workflow. Can be either "public" or "private". Default is "public" | string | no |
+| visibility | Defines the visibility of this workflow. Can be either "public" or "private". Default is "public" | string | no |
 | [dataInputSchema](#Workflow-Data-Input) | URI to JSON Schema that workflow data input adheres to | string | no |
 | [dataOutputSchema](#Workflow-data-output) | URI to JSON Schema that workflow data output adheres to | string | no |
 | [events](#Event-Definition) | Workflow event definitions.  | array or string | no |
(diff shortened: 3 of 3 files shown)


ricardozanini (approved):

jorgenj:
From the discussion in the weekly call, i think there's 2 aspects to this idea:

a) authorization; whom can invoke a workflow, which actors/persons has the appropriate authorization to invoke a workflow.  To me, AuthN and AuthZ feel like something that will be very runtime specific and should be left out of the spec.

b) Invocation methods; or 'how' can a workflow be invoked.  For example, public implies some API can be called to invoke the service as well as potentially via event or as a sub-workflow, where-as private implies that only a sub-workflow invocation will be allowed (if I understood the idea correctly).  Perhaps instead of adding scope, we should have the definition explicitly list the ways in which it can be invoked?  We already do this for event-invoked and scheduler-invoked workflows.  Currently, iirc, the spec doesn't really say much about api-invoked workflows, perhaps instead of implying it we should have definitions list when they can be API-invoked, and same for when they can be sub-workflow-invoked?

To give a concrete example, what if we changed slightly the way 'Start' for a workflow state could be defined such that you list there the exact ways in which this workflow is allowed to be invoked.  For example:

```
{  
...
"states":[  
  {  
   "start": {
     "kind": ["subWorkflow","scheduled"],
     "schedule": ...
   },
   "name":"firstState",
    ...
  },
  {  
   "name":"secondState",
   ...
  }
],
...
}
```

There's probably a better way to model this approach, but hopefully that gets the idea across.  In that example, that tells us that this workflow can be invoked by schedule or as a sub-workflow (but it's missing the 'default', so it can be invoked as a by some external api call).

manuelstein:
Thanks, @jorgenj, for summarizing these two aspects.
I can think of a third one where the workflow definition is either visible by just its creator or to all developers on the same runtime (like GitHub repos, Docker container images or OpenStack VM images), but that's a property that should be managed by the runtime IMHO, so that's why I said `scope` would fit as an annotation. I think such annotations would for now go into `metadata`.

**Invocation methods**
Regarding workflow invocation, we have the special handling of [Event-type states](https://github.com/serverlessworkflow/specification/blob/master/specification.md#event-state), that, if defined as a starting state, control when the workflow instances should be created. To me, a webhook invocation, OpenAPI binding, scheduled execution or invocation as a sub-flow are just additional ways to bind a workflow definition to invocation methods.

Some trivia: Argo and Tekton are treating workflow invocations as resources, i.e. a Workflow CRD (Argo) or TaskRun/PipelineRun CRD (Tekton) instance is created as invocation. Their coupling with Events uses event gateways that receive an event and create the resource as a result. IIRC, OpenWhisk also treats the invocation as a resource. Likewise, invocations in OpenStack Mistral are treated as addressable resources in a RESTful API.

Personally, I believe it shouldn't matter what creates the invocation. I think they're just bindings of the workflow specification to interfaces, whether it's a timer, a CloudEvent, a service function or a RESTful API. I'd prefer to have some common approach to any of these invocations/workflow bindings. CloudEvents are consumed in the EventState, a scheduled start state can trigger any state but with no input and we're now searching for a way to consume an exposed API call (e.g. a webhook call).

- Should we create a special state type for this that can be called using an OpenAPI call? (Like the event state that can receive a CloudEvent)
- Would you rather unify it all in a global section that names all the possible invocations and their starting states? (like in Moore/Mealy automata that have a starting state with a bunch of conditions/transitions that put the workflow in action)
- Should we unify this in the start struct and also allow event/function triggers like @jorgenj suggested?

tsurdilo:
I just wanted to add a simple construct that marks a workflow as "exposed publicly" or "not exposed publicly". Possibly made the text sounds like its tied to invocations, which is not. This is just simple, for example a public or private class or method in Java. Thats all this is ;)

tsurdilo:
rebased

tsurdilo:
@manuelstein @jorgenj After reading your comments I realized my pr was misleading (see my previous comment). I have updated this pr to:
* call the added top-level workflow paral "visibility" as that is closer to its intended purpose
* updated the parameter description

Please let me know if this is ok now. I think per your comments there is definitely things we can add to further define/restrict the invocation of workflows and I will look at that asap, but the purpose of this property/pr was not to deal with actual invocation changes.

Please review again if you have the chance.

jorgenj:
I'm confused about what it means for something to be 'visible'. I think this implies that the spec needs to describe other things that come with this.  For example, if a workflow is 'public' or 'private', then don't we need to describe the conditions under which a user/principal can see private workflows/APIs or not?  And how to determine what type of user a user/principal is (public vs. private).  IIRC, the spec doesn't currently describe the invocation API at all, but this seems to be adding restrictions to the process of invocation (ie, who is the user and can they access private workflows).

tsurdilo:
@jorgenj I think you might be mixing up visibility and access control here. 
A workflow being "private" just means it cannot be instantiated from outside and allows us to declare subflows that cannot be triggered as standalone (thus "private").  We will hopefully soon be adding access control to the language, but this is not part of those efforts. HTH

manuelstein:
Sorry, @tsurdilo, but I also can't follow what is "outside". I see the current idea is to make a distinction between a workflow being used as a subflow only (private) and a workflow being exposed through an API or Event (public). I don't think that the param is addressing this issue the right way. Whether or not a workflow is bound to an API (gateway) exposes it to external calls (with and without authorization, i.e. allowing the public or only authenticated and authorized users to invoke them).

What confuses me is: when a workflow is marked private and I have more than one tenant on my system, can tenant A still invoke tenant B's workflow as a subflow? And if no, what would I choose if I don't want it exposed through the external-facing API, but still I want everyone on my platform to be able and use it as a subflow?

Should a workflow that can be triggered by an event (starting state is Event-type) still exposed through an API?
IIUC, we want to have workflows that have no external-facing exposure (neither events, nor API calls), so I think we should make these bindings optional then.

tsurdilo:
@manuelstein 
regarding:
**when a workflow is marked private and I have more than one tenant on my system, can tenant A still invoke tenant B's workflow as a subflow? And if no, what would I choose if I don't want it exposed through the external-facing API, but still I want everyone on my platform to be able and use it as a subflow?** 
Yes. Marking a workflow as "private" means it can be used by other workflows only (via the workflowId parameter in parallel states or subflow states etc)

regarding:
**Should a workflow that can be triggered by an event (starting state is Event-type) still be exposed with an URL endpoint?**
It can, up to the runtimes how they deal with even delivery. See https://github.com/cloudevents/spec/blob/v1.0/http-protocol-binding.md for example. Even if the event delivery is a message broker for example, exposing workflow still as a restful service might be very useful for testing scenarios.

**IIUC, we want to have workflows that have no external-facing exposure (neither events, nor API calls), so I think we should make these bindings optional then.**
Only API calls. If a public workflow triggers exec of a private subflow and that subflows includes an event state, consumption of events should be possible for it.

tsurdilo:
rebased

ricardozanini:
> Should a workflow that can be triggered by an event (starting state is Event-type) still be exposed with an URL endpoint?

Hey @manuelstein just adding to @tsurdilo reply, CloudEvents can also be sent via HTTP, so a REST endpoint to receive the event can be an alternative for such cases. Knative for instance makes `POST` requests to the service root path (`/`)  in the root path.

manuelstein (changes requested):
Ok, thank you for the replies, I think I got its purpose.
I'm suggesting to rename this property to subflow: true (default: false), meaning if the property is present, it can only be used as a subflow (I think that's consistent with our terminology of sub-flow invocation). Visibility/Public/private has to many connotations of access control and tenant resource sharing (making things visible to others). But I understand it is completely independent from resource sharing or access control, meaning a workflow marked as subflow (i.e. it can be invoked only as subflow) can _still_ be access controlled (e.g. team members can, other departments can't) and can _still_ be exposed on public-facing endpoints (e.g. message-broker, API gateway)

So, the current solution using `visibility: private` could mean that _still_, I can share it with others and _still_, it may receive events or callbacks from a public interface. I think if it was renamed to `subflow: true (default: false)`, then the majority of workflows wouldn't care and can be instantiated by any sort of trigger, but if marked a `subflow`, it can only be used as such, i.e. triggered by another workflow.

Last question though: How could I write a workflow that runs on one platform, but triggers a subflow running on another platform?

ricardozanini:
I think `subflow: true` could work on scenarios where the workflow won't be trigger by any other means but from a parent workflow. These sub-flows won't receive or react upon events. Not sure if this is what we want to achieve in this PR.

I agree that `visibility: true` could confuse people with `access`, but I prefer this option to `subflow`.

tsurdilo:
After meeting discussions, we decided to close this for now and may re-look if there is a need for it down the line

PR #210, opened 2021-01-03
Update to start/end definitions

tsurdilo:

- [ x] Specification
- [x ] Schema
- [x ] Examples
- [ ] Usecases
- [ x] Extensions
- [ x] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

Updates the definitions of start and end definitions. Allows them to be either boolean or object types.
This removes the need for:
```json
"start": { "kind": "default" }
```
```json
"end": { "kind": "default" }
```
to
```json
"start": true
```
```json
"end": true
```

Also streamlines the definitions by removing their "kind" property completely.

### DIFF
--- a/examples/examples.md
+++ b/examples/examples.md
@@ -62,15 +62,11 @@ data output, which is:
   {  
      "name":"Hello State",
      "type":"inject",
-     "start": {
-       "kind": "default"
-     },
+     "start": true,
      "data": {
         "result": "Hello World!"
      },
-     "end": {
-        "kind": "default"
-     }
+     "end": true
   }
 ]
 }
@@ -87,12 +83,10 @@ description: Inject Hello World
 states:
 - name: Hello State
   type: inject
-  start:
-    kind: default
+  start: true
   data:
     result: Hello World!
-  end:
-    kind: default
+  end: true
 ```
 
 </td>
@@ -156,9 +150,7 @@ Which is added to the states data and becomes the workflow data output.
   {  
      "name":"Greet",
      "type":"operation",
-     "start": {
-       "kind": "default"
-     },
+     "start": true,
      "actions":[  
         {  
            "functionRef": {
@@ -172,9 +164,7 @@ Which is added to the states data and becomes the workflow data output.
            }
         }
      ],
-     "end": {
-       "kind": "default"
-     }
+     "end": true
   }
 ]
 }
@@ -194,8 +184,7 @@ functions:
 states:
 - name: Greet
   type: operation
-  start:
-    kind: default
+  start: true
   actions:
   - functionRef:
       refName: greetingFunction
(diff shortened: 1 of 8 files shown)


examples/examples.md
```diff
@@ -1747,7 +1661,7 @@ states:
         parameters:
           patientid: "{{ $.patientId }}"
   end:
-    kind: terminate
+    terminate: true
```
ricardozanini, line 1664: What does this terminate attribute mean? Just `end: true` doesn't mean the same thing?

extensions/kpi.md
```diff
@@ -176,8 +175,7 @@ states:
         refName: callPulmonologist
         parameters:
           patientid: "{{ $.patientId }}"
-  end:
-    kind: terminate
+  end: true
```
ricardozanini, line 178: Here we are using `end: true` instead of `terminate: true` :thinking:

specification.md
```diff
@@ -2736,13 +2707,14 @@ schedule:
...
 
-Any state can declare to be the start state of the workflow, meaning that when a workflow instance is created it will be the initial
+Any state can declare to be the start state of the workflow, meaning when a workflow instance is created it will be the initial
 state to be executed. A workflow definition can declare one workflow start state.
 
-The start definition provides a `kind` property, which describes the starting options:
-
-- **default** - The start state is always "active" and there are no restrictions imposed on its execution.
-- **scheduled** - Scheduled start states have two different choices. You can define time-based intervals during which workflow instances are **allowed**
+The start definition can be either `boolean` or `object` type.
```
ricardozanini, line 2713: Wouldn't be better to add a new attribute to `start`, this way our tooling would know how to properly parse it? I mean instead of `try parse object else boolean`, we parse without conditionals. I think it will bring more complexity than it should.

ricardozanini (commented):
I just left a couple of questions, looks good to me overall. I always felt that `kind` attribute was a little bit awkward.

examples/examples.md
tsurdilo: yeah its different and i will do another pr to describe this better.
"end": true (without the terminate) stops instance execution, but does **not** terminate the instance itself. The instance is still active, meaning that correlated events can still be associated with that instance

"terminate": true terminate the instance completely, meaning its execution completely stops. new comsumed events at that point even if correlated will create a new workflow instance.

hope this helps.

extensions/kpi.md
tsurdilo: good catch ill update

specification.md
tsurdilo: its about the language expression not the tooling complexity. you can try to parse it as boolean, and if its not then assume object.

"start": true and "end": true as default are much easier to understand for users imo

ricardozanini: Yes `start: end|true` is super easier to understand indeed. But just a note about parsing, some tooling will always parse "boolean" as false if it's not `true`. It's a nitpick, I agree we should prefer usability to implementation complexity.

ricardozanini (approved):

tsurdilo:
rebased

Issue #213, opened 2021-01-05
Group dataInputSchema/dataOutputSchema


jensg-st:
**What would you like to be added**:

Replace the `dataInputSchema` and `dataOutputSchema` parameters attached to each state with a single object `dataSchema`, which has two analogous fields `input` and `output`. 

Alternatively, merge both fields with the `stateDataFilter` to group all four. Something like `stateData.inputFilter`, `stateData.outputFilter`, `stateData.inputSchema`, `stateData.outputSchema`.

**Why is this needed**:

I notice that `stateDataFilter` already groups two similar fields together. For consistency, the data schema fields should get the same treatment. This will consolidate the documentation about data schemas into a single place in the specification, and it will have a side benefit of making each state definition appear simpler and cleaner.

tsurdilo:
@jensg-st makes sense. will work on it asap.

tsurdilo:
@jensg-st https://github.com/serverlessworkflow/specification/pull/218. Please review when you get the chance.

jensg-st:
Looks good to me

PR #218, opened 2021-01-05
Data input/output schema update

tsurdilo:

- [ x] Specification
- [ x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

Updates the data schema input/output definitions


ricardozanini (approved):
Much better this way. Thanks!

manuelstein:
Existing parameter names are long, so I see grouping is nice. They could be renamed schemaInput/schemaOutput or inputSchema/outputSchema. But I guess this is up to taste. Grouping also adds a level.

I think we should add clarification on the filtering. Currently, input only validates and output seems to both validate and scrub, but I can see where scrubbing might be useful on input. What happens if the schema fails to match (is there a named error)? Are additional JSON object parameters on input data kept (optional), cut (filtered) or does it not match if there's more data?

tsurdilo:
@manuelstein good questions and btw this was requested via issue: https://github.com/serverlessworkflow/specification/issues/213

We will update all of this stuff once we decide what to do with expressions (using jq? jmespath? other?). For now this is just 
syntax update that is not update to functionality. There will be overarching updates on all this once this big question is clarified...dont worry :)

tsurdilo:
rebased

PR #226, opened 2021-01-11
Adding validUntil property to cron definition

tsurdilo:

- [ x] Specification
- [ x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

Adds "deadline" property to cron definition. Currently we only have expression which cannot alone say when 
the creation of instances according to the cron expression should end (there are use cases for this)
Deadline allows to set a hard date when the cron expression becomes obsolete.

### DIFF
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -189,6 +189,24 @@
     "states"
   ],
   "definitions": {
+    "crondef": {
+      "type": "object",
+      "properties": {
+        "expression": {
+          "type": "string",
+          "description": "Repeating interval (cron expression) describing when the workflow instance should be created",
+          "minLength": 1
+        },
+        "deadline": {
+          "type": "string",
+          "description": "Specific date and time (ISO 8601 format) when the cron expression invocation should be stopped",
+          "minLength": 1
+        }
+      },
+      "required": [
+        "expression"
+      ]
+    },
     "transition": {
       "type": "object",
       "properties": {
@@ -1821,8 +1839,7 @@
           "description":  "Time interval (ISO 8601 format) describing when the workflow starting state is active"
         },
         "cron": {
-          "type": "string",
-          "description":  "Repeating interval (cron expression) describing when the workflow starting state should be triggered"
+          "$ref": "#/definitions/crondef"
         },
         "directInvoke": {
           "description": "Defines if workflow instances can be created outside of the interval/cron interval",
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -197,9 +197,9 @@
           "description": "Repeating interval (cron expression) describing when the workflow instance should be created",
           "minLength": 1
         },
-        "deadline": {
+        "validUntil": {
           "type": "string",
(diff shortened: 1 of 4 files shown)


schema/workflow.json
```diff
@@ -203,6 +203,24 @@
...
   "definitions": {
+    "crondef": {
+      "type": "object",
+      "properties": {
+        "expression": {
+          "type": "string",
+          "description": "Repeating interval (cron expression) describing when the workflow instance should be created",
+          "minLength": 1
+        },
+        "deadline": {
```
jorgenj, line 214: Nit; I prefer a name like 'valid-until' as deadline seems a bit ambiguous to me.

specification.md
```diff
@@ -2855,6 +2859,65 @@ the needed events at the defined times to trigger workflow instance creation.
 
 The `directInvoke` property defines if workflow instances are allowed to be created outside of the defined interval or cron expression.
 
+#### Cron Definition
+
+| Parameter | Description | Type | Required |
+| --- | --- | --- | --- |
+| expression | Repeating interval (cron expression) describing when the workflow instance should be created | string | yes |
+| deadline | Specific date and time (ISO 8601 format) when the cron expression invocation should be stopped | string | no |
```
jorgenj, line 2867: I'm not sure I've experienced a need for this deadline use-case before but it makes me wonder if we should also add a corresponding attribute for 'start time' or 'valid-after'?

jorgenj (approved):

schema/workflow.json
ricardozanini: We lend the term from k8s: https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/#starting-deadline. I understand that it might not relate, and I liked `valid-until` better. @tsurdilo wdyt?

tsurdilo: @jorgenj @ricardozanini would "validUntil" be ok? That would fit our camel-case parameter names used everywhere else in lang.

specification.md
tsurdilo: @jorgenj idk what do you think? To me a cron definition does in a way define the start time (next time the cron fires after deployment). validAfter could be useful. @ricardozanini wdyt too?

jorgenj: I think there are very limited use-cases that I've seen where a 'validAfter' would've been useful but there were always easy alternatives, but since we're adding an 'end' it seemed relevant to consider adding a 'begin' as well.   My vote is probably to not add 'validAfter' unless someone actually asks for it.

tsurdilo:
rebased

schema/workflow.json
tsurdilo: updated to validUntil :)

tsurdilo:
@jorgenj @ricardozanini updated param to "validUntil". Can you guys please take another look? Thanks

ricardozanini (approved):

PR #232, opened 2021-01-14
Add ability to define reusable expression functions

tsurdilo:

- [x ] Specification
- [x ] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

Adds ability to define reusable expression functions inside the "functions" array.
Inline expression are still possible.

Also:
* update to the spec overview section
* remove the transition "expression" property as we do not define what to do in case the expression does not evaluate to true. We will address this very soon again as part of the dynamic data condition capabilities coming soon.


tsurdilo:
@ricardozanini the core part of this pr is the ability to define expression-type functions. 
Completely abstracting expressions to function definitions is not a good idea as for bigger workflows that an end up in endless amount of function definitions. Still allowing this can be huge as you can define a number of reusable expression "functions" that can be referenced all over the workflow states...

tsurdilo:
@ricardozanini this is a pre-requisite for the possible changes to jq ..just fyi.

also i ended up not changing the expression syntax. the way it is currently really works best cause we still allow inline expressions 

so to reference a reusable expression function def you would do:

{{ func(expressionFunctionName) }}

let me know if this reserved word is fine or you think something else would work better

ricardozanini:
I'm fine with it, and that looks ok to me. Not sure how AWS Lambda or MS Functions do that. Sounds reasonable. 

Here's how on AWS:

```javascript
var aws = require('aws-sdk');
var lambda = new aws.Lambda({
  region: 'us-west-2' //change to your region
});

lambda.invoke({
  FunctionName: 'name_of_your_lambda_function',
  Payload: JSON.stringify(event, null, 2) // pass params
}, function(error, data) {
  if (error) {
    context.done('error', error);
  }
  if(data.Payload){
   context.succeed(data.Payload)
  }
});
```

I guess you have to use their SDK in whatever language you're using, but I liked the `invoke` keyword. 

And on MS Functions:

```python
let request_options = {
        method: 'GET',
        host: 'my-functn-app-1.azurewebsites.net',
        path: '/path1/path2?&q1=v1&q2=v2&code=123412somecodehereemiii888ii88k123m123l123k1l23k1l3',
        headers: {
            'Content-Type': 'application/json'
        }
};
require('https')
    .request(
         request_options,
         function (res) {
               // do something here
         });
```

Just a plain HTTP call...

specification.md
```diff
@@ -421,38 +593,17 @@ operation: https://hellworldservice.api.com/api.json#helloWorld
...
-For example `https://petstore.swagger.io/v2/swagger.json#getPetById`. 
-In this example "getPetById" is the OpenAPI ["operationId"](https://swagger.io/docs/specification/paths-and-operations/)
-property in the services OpenAPI definition document which uniquely identifies a specific service operation that needs to be invoked.
+The `type` property defines the function type. Its value can be either `rest` or `expression`. Default value is `rest`.
 
-The [`metadata`](#Workflow-Metadata) property allows users to define custom information to the custom definitions.
-This allows runtimes to define services and their operations that cannot be described via OpenAPI.
-An example of such definition could be defining invocation of a command on an image, for example:
+Depending on the function `type`, the `operation` property can be:
+* If `type` is "rest", a combination of the function/service OpenAPI definition document URI and the particular service operation that needs to be invoked, separated by a '#'.
```
ricardozanini, line 601:
```suggestion
* If `type` is `rest`, a combination of the function/service OpenAPI definition document URI and the particular service operation that needs to be invoked, separated by a '#'. 
```

specification.md
```diff
@@ -421,38 +593,17 @@ operation: https://hellworldservice.api.com/api.json#helloWorld
...
-property in the services OpenAPI definition document which uniquely identifies a specific service operation that needs to be invoked.
+The `type` property defines the function type. Its value can be either `rest` or `expression`. Default value is `rest`.
 
-The [`metadata`](#Workflow-Metadata) property allows users to define custom information to the custom definitions.
-This allows runtimes to define services and their operations that cannot be described via OpenAPI.
-An example of such definition could be defining invocation of a command on an image, for example:
+Depending on the function `type`, the `operation` property can be:
+* If `type` is "rest", a combination of the function/service OpenAPI definition document URI and the particular service operation that needs to be invoked, separated by a '#'. 
+  For example `https://petstore.swagger.io/v2/swagger.json#getPetById`. 
+* If `type` is "expression", defines the expression syntax. Take a look at the [workflow expressions section](#Workflow-Expressions) for more information on this.
```
ricardozanini, line 603:
```suggestion
* If `type` is `expression`, it defines the expression syntax. Take a look at the [workflow expressions section](#Workflow-Expressions) for more information on this.
```

tsurdilo:
@ricardozanini there is a difference with the type of function definition being "rest" (default) and "expression" (new). 
the "rest" type of functions are used inside actions and as such can be used as usual to add things like timeouts, error handling, retries, etc 
with "expression" functions, those are referenced in params of string type. As such they represent a run-time of evaluation of of some data input (state data, operation results, event paylods, etc)

so no..referencing an expression-type function via {{ ... }} is not the same as invoking a restful service. hope this helps.

specification.md
```diff
@@ -106,64 +113,238 @@ Data flow during workflow execution can be divided into:
...
 
-To show some expression examples, let's say that we have the following JSON data:
+To reference a defined [expression function definition](#Using-Functions-For-Expression-Evaluation)
+the expression must have the following syntax:
+
+```text
+{{ func(definedExpressionName) }}
+```
+
+where `func` is a reserved keyword that declares that it is referencing a defined expression function, and
```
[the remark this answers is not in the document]
tsurdilo: sure thing

ricardozanini:
@tsurdilo yeah I got it. When I saw `{{ func(myExpr) }}` resembled me a function call, that's why I was trying to figure something different. The idea to have expressions being represented as function definitions to be reused in the workflows is super nice and we need this. 

Evaluating, invoking, calling them that confused me at first with the `{{ func(abc) }}`. Like we discussed off-line, and you suggested `eval`, I've found that better?

tsurdilo:
@ricardozanini thought about this for a bit and think this looks ok:

{{ expression }}

where you can do inline as already possible:

{{ $.[?(@.age >= 18)] }} 

or define reusable expr in function defs:

```json
{
"functions": [
  {
    "name": " Is Adult",
    "operation": "$.[?(@.age >= 18)]",
    "type": "expression"
  }
}
```

and use in expression as:

{{ 'Is Adult' }}

note that it is just enclosed in single quotes..this would allow you to use spaces also in your defined expression names which is useful

wdyt? yeah let's bring this up in next meet and see what others think as well

ricardozanini:
> note that it is just enclosed in single quotes, no need for some reserved keywords
   wdyt? yeah let's bring this up in next meet and see what others think as well

Super like!

My only concern is with runtime implementations. It will be a little difficult now for them to figure this as a reference to an expression. It's not impossible tasks of course, but I have to play the devil's advocate :smiling_imp:

jensg-st:
I agree with @ricardozanini it should either mandate that ALL expressions are functions or we need some kind of keyword to make implementations faster. Of course there is always the option to parse it and if it fails you look up the function or, even worse, use regular expressions to waste even more CPU cycles :).

tsurdilo:
@jensg-st @ricardozanini 
1) regarding: enforcing all expressions to be functions - sorry can't do that, as mentioned before just an action definition that calls a service with 10 params might demand 10 function definitions. For simple selections it makes more sense to do them inline, in whatever expression lang we end up picking
2) Having a keyword - we can define a keyword, just need to chose which one. i had "func(exprname)", @ricardozanini recommended "eval(exprname)". Maybe we can vote on one and go w/ it . The presented idea was that if its string surrounded by single quotes would be enough indication that its a lookup, but you guys think thats not a good idea ?

ricardozanini:
> The presented idea was that if its string surrounded by single quotes would be enough indication that its a lookup, but you guys think thats not a good idea ?

I think we can bring this topic to today's meeting and maybe vote? 'Cause adding a keyword to identify a function/expression call will make things super easy from the runtimes implementation, IMO.

tsurdilo:
@ricardozanini @manuelstein I will change back to {{ func() }} as discussed in meet. ty

tsurdilo:
@ricardozanini @manuelstein updates but wanted to make it a litte bit shorter so changed to:

{{ fn(expressionName) }}
if thats ok with you please +1 so I can merge. thanks

ricardozanini (approved):

tsurdilo:
will merge as we approved this in meeting after the made updates in this pr.

PR #242, opened 2021-01-28
Fixes #227 - Remove dataSchema property

### LINKED ISSUE #227: Validation State
**What would you like to be added:**

A new state that does nothing except validate the workflow's state data, so we can remove the `dataInputSchema` and `dataOutputSchema` fields from other states (refactor).

#### Validation State

| --- | --- | --- | --- |
| id | Unique state id | string | no |
| name | State name | string | yes |
| type | State type = "validate" | string | yes |
| stateDataFilter | State data filter | object | no |
| dataSchema | URI to JSON Schema that state data adheres to | string | yes |
| usedForCompensation | If true, this state is used to compensate another state. Default is "false" | boolean | no |
| metadata | Metadata information| object | no |
| [failure](#Next-Definition) | What to do if the data failed to validate. | object | yes |
| [success](#Next-Definition) | What to do if the data validates successfully. | object | yes |

#### Next Definition
| --- | --- | --- | --- |
| transition or end | Defines what to do next. Transition to another state, or end workflow | object | yes |

**Why is this needed:**

There's a number of parameters that optionally appear on *every* existing state because users might need them at any point during the workflow. The `dataInputSchema` and `dataOutputSchema` fields are one such example. Because they're likely to be rarely used, we can refactor the specification so that schema validation is a separate and very explicit step in a workflow. Then we could simplify the spec for all existing states by removing these fields. 

No functionality would be lost as a result of this change. 

The [Next Definition](#Next-Definition) keeps things a little bit tidier. It's necessary to be able to support both transition *and* end because a validation makes sense as the final state when writing subflows.

tsurdilo:

- [x] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

Removes the dataschema property. This was discussed in Issue 
https://github.com/serverlessworkflow/specification/issues/227
In end we decided that its best for it not to be in spec until requested and then we should make sure its 
clearly understood

### DIFF
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -30,9 +30,6 @@
           "description": "Serverless Workflow schema version",
           "minLength": 1
         },
-        "dataSchema": {
-          "$ref": "#/definitions/dataschema"
-        },
         "execTimeout": {
           "$ref": "#/definitions/exectimeout"
         },
@@ -125,9 +122,6 @@
           "type": "string",
           "description": "Serverless Workflow schema version",
           "minLength": 1
-        },
-        "dataSchema": {
-          "$ref": "#/definitions/dataschema"
         },
         "execTimeout": {
           "$ref": "#/definitions/exectimeout"
@@ -525,9 +519,6 @@
           "description": "Next transition of the workflow after the time delay",
           "$ref": "#/definitions/transition"
         },
-        "dataSchema": {
-          "$ref": "#/definitions/dataschema"
-        },
         "compensatedBy": {
           "type": "string",
           "minLength": 1,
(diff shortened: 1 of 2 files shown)


ricardozanini (approved):

tsurdilo:
rebased

tsurdilo:
Approved in our team meet today - https://docs.google.com/document/d/1xwcsWQmMiRN24a7o7oy9MstzMroAup31oOkM5Dru1jQ/edit#

PR #243, opened 2021-01-28
Closes #239: Add SubFlow state repeat ability

### LINKED ISSUE #239: Need repeat execution of state
What would you like to be added:
Hope to support repeat execution of state.

Why is this needed:
In some cases, we need to run some states repeatedly for x times (without parameter, can't use ForEach). Although we can put the same action in x states to resolve this scenario, it is not convenient.

tsurdilo:

- [x] Specification
- [x] Schema
- [x] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

Adds SubFlow state "repeat" ability and adds example


tsurdilo:
rebased

examples/README.md
```diff
@@ -3241,3 +3242,253 @@ functions:
...
 </tr>
 </table>
+
+### Car Vitals Checks
+
+#### Description
+
+In this example we want to automate a workflow which checks all our car vital signs while our car is driving. 
+The workflow should start when we receive the "carOn" event and stop when the "carOff" event is consumed.
+While the car is driving our workflow should repeatedly check the vitals every 2 minutes.
```
ricardozanini, line 3252:
```suggestion
In this example, we want to automate a workflow that checks all our car vital signs while driving. 
The workflow should start when we receive the "carOn" event and stop when the "carOff" event is consumed.
While the car is driving, our workflow should repeatedly check the vitals every 2 minutes.
```

specification.md
```diff
@@ -2419,16 +2420,24 @@ Reusable workflow are referenced by their `id` property via the SubFlow states`w
...
 
-Referenced workflows must declare their own [function](#Function-Definition) and [event](#Event-Definition) definitions.
+The `repeat` property defines the SubFlow states repeated execution (looping) behavior. This allows you to specify that 
+the sub-workflow should be executed multiple times repeatedly.
+If the `repeat` property is defined, the `waitForCompletion` should be assumed have the value of `true`.
+If the workflow explicitly triggers [compensation](#Workflow Compensation) and the SubFlow state 
+was executed and defines its compensation state, it should be compensated once, no matter how many times
+its exection was executed as defined by the `repeat` property.
+After each execution of the SubFlow state, if `repeat` is defined, the SubFlow state data at the end of the 
+one execution should become the state data of the next execution.
```
ricardozanini, line 2436:
```suggestion
The `repeat` property defines the SubFlow states repeated execution (looping) behavior. This allows you to specify that 
the sub-workflow should be executed multiple times repeatedly.
If the `repeat` property is defined, the `waitForCompletion` should be assumed to have the value of `true`.
If the workflow explicitly triggers [compensation](#Workflow Compensation) and the SubFlow state 
was executed and defines its compensation state, it should be compensated once, no matter how many times
its execution was executed as defined by the `repeat` property.
After each execution of the SubFlow state, if `repeat` is defined, the SubFlow state data at the end of the 
one execution should become the state data of the next execution.
```

specification.md
```diff
@@ -2969,6 +2978,81 @@ The Callback state `timeout` property defines a time period from the action exec
...
+If `continueOnError` is set to `false` (default value), and an unhandled error occurs, it should be handled 
+as any other unhandled workflow error, meaning repeat execution shall stop and workflow should stop its exception.
+
+If an error occurs which propagates to the SubFlow state, and is handled explicitly by the 
+SubFlow states [`onErrors`](#Error-Definition) definition, the control flow must take the path of the error handling definition
+and repeat execution must halt.
+
+An alternative way to limit repeat executions is via the `stopOnEvents` property. It contains a list of one or more 
+defined consumed workflow events (referened by the unique event name). When `stopOnEvents` is defined,
+SubFlow will repeat execution until one of the defined events is consumed, or until the max property count is reached.
```
ricardozanini, line 3051:
```suggestion
The `checkBefore` property can be used to decide if the `expression` evaluation should be done before or after each SubFlow state execution. The default value of this property is `true`.

The `max` property sets the maximum count of repeat executions. It should be a positive integer value.
Runtime implementations must define an internal repeat/loop counter, which is incremented for each of the 
SubFlow state repeated executions. If this counter reaches the max value, repeated executions should end.

The `continueOnError` property defines if repeated executions should continue or not if the sub-workflow propagates unhandled errors to the SubFlow state. The default value of this property is `false`.
Unhandled errors are errors that are not explicitly handled by the sub-workflow, and the SubFlow state 
via its [`onErrors`](##Error-Definition) definition.

If `continueOnError` is set to `false` (default value), and an unhandled error occurs, it should be handled as any other unhandled workflow error, meaning repeat execution shall stop and workflow should stop its exception.

If an error occurs which propagates to the SubFlow state and is handled explicitly by the 
SubFlow states [`onErrors`](#Error-Definition) definition, the control flow must take the path of the error handling definition, and repeat execution must halt.

An alternative way to limit repeat executions is via the `stopOnEvents` property. It contains a list of defined consumed workflow events (referenced by the unique event name). When `stopOnEvents` is defined, SubFlow will repeat execution until one of the defined events is consumed or until the max property count is reached.
```

ricardozanini (approved):
Feel free to reject the suggestions. :)

ricardozanini (approved):

tsurdilo:
rebased

Approved in our team meet today: https://docs.google.com/document/d/1xwcsWQmMiRN24a7o7oy9MstzMroAup31oOkM5Dru1jQ/edit#

PR #249, opened 2021-02-09
Replacing jsonpath with jq and adding 'expressionLang' top level property

### LINKED ISSUE #216: Replace JSONPath with 'jq'

Replace JSONPath with [`jq`](https://stedolan.github.io/jq/) for all [Workflow Data Expressions](https://github.com/serverlessworkflow/specification/blob/master/specification.md#Workflow-Data-Expressions).

**Why is this needed**:

1. JSONPath lacks power and flexibility.

JSONPath was designed to query JSON documents for matches, especially arrays. It wasn't designed to handle complex logic, and the limitations in the spec show. JSONPath's inability to reference parent objects or property names of matching items forces users to contort their data in all sorts of strange ways. 

For example, I wanted to put together a simple workflow that transforms an input from 

```json 
{
	"name": "Alan"
}
```

to 

```json 
{
	"greeting": "Hello, Alan!"
}
```

Here's my solution:

```yaml 
id: helloworld
name: "Hello, World!"
states:
- name: Hello
  type: inject
  data:
    output:
      greeting: "Hello, {{ $.name }}!"
  start:
    kind: default
  end:
    kind: default
  stateDataFilter:
    dataOutputPath: "$.output"
```

It's possible, but it's convoluted. And that's for a simple helloworld. I had to nest my solution in a sub-property named output just so that I could use the stateDataFilter to replace the entire state with the constructed output afterwards. 

If we'd been using `jq` instead of JSONPath, that same solution could look like this:

```yaml
id: helloworld
name: "Hello, World!"
states:
- name: Hello
  type: inject
  start:
    kind: default
  end:
    kind: default
  stateDataFilter:
    dataOutputPath: '{ greeting: ("Hello, " + .name + "!") }'
```

We wouldn't even need the functionality of the Inject State to pull it off. In fact, we could do all of this and more using just stateDataFilters on any state, dramatically simplifying the code.

2. Language Support 

These limitations are well known about JSONPath, which has resulted in a fork of the JSONPath spec called [JSONPath Plus](https://www.npmjs.com/package/jsonpath-plus), which gets >25% more weekly downloads than [the original](https://www.npmjs.com/package/jsonpath).

Unfortunately, JSONPath does not have great language support. JSONPath Plus has even less. For Go I found four different libraries for the former and none for the latter. Of the four libraries, not one produced the correct result in every test I tried over the course of an afternoon. 

By comparison `jq` is relatively less popular amongst JavaScript developers, but seems to have implementations where it matters: languages that will be used to build implementations. 

* [nodejs](https://www.npmjs.com/package/node-jq)
* [go](https://github.com/itchyny/gojq)
* [python](https://pypi.org/project/jq/) [python](https://pypi.org/project/pyjq/)
* [java](https://github.com/eiiches/jackson-jq)

3. JSON Patch 

In my earlier example, nesting the output was necessary to filter it later (due to the aforementioned limitations of JSON Path). But the reason I needed to filter it later is because my requirement was to *replace* the data, not enrich it. 

The current specification doesn't have any good way of dealing with the issue of deleting fields, which has resulted in discussions like [this one](https://github.com/serverlessworkflow/specification/issues/149) where bolting on another solution (JSON Patch) is being considered. 

In [an example from that discussion](https://github.com/serverlessworkflow/specification/issues/149#issuecomment-729217479),  input contains two properties that should be removed (`vegetables` and `snacks`):

```json
{
  "fruits": [ "apple", "orange", "pear" ],
  "drinks": ["capirinha", "whisky", "vodka", "wine"],
  "snacks": ["french fries", "jelly beans"],
  "vegetables": [
    {
      "veggieName": "potato",
      "veggieLike": true
    },
    {
      "veggieName": "broccoli",
      "veggieLike": false
    }
  ]
}
```

The proposed JSON Patch solution is to include the following verbose snippet in the state definition:

```yaml
stateDataTransform:
  dataInputPatch:
  - op: remove
    path: "/snacks"
  - op: remove
    path: "/vegetables"
```

In `jq`, this same problem can be tackled with this concise line in the output filter:

```yaml
dataOutputPath: 'del(.vegetables) | del(.snacks)'
```

JSONPath is only capable of returning arrays and/or subsets of data. JSON Patch can build data, but it's a laborious process. `jq` isn't just for searching or querying data, it can also be used to construct, restructure, combine, or transform data. 

4. Adoption

Why combine two solutions to get less functionality? Especially when `jq` can not only do it all, it has an enormous wealth of users and help information readily available? Hundreds of Stack Overflow questions answered. 

**Summary** 

Compared to JSON Path, `jq` usually produces cleaner workflow definitions, it has greater language support, and it is much easier for users to figure out. All while being far more powerful. In fact, `jq` is so powerful that it would render the Inject State obsolete.

tsurdilo:

- [x] Specification
- [x] Schema
- [x] Examples
- [x] Extensions
- [x] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

* replaces jsonpath with jq
* updates expressions syntax

Fixes #216


comparisons/comparison-google-cloud-workflows.md
```diff
@@ -228,7 +228,7 @@ The second step, "exit-loop" is then used alongside the "return" keyword to spec
 workflow results.
 With Serverless Workflow we can inject the array data via the "inject" state again, or 
 it can simply be passed as workflow data input. There is no need for looping here as
-we can just utilize the JsonPath "concat" method as shown in the states data filter.
+we can just utilize the [jq "join" function](https://stedolan.github.io/jq/manual/) as shown in the states data filter.
```
ricardozanini, line 231:
```suggestion
we can just utilize the [jq "join" function](https://stedolan.github.io/jq/manual/#join(str)) as shown in the states data filter.
```

ricardozanini (approved):

specification.md
```diff
@@ -1942,7 +1960,7 @@ If events defined in event-based conditions do not arrive before the states `eve
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
 | name | Data condition name | string | no |
-| condition | JsonPath expression evaluated against state data. True if results are not empty | string | yes |
+| condition | jq expression evaluated against state data. Must evaluate to true or false | string | yes |
```
jorgenj, line 1963: According to the jq docs:

> false and null are considered "false values", and anything else is a "true value".

Should this follow that standard?  If not, what should happen to a workflow where some non-boolean value is returned by the filter.

specification.md
```diff
@@ -2677,10 +2715,10 @@ Each iteration of the ForEach state should be executed in parallel.
 You can use the `max` property to set the upper bound on how many iterations may run in parallel. The default
 of the `max` property is zero, which places no limit on number of parallel executions.
 
-The `inputCollection` property is a JsonPath expression which selects an array in the states data. All iterations 
-are done against data elements of this array. This array must exist.
+The `inputCollection` property is a workflow expression which selects an array in the states data. All iterations 
+are performed against data elements of this array. This array must exist.
```
jorgenj, line 2719: Should the workflow fail or throw an exception if the array does not exist?  Or can we say the runtime should assume an empty array if the inputCollection would return a non-array value?

specification.md
```diff
@@ -286,12 +286,12 @@ Our expression function definitions can now be referenced by workflow states whe
      "dataConditions": [
         {
           "name": "Applicant is adult",
-          "condition": "{{ fn(isAdult) }}",
+          "condition": "${ fn:isAdult }",
```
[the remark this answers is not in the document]
tsurdilo: yes, we need a language specific mapping to tell parsers/validators that its a call to an expression functions so they substitute its "operation" value that includes the actual expression that needs to be evaluated. "fn" is just a namespace in this case.

the data the expression is evaluated against depends on if its used inside a certain filter, in a state property or in a top-level workflow property. 
I will look at possibility to pass in the input specifically, for example "fn:isAdult(.customer)". Will test it out and do that in a separate pr if possible. Right now its assumed its always "fn:expressionFunctionName(.)" but users dont have to do the "(.)" part - is assumed

specification.md
```diff
@@ -313,38 +313,39 @@ For more information about workflow expressions, reference the [Workflow Express
...
-Where `expression` can be either an in-line expression using the  [JsonPath](https://github.com/json-path/JsonPath) format
+Where `expression` can be either an in-line expression using the [jq](https://stedolan.github.io/jq/) syntax,
 or a reference to a defined [expression function definition](#Using-Functions-For-Expression-Evaluation).
 
 To reference a defined [expression function definition](#Using-Functions-For-Expression-Evaluation)
 the expression must have the following format, for example:
 
 ```text
-{{ fn(myExprName) }}
+${ fn:myExprFuncName }
```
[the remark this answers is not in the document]
tsurdilo: i dont think its something we should enforce (exposing function defs with type "expression" as jq functions). Some runtimes just might want to just substitute the operation value of the function def that includes the expression and just eval it.

specification.md
tsurdilo: Good point. I think it would make more sense then to say
"jq expression evaluated against state data. Condition is considered as true if the rest of the expression is true. If the result of the expression is anything else, the condition is considered false."

specification.md
tsurdilo: if inputcollection is empty or does not exist there is nothing to loop over. i think this should be a runtime error and should be able to be handled in the onErrors def. will update to add that part

tsurdilo:
@jorgenj @ricardozanini
take a look at https://github.com/serverlessworkflow/specification/issues/250
enforcing a single expression lang seems afterall not feasible for real world use. i have added the needed updates
to this pr (added expressionLang) property back where users can define a unique identifier of the expression language 
used in workflow expressions (default is "jq").

ricardozanini:
Adding such `expression type` attribute weren't we opening the gates of hell? Having an extension to handle this kind of scenario wouldn't be better? So users could import the extension they need and at the same time reuse the same workflow between runtimes.

tsurdilo:
not sure extension is needed as all expressions are inside workflow definition. we had this before and removed it to try to enforce portability as expression syntax has affects on portability..but it seems that pushing a single language only does significantly lower the adoption possibilities, as many runtimes have hard requirements to use a specific syntax for expressiosn which is not our "default" one.

specification.md
```diff
@@ -319,17 +319,28 @@ Note that different data filters play a big role as to which parts of the states
...
-All expressions must follow the [jq](https://stedolan.github.io/jq/) syntax. Jq is a very powerful JSON processor. You 
-can find more information in the [jq manual](https://stedolan.github.io/jq/manual/).
+By default, all workflow expressions should be defined using the [jq](https://stedolan.github.io/jq/) syntax.
+You can find more information on jq in its [manual](https://stedolan.github.io/jq/manual/).
 
-Expressions have the following format:
+Serverless Workflow does not mandate the use of jq and it's possible to use an expression language 
+of your choice with the restriction that a single one must be used for all expressions
+in a workflow definition. If a different expression language needs to be used, make sure to set the workflow 
+`expressionLang` property to identify it to runtime implementations.
```
manuelstein, line 328: call on 15/02: Do we need to distinguish transformationLang from conditionLang?
jq can do transformation (outputPath, etc.), but when we looked at JSONPath and JSONPatch, path was able to list found elements in an array while JSONPatch was able to make modifications (add/replace/remove/move/copy/...) and we didn't discuss languages that are just for writing conditions, but there are some. We need conditions for transitions, switch statements, etc and we need transformations for input/output path. If I specify expressionLang to be SQL, I wouldn't be able to do any transformations. And for conditions, it needs to be clear how to interpret the result (e.g. if an empty/nil result means false or if it's an error).

specification.md
```diff
@@ -319,17 +319,28 @@ Note that different data filters play a big role as to which parts of the states
...
+Workflow expressions have the following format:
 
 ```text
 ${ expression }
 ```
 
-Where `expression` can be either an in-line expression using the [jq](https://stedolan.github.io/jq/) syntax,
-or a reference to a defined [expression function definition](#Using-Functions-For-Expression-Evaluation).
+Where `expression` can be either an in-line expression, or a reference to a 
+defined [expression function definition](#Using-Functions-For-Expression-Evaluation).
```
manuelstein, line 343: Need to add how to escape in case `"${ notanexpression }"` is supposed to not be an expression, but a string. Suggesting to escape the dollar sign `\$`.

manuelstein: An alternative could be to use an object instead of a string, i.e. leave string as string and have a type `StringOrExpression` for e.g. operation parameters, that allows either a string or an object with the parameter `expression` and an optional language:
```
parameters:
  firstParam: "StaticText"
  secondParam:
    expression: "${ .message }"
  thirdParam:
    language: "JSONPath"
    expression: "{{ .message }}"
```
or in designing the object, we could allow the language to be the parameter name:
```
parameters:
  firstParam: "StaticText"
  secondParam:
    jq: "${ .message }"
  thirdParam:
    jsonpath: "{{ .message }}"
```
I'd like the last one most. But feel free to drop it now and we can add this in a follow-up PR

specification.md
```diff
@@ -1960,7 +1976,7 @@ If events defined in event-based conditions do not arrive before the states `eve
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
 | name | Data condition name | string | no |
-| condition | jq expression evaluated against state data. Must evaluate to true or false | string | yes |
+| [condition](#Workflow-Expressions) | Workflow expression evaluated against state data. Must evaluate to true or false | string | yes |
```
manuelstein, line 1979: For conditions, we need to have bindings on expression languages, meaning we need to specify how to use them. Any third-party extension is fine, but we need to say for jq that we want the result of the expression to be either JSON `true` or JSON `false` and that nil causes a runtime error.

manuelstein: also feel free to postpone this in an issue

comparisons/comparison-brigade.md
```diff
@@ -403,18 +403,18 @@ states:
...
     eventDataFilter:
-      dataOutputPath: "{{ $.event }}"
+      dataOutputPath: "${ .event }"
     actions:
     - name: eventInfoAction
       functionRef:
         refName: consoleFunction
         parameters:
-          log: ">>> event $event.type caused by $.event.data.provider"
+          log: ">>> event ${ .event.type } caused by ${ .event.data.provider }"
```
manuelstein, line 412: String formatting seems like a third kind of use for expression languages. It looks familiar, is there some common string formatting in e.g. helm templating that can be used?

manuelstein (approved):
not actually asking for changes, but a couple of questions on the PR to continue the discussion

specification.md
tsurdilo: I'm not sure we can tell users which parts of jq to use at what places, not sure how to honestly enforce that (even on the runtime level). 
a selection in case of our data filters is really also a modification, for example a state filter data output becomes the final output of that state, so by selection you are really modifying the state data that for example is passed to the next state. 

i think for conditions we should just say they are boolean conditions and leave it at that

specification.md
tsurdilo: in order to model the second scenario we would have to have a defined enum of "possible" expression languages in our schema. don't think we really wanna go that direction as we will never be able to keep up with it and support all adopters

i would like to see the case where atm defining more than one expression language would be needed. i think best maybe is for now to have the top-level expressionLang and to keep enforcing a single expression lang throughout the workflow definition until we have someone bring up a good use case where multiple ones would be needed. wdyt

specification.md
tsurdilo: jq can return true/false i would assume that anything else than true could be interpreted as false. so imo null/nil would be false in evaluation. i think other types of exceptions during expression evaluation should also just be handled as false in this case

comparisons/comparison-brigade.md
tsurdilo: sorry idk much about helm. could you show example so i understand what this means?

tsurdilo:
rebased

specification.md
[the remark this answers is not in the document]
jorgenj: I think what bothers me about this is we need to inject special behavior into the middle of the jq expressions in order to call these functions, it's no longer JQ at this point but some strange hybrid language.  This puts a burden on users of the workflow language because they have to learn this new expression language special to calling workflow functions, and I worry that it actually makes the runtimes more difficult to implement.

It seems like if we have the expectation that all of the workflows expression 'functions' are automatically exposed as jq functions by the runtime, then we shouldn't need a special syntax to tell the runtime to 'substitute' the call.  The caveat would be that, if some of the expression functions aren't jq then they would not be accessible from jq expressions (which imho, is a reasonable approach).

Given the example above, it seems like the runtime under-the-covers could define the expression functions as jq functions before evaluating a condition.

Given this alternative way to define this line:
``` "condition": "${isAdult(.)}"```

Then runtime could do something functionally equivalent to the follow to evaluate the expression:

```
def isAdult: .applicant | .age >= 18;
def isMinor: in.applicant | .age < 18;
isAdult(.)
```

**Comment by tsurdilo** [comment]
need to rebase :)

File: specification.md (hunk 8 above)
Comment (jorgenj) [reply specification.md:2030]: According to the JQ docs:

> false and null are considered "false values", and anything else is a "true value".

It seems like this would be really confusing for users of the workflow language to have to bridge that gap.  For example, they'd have to take extra special care for a pretty common/basic scenario like the following:

Input data:
```
{ "aNullableAgeField": null }
```

Given input like the above, a workflow might want to do something special if the user is a minor (or didn't provide their age).  The natural way to specify this in JQ would be:
```
"functions": [
  {
    "name": "isMinorOrNoAgeProvided",
    "operation": "${.aNullableAgeField < 18}",
    "type": "expression"
  }
]
```

Unfortunately that would return null for the sample input above, which is suggested to cause a runtime exception.

Workflow users would have to add a lot of boiler-plate to their expressions to guard against this, like so:
```
"functions": [
  {
    "name": "isMinorOrNoAgeProvided",
    "operation": "${.aNullableAgeField == null OR .aNullableAgeField < 18}",
    "type": "expression"
  }
]
```

tsurdilo:
will go ahead and merge - @jorgenj if you find any issues w/ this please raise them and i'll fix asap

specification.md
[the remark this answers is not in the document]
tsurdilo: ugh sorry i went ahead and merged before realizing to see this comment
i will open an issue on this now and work on it for next week

specification.md
tsurdilo: what would be the suggestion to fix this?

jorgenj: Since we're targeting support for many expression lang choices, perhaps we can just mandate that the runtime is responsible for translating any potential value to a concrete true or false based on the standard behavior of that expression lang?   For the JQ example, this means that the runtime should translate according to JQ behavior?  Basically, we set the expectation that for any boolean expression, it's important that the runtime ensures that they properly translate any potential value from the user-defined expression to return exactly true or false and never any other value?

tsurdilo: +1
@jorgenj could you rewrite the current text in the spec to express what you stated here please?

PR #260, opened 2021-02-18
Simplify schedule and cron definitions

tsurdilo:

- [ x] Specification
- [ x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

simplifies the "cron" definition. if you dont need to specify the "validUntil" param you can now use it as a string type, for example;

```json
"cron": "0 15,30,45 * ? * *"
```

simplifies the "schedule' definition. if you dont need to specify cron, directiinvoke and timezone (so only 'interval') you can do:

```json
"schedule": "2020-03-20T09:00:00Z/2020-03-20T15:00:00Z"
```

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -2208,9 +2208,7 @@ The results of the inbox service called is expected to be for example:
         "type": "operation",
         "start": {
             "schedule": {
-                "cron": {
-                   "expression": "0 0/15 * * * ?"
-                }
+                "cron": "0 0/15 * * * ?"
             }
         },
         "actionMode": "sequential",
--- a/examples/README.md
+++ b/examples/README.md
@@ -2082,9 +2082,7 @@ Bidding is done via an online application and bids are received as events are as
           "name": "StoreCarAuctionBid",
           "type": "event",
           "start": {
-              "schedule": {
-                "interval": "2020-03-20T09:00:00Z/2020-03-20T15:00:00Z"
-              }
+              "schedule": "2020-03-20T09:00:00Z/2020-03-20T15:00:00Z"
           },
           "exclusive": true,
           "onEvents": [
@@ -2125,8 +2123,7 @@ states:
 - name: StoreCarAuctionBid
   type: event
   start:
-    schedule:
-      interval: 2020-03-20T09:00:00Z/2020-03-20T15:00:00Z
+    schedule: 2020-03-20T09:00:00Z/2020-03-20T15:00:00Z
   exclusive: true
   onEvents:
   - eventRefs:
@@ -2258,8 +2255,7 @@ states:
   type: operation
   start:
     schedule:
-      cron:
-        expression: 0 0/15 * * * ?
+      cron: 0 0/15 * * * ?
   actionMode: sequential
   actions:
   - functionRef: checkInboxFunction
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -1777,35 +1777,45 @@
       ]
     },
     "schedule": {
-      "type": "object",
-      "description": "Start state schedule definition",
-      "properties": {
-        "interval": {
-          "type": "string",
-          "description": "Time interval (ISO 8601 format) describing when workflow instances can be created."
-        },
-        "cron": {
-          "$ref": "#/definitions/crondef"
-        },
-        "directInvoke": {
-          "description": "Define if workflow instances can be created outside of the defined interval/cron",
-          "type": "boolean",
-          "default": false
-        },
-        "timezone": {
-          "type": "string",
-          "description": "Timezone name used to evaluate the cron expression. Not used for interval as timezone can be specified there directly. If not specified, should default to local machine timezone."
-        }
-      },
       "oneOf": [
         {
-          "required": [
-            "interval"
-          ]
+          "type": "string",
+          "description": "Time interval (ISO 8601 format) describing when workflow instances can be created.",
+          "minLength": 1
         },
         {
-          "required": [
-            "cron"
+          "type": "object",
+          "description": "Start state schedule definition",
+          "properties": {
+            "interval": {
+              "type": "string",
+              "description": "Time interval (ISO 8601 format) describing when workflow instances can be created.",
+              "minLength": 1
+            },
+            "cron": {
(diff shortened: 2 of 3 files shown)


ricardozanini (approved):

specification.md
```diff
@@ -3146,6 +3143,17 @@ the needed events at the defined times to trigger workflow instance creation.
 
 #### Schedule Definition
 
+`Schedule` definition can have two types, either `string` or `object`.
+If `string` type, it defines time interval describing when the workflow instance can be created.
```
jorgenj, line 3147: I don't really understand why this 'interval' property exists and why it's related to the 'schedule'.  I can see 2 potential use-cases:
a) The ability to define a 'validity' period for a workflow, which limits the time period within which it's allowed to be launched.  Outside of that time period, the runtime would reject launch requests for this workflow.
b) The ability to define an 'interval' for periodic launches instead of cron.  For example, run the workflow every 7 minutes, or perhaps run it 7 minutes after the last time it completed.

For use-case A, it seems like we should elevate this parameter outside of the 'schedule' since it seems more generally useful than just for limiting schedules.  My personal opinion is that this feature isn't necessary, since it doesn't feel super common and a workflow could enforce this itself by checking the time period in it's first step and ending early.

For use-case B, I could see customers wanting something other than cron but I'm not sure if that has been a requested feature or not yet.

examples/README.md
```diff
@@ -2082,9 +2082,7 @@ Bidding is done via an online application and bids are received as events are as
           "name": "StoreCarAuctionBid",
           "type": "event",
           "start": {
-              "schedule": {
-                "interval": "2020-03-20T09:00:00Z/2020-03-20T15:00:00Z"
-              }
+              "schedule": "2020-03-20T09:00:00Z/2020-03-20T15:00:00Z"
```
jorgenj, line 2085: This is confusing because the field is "schedule" but that's not a schedule, it's a time interval...

tsurdilo: it can also be a time repeating interval. would it make more sense to use that for the examples?

schema/workflow.json
```diff
@@ -1767,35 +1777,45 @@
...
-          "description": "Timezone name used to evaluate the cron expression. Not used for interval as timezone can be specified there directly. If not specified, should default to local machine timezone."
-        }
-      },
       "oneOf": [
         {
-          "required": [
-            "interval"
-          ]
+          "type": "string",
+          "description": "Time interval (ISO 8601 format) describing when workflow instances can be created.",
```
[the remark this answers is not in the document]
tsurdilo: schedule def currently has properties "interval" and "cron". cron was added after interval so i just picked that. we can switch short hand to cron if you think its more appropriate

schema/workflow.json
```diff
@@ -1767,35 +1777,45 @@
...
+          "properties": {
+            "interval": {
+              "type": "string",
+              "description": "Time interval (ISO 8601 format) describing when workflow instances can be created.",
+              "minLength": 1
+            },
+            "cron": {
+              "$ref": "#/definitions/crondef"
+            },
+            "directInvoke": {
```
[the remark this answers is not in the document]
tsurdilo: iirc we added directInvoke specifically when we added cron definition, it was requested (by Kay if I'm not mistaken) as it would indicate if a cron-based workflow could be triggered explicitly outside of the cron definition

specification.md
tsurdilo: You can write a schedule w/ it imo, for example:

R/0402T/P1Y is a repeating interval of "repeat every year on april second"

or

R/20210401/P1M is repeating interval of "repeat every month on first of month starting with april 1st"

is that not a schedule?

tsurdilo:
@jorgenj interval defines the schedule (can also be repeating) on when workflow instances can be created. if a requested creation of a workflow instance does not fall under the interval described it cannot be created.

tsurdilo:
@jorgenj if you feel like we need to change this stuff please feel free to do so (like externalizing some of this into independent props etc)

examples/README.md
jorgenj: Yes, that would probably make more sense for the examples.  Is there a way to restrict that it can only be a repeating interval?  Supporting a time range like the above seems to cause a lot of confusion on what that would mean.

specification.md
jorgenj: Got ya, I was confused because all the examples are shown as a time range, which it's hard to understand how that works as a schedule.

examples/README.md
tsurdilo: +1
Good point. I think also we should restrict in text.

schema/workflow.json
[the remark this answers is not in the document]
jorgenj: I'll speak with Kay about that, but I don't think it makes sense to have it at this level.  Really, I question whether we should even have a flag to disable directInvoke at all.  What do you think about completely dropping that?

Since this PR is already merged and I'm late to respond, would you prefer if I send a follow-up PR for the following proposal?

a) remove directInvoke flag (or alternatively move it out of the 'schedule' definition?)
b) Change the examples to use repeating interval like you have in your other comment below
c) Change the description of interval to say that it must be a repeating interval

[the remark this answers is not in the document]
tsurdilo: sure, go for it. might want to do after we decide on https://github.com/serverlessworkflow/specification/pull/266 tho as that moves the start definition from workflow states to top-level prop (just saying so you dont have to rebase later in case that one gets merged first).

and +1 for removing directInvoke

schema/workflow.json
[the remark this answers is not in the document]
jorgenj: Ah, I see that I was very confused by the purpose of interval, sorry for the confusion here.  If we clear up the confusion to require that 'interval' must be an ISO 'repeating interval' then it seems reasonable to me to have the string form of schedule be an interval.   

I think how we choose which one is supported for the string form of interval, we'd ideally want to pick the most used option, but alas there's no way for us to know which is likely to be more used so we either pick one arbitrarily or don't support the string form at all.   Interval seems like a fine choice.

PR #265, opened 2021-02-23
[WIP] updated error handling

jplane:

- [X] Specification
- [X] Schema
- [X] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

This PR updates error handling behavior to replace error/code with a single 'condition' predicate that matches against state data to select an error handler. The first matching error handler is the "winner".

This is a more general purpose mechanism than using 'error' and 'code' properties, which may or may not be directly supported by any given runtime and its error semantics.

I'm not an expert with jq syntax, so I'd appreciate extra attention to the jq predicate examples I've provided, to ensure correctness. Thanks for that.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -890,21 +890,18 @@ The data output of the workflow contains the information of the exception caught
           }
        }
     ],
-    "stateDataFilter": {
-       "dataOutputPath": "${ .exceptions }"
-    },
     "transition": "ApplyOrder",
     "onErrors": [
        {
-         "error": "Missing order id",
+         "condition": "${ .exceptions | .[] | .error == 'Missing order id' }",
          "transition": "MissingId"
        },
        {
-         "error": "Missing order item",
+         "condition": "${ .exceptions | .[] | .error == 'Missing order item' }",
          "transition": "MissingItem"
        },
        {
-        "error": "Missing order quantity",
+         "condition": "${ .exceptions | .[] | .error == 'Missing order quantity' }",
         "transition": "MissingQuantity"
        }
     ]
@@ -962,11 +959,11 @@ states:
     dataOutputPath: "${ .exceptions }"
   transition: ApplyOrder
   onErrors:
-  - error: Missing order id
+  - condition: ${ .exceptions | .[] | .error == 'Missing order id' }
     transition: MissingId
-  - error: Missing order item
+  - condition: ${ .exceptions | .[] | .error == 'Missing order item' }
     transition: MissingItem
-  - error: Missing order quantity
+  - condition: ${ .exceptions | .[] | .error == 'Missing order quantity' }
     transition: MissingQuantity
 - name: MissingId
   type: subflow
@@ -1064,10 +1061,10 @@ In the case job submission raises a runtime error, we transition to a SubFlow st
       }
       ],
       "onErrors": [
-      {
-        "error": "*",
-        "transition": "SubmitError"
-      }
+        {
+          "condition": "${ .error }",
+          "transition": "SubmitError"
+        }
       ],
       "stateDataFilter": {
           "dataOutputPath": "${ .jobuid }"
@@ -1192,7 +1189,7 @@ states:
     actionDataFilter:
       dataResultsPath: "${ .jobuid }"
   onErrors:
-  - error: "*"
+  - condition: ${ .error }
     transition: SubmitError
   stateDataFilter:
     dataOutputPath: "${ .jobuid }"
@@ -2695,8 +2692,7 @@ If the retries are not successful, we want to just gracefully end workflow execu
       ],
       "onErrors": [
         {
-          "error": "ServiceNotAvailable",
-          "code": "503",
+          "condition": "${ .error.message == 'ServiceNotAvailable' and .error.code == 503 }",
           "retryRef": "ServicesNotAvailableRetryStrategy",
           "end": true
         }
@@ -2754,8 +2750,7 @@ states:
     - functionRef: AssignDoctor
     - functionRef: ScheduleAppt
   onErrors:
-  - error: ServiceNotAvailable
-    code: '503'
+  - condition: ${ .error.message == 'ServiceNotAvailable' and .error.code == 503 }
     retryRef: ServicesNotAvailableRetryStrategy
     end: true
   end: true
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -193,14 +193,9 @@
     "error": {
       "type": "object",
       "properties": {
-        "error": {
+        "condition": {
           "type": "string",
-          "description": "Domain-specific error name, or '*' to indicate all possible errors",
-          "minLength": 1
-        },
-        "code": {
-          "type": "string",
-          "description": "Error code. Can be used in addition to the name to help runtimes resolve to technical errors/exceptions. Should not be defined if error is set to '*'",
+          "description": "Predicate JQ expression to match this error handler against incoming state data",
           "minLength": 1
         },
         "retryRef": {
@@ -220,13 +215,13 @@
       "oneOf": [
         {
           "required": [
-            "error",
+            "condition",
             "transition"
           ]
         },
         {
           "required": [
-            "error",
+            "condition",
             "end"
           ]
         }
--- a/specification.md
+++ b/specification.md
@@ -1622,8 +1622,7 @@ to the trigger/produced event.
 
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
-| error | Domain-specific error name, or '*' to indicate all possible errors | string | yes |
-| code | Error code. Can be used in addition to the name to help runtimes resolve to technical errors/exceptions. Should not be defined if error is set to '*' | string | no |
+| condition | jq expression to resolve against input to error-handling logic | string | yes |
 | retryRef | Defines the unique retry strategy definition to be used | string | no |
 | [transition](#Transitions) or [end](#End-Definition) | Transition to next state to handle the error, or end workflow execution if this error is encountered | object | yes |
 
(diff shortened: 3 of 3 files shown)


tsurdilo:
@jplane the DCO is failing cause the commit is not signed. Could you please sign your commits? Thanks

examples/README.md
```diff
@@ -890,21 +890,18 @@ The data output of the workflow contains the information of the exception caught
...
        }
     ],
-    "stateDataFilter": {
-       "dataOutputPath": "${ .exceptions }"
-    },
     "transition": "ApplyOrder",
     "onErrors": [
        {
-         "error": "Missing order id",
+         "condition": "${ .exceptions | .[] | .error == 'Missing order id' }",
```
tsurdilo, line 896: I assume your condition has to evaluate to true or false here (which we should also document in the specification.md).

Also this assumes (which should prob also be documented) that runtimes have to provide and inject an 'exceptions' json array into the state if error happens which has the format

```json
{
    "exceptions": [
        {
            "error": "...",
            "code": "..."
        },
        ...
    ]
    
}
```

In your example the jq should be "any(.exceptions[]; .error == "Missing order id")" which will return you true/false if an item in the exceptions array exists that has the error property set to "missing order id".

examples/README.md
```diff
@@ -890,21 +890,18 @@ The data output of the workflow contains the information of the exception caught
...
          "transition": "MissingId"
        },
        {
-         "error": "Missing order item",
+         "condition": "${ .exceptions | .[] | .error == 'Missing order item' }",
```
tsurdilo, line 900: same here, jq:
any(.exceptions[]; .error == "Missing order item")

examples/README.md
```diff
@@ -890,21 +890,18 @@ The data output of the workflow contains the information of the exception caught
...
          "transition": "MissingItem"
        },
        {
-        "error": "Missing order quantity",
+         "condition": "${ .exceptions | .[] | .error == 'Missing order quantity' }",
```
tsurdilo, line 904: any(.exceptions[]; .error == "Missing order quantity")

examples/README.md
```diff
@@ -962,11 +959,11 @@ states:
     dataOutputPath: "${ .exceptions }"
   transition: ApplyOrder
   onErrors:
-  - error: Missing order id
+  - condition: ${ .exceptions | .[] | .error == 'Missing order id' }
```
tsurdilo, line 962: any(.exceptions[]; .error == "Missing order id")

examples/README.md
```diff
@@ -962,11 +959,11 @@ states:
...
     transition: MissingId
-  - error: Missing order item
+  - condition: ${ .exceptions | .[] | .error == 'Missing order item' }
```
tsurdilo, line 964: any(.exceptions[]; .error == "Missing order item")

examples/README.md
```diff
@@ -962,11 +959,11 @@ states:
...
     transition: MissingItem
-  - error: Missing order quantity
+  - condition: ${ .exceptions | .[] | .error == 'Missing order quantity' }
```
tsurdilo, line 966: any(.exceptions[]; .error == "Missing order quantity")

examples/README.md
```diff
@@ -1064,10 +1061,10 @@ In the case job submission raises a runtime error, we transition to a SubFlow st
       }
       ],
       "onErrors": [
-      {
-        "error": "*",
-        "transition": "SubmitError"
-      }
+        {
+          "condition": "${ .error }",
```
tsurdilo, line 1065: I assume here, given the previous condition expressions that you want to check if the "exceptions" array is empty or not (again this should return true / false)

In that case this expression should be:

".exceptions | length > 0"

examples/README.md
```diff
@@ -1192,7 +1189,7 @@ states:
     actionDataFilter:
       dataResultsPath: "${ .jobuid }"
   onErrors:
-  - error: "*"
+  - condition: ${ .error }
```
tsurdilo, line 1192: ".exceptions | length > 0"

examples/README.md
```diff
@@ -2695,8 +2692,7 @@ If the retries are not successful, we want to just gracefully end workflow execu
       ],
       "onErrors": [
         {
-          "error": "ServiceNotAvailable",
-          "code": "503",
+          "condition": "${ .error.message == 'ServiceNotAvailable' and .error.code == 503 }",
```
tsurdilo, line 2695: should return true/false imo here too..so in this case use

any(.exceptions[]; .error == "ServiceNotAvailable" and .code == "503")

examples/README.md
```diff
@@ -2754,8 +2750,7 @@ states:
     - functionRef: AssignDoctor
     - functionRef: ScheduleAppt
   onErrors:
-  - error: ServiceNotAvailable
-    code: '503'
+  - condition: ${ .error.message == 'ServiceNotAvailable' and .error.code == 503 }
```
tsurdilo, line 2753: any(.exceptions[]; .error == "ServiceNotAvailable" and .code == "503")

specification.md
```diff
@@ -1622,8 +1622,7 @@ to the trigger/produced event.
 
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
-| error | Domain-specific error name, or '*' to indicate all possible errors | string | yes |
-| code | Error code. Can be used in addition to the name to help runtimes resolve to technical errors/exceptions. Should not be defined if error is set to '*' | string | no |
+| condition | jq expression to resolve against input to error-handling logic | string | yes |
```
tsurdilo, line 1625: what input? imo if we do this we need to describe that the mentioned "exceptions" json array needs to be injected by runtimes into the state data

specification.md
```diff
@@ -1649,7 +1648,7 @@ to the trigger/produced event.
 <td valign="top">
 
 ```yaml
-error: Item not in inventory
+condition: ${ item | .quantity == 0 }
```
tsurdilo, line 1651: .item | .quantity == 0

specification.md
```diff
@@ -1640,7 +1639,7 @@ to the trigger/produced event.
 
 ```json
 {
-   "error": "Item not in inventory",
+   "condition": "${ .item | .quantity == 0 }",
```
tsurdilo, line 1642: .item | .quantity == 0

specification.md
```diff
@@ -1661,12 +1660,7 @@ transition: IssueRefundToCustomer
 
 Error definitions describe errors that can occur during workflow execution and how to handle them. 
 
-The `error`property defines the domain-specific name of the error. Users can also set the name to 
-`*` which is a wildcard specifying "all" errors, in the case where no other error definitions are defined,
-or "all other" errors if there are other errors defined within the same states `onErrors` definition.
-
-The `code` property can be used in addition to `name` to help runtimes resolve the defined 
-domain-specific error to the actual technical errors/exceptions that may happen during runtime execution.
+The `condition` property defines a query against error data returned from, say, a function call. Runtimes should detect data using technology-appropriate means (error codes, exception handling, etc.) and invoke error handling logic as appropriate.
```
tsurdilo, line 1663: the way you structure your jq queries, again, its assumed that there is an "exceptions" array provided. I think this should again be mentioned somewhere

specification.md
```diff
@@ -4225,7 +4130,7 @@ Different states now can use this retry strategy defined. For example:
 {
 "onErrors": [
   {
-    "error": "Inventory service timeout",
+    "condition": "${ .item.error == 'Inventory service timeout' }",
```
tsurdilo, line 4133: Why does the item property all of sudden have error now? It would make more sense to always query against the previously mentioned "exceptions" array. no?

tsurdilo: if you want to keep it like that (which i think we probably should not) the jq query would be:

.item | .error == "Inventory service timeout"

(and the quotes need to be escaped)

specification.md
```diff
@@ -4238,7 +4143,7 @@ Different states now can use this retry strategy defined. For example:
 
 ```yaml
 onErrors:
-- error: Inventory service timeout
+- condition: ${ .item.error == 'Inventory service timeout' }
```
tsurdilo, line 4146: same here 

.item | .error == "Inventory service timeout"

and you will need to escape the quotes

schema/workflow.json
```diff
@@ -193,14 +193,9 @@
     "error": {
       "type": "object",
       "properties": {
-        "error": {
```
tsurdilo, line 196: I would not remove this, but rename it to "name" so users can still have a domain-specific name that they can use for this error. 
so replace with "name" property type string with description "Domain-specific error name"

tsurdilo (changes requested):

tsurdilo:
@jplane thanks for the pr! left some comments

just a little background - we did not want originally to use expressions for error handling. using expressions runs into the issue that we enforce some json data structure of provided errors. 
the way we wanted to do is just allow users to provide a domain specific error name (and in addition some code) 
which then runtimes can map themselves to associated errors that happen during runtime.

I think if we go this route we need to clearly describe the structure of the error data that users are querying against, and that imo puts runtimes in a position that they have to map their internal runtime errors into this json structure. 
wdyt?

tsurdilo:
@jplane just one more thing :) Please also update the roadmap/README.md to add your change(s) as well (under the Next planned release table)

schema/workflow.json
```diff
@@ -220,13 +215,13 @@
       "oneOf": [
         {
           "required": [
-            "error",
+            "condition",
```
tsurdilo, line 218: i would not make condition required here so that runtimes that still want to make a logical mapping against the domain-specific error name can do that

jplane:
Thanks for the feedback. Since you've previously decided against this approach, it seems best that I abandon this PR and adhere to the direction you've already taken with error handling. We can revisit later if needed.

tsurdilo:
@jplane if its OK let's keep this PR open and talk about it in our next team meet so everyone can give their opinion. 
i think the expression approach has its value as well. Was just trying to see if its possible to keep the domain-specific name of the error in addition to your proposed expression. I think this PR can be added I would maybe just look into
1) define the json structure of exceptions/errors that users can query against
2) define if the condition parameter should return true/false

PR #270, opened 2021-02-24
Fix switch state event conditions required params + make default para…

tsurdilo:
…m not required

- [ x] Specification
- [ x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

1) Fixes the switch state event conditions schema (bad cut/paste)
2) Makes the "default" property of conditions not required

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -1950,7 +1950,7 @@ Once all actions have been performed, a transition to another state can occur.
 | [stateDataFilter](#state-data-filter) | State data filter | object | no |
 | [onErrors](#Error-Definition) | States error handling and retries definitions | array | no |
 | eventTimeout | If eventConditions is used, defines the time period to wait for events (ISO 8601 format). For example: "PT15M" (15 minutes), or "P2DT3H4M" (2 days, 3 hours and 4 minutes)| string | yes only if eventConditions is defined |
-| default | Default transition of the workflow if there is no matching data conditions or event timeout is reached. Can be a transition or end definition | object | yes |
+| default | Default transition of the workflow if there is no matching data conditions or event timeout is reached. Can be a transition or end definition | object | no |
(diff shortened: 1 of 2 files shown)


ricardozanini (approved):

specification.md
```diff
@@ -1950,7 +1950,7 @@ Once all actions have been performed, a transition to another state can occur.
 | [stateDataFilter](#state-data-filter) | State data filter | object | no |
 | [onErrors](#Error-Definition) | States error handling and retries definitions | array | no |
 | eventTimeout | If eventConditions is used, defines the time period to wait for events (ISO 8601 format). For example: "PT15M" (15 minutes), or "P2DT3H4M" (2 days, 3 hours and 4 minutes)| string | yes only if eventConditions is defined |
-| default | Default transition of the workflow if there is no matching data conditions or event timeout is reached. Can be a transition or end definition | object | yes |
+| default | Default transition of the workflow if there is no matching data conditions or event timeout is reached. Can be a transition or end definition | object | no |
```
jorgenj, line 1953: Imho, if default is not provided, then we should treat it like the user specified an 'end' definition, and that the doc should be updated here to clearly call that out.  On the call, the alternative idea was to say that a runtime exception should be thrown, imho these are functionally the same but 'end' seems less surprising.

tsurdilo: imho exception is better as modellers can catch it during testing. typically not being able to transition as wanted from a switch state is a really bad thing and masking it with an "end" (workflow termination) might be very confusing imo. wdyt?

tsurdilo:
i think there were no objections to move this forward. will go ahead and merge. please feel free to sent any update prs if there are any needed. ty

PR #271, opened 2021-02-24
Update parallel state branches and foreach states 'workflowid' parameter

tsurdilo:

- [ x] Specification
- [ x] Schema
- [ x] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

While working on a new example I noticed that in parallel state branches and foreach state
the "workflowid" parameter is very weak. It currently references the id of a reusable workflow. 
The problem with this however is that it cannot define things like compensation and the subflow states "repeat" ability.

1) changed the "workflowid" parameter to "subflow"
2) the "subflow" parameter now points to a SubFlow state (via its name), so that all of the SubFlow additional properties can also be used 
in branches and foreach state now
3) added clause that the referenced SubFlow state cannot be part of the main control flow logic (cannot have incoming/outgoing transitions nor can it end workflow execution)
4) updated/fixed required parameters in workflow schema for branches subflow state and foreach state

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -2307,8 +2307,8 @@ Exceptions may occur during execution of branches of the Parallel state, this is
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
 | name | Branch name | string | yes |
-| [actions](#Action-Definition) | Actions to be executed in this branch | array | yes if workflowId is not defined |
-| workflowId | Unique Id of a workflow to be executed in this branch | string | yes if actions is not defined |
+| [actions](#Action-Definition) | Actions to be executed in this branch | array | yes if subflow is not defined |
+| subflow | Name of the SubFlow state to be executed in this branch | string | yes if `actions` is not defined |
 
 <details><summary><strong>Click to view example definition</strong></summary>
 <p>
@@ -2369,8 +2369,11 @@ actions:
 
 Each branch receives the same copy of the Parallel state's data input.
 
-A branch can define either actions or a workflow id of the workflow that needs to be executed.
-The workflow id defined cannot be the same id of the workflow there the branch is defined.
+A Parallel state branch can either define `actions` to be executed, or define a `subflow` property which
+is the name of a workflow SubFlow state to be executed. 
+Note that the referenced SubFlow state
+ should not be part of the core workflow control flow logic (should not have any incoming or outgoing transitions,
+ and should not be able to end workflow execution).
 
 #### <a name="parallel-state-exceptions"></a>Parallel State: Handling Exceptions
 
@@ -2383,9 +2386,9 @@ If the parallel states branch defines actions, all exceptions that arise from ex
  are propagated to the parallel state 
 and can be handled with the parallel states `onErrors` definition.
 
-If the parallel states defines a `workflowId`, exceptions that occur during execution of the called workflow
-can chose to handle exceptions on their own. All unhandled exceptions from the called workflow
-execution however are propagated back to the parallel state and can be handled with the parallel states
+If the parallel state branch defines a `subflow`, exceptions that occur during execution of the subflow
+can be handlerd by the defined SubFlow state. All unhandled exceptions however should be 
+propagated back to the parallel state and can be handled with the parallel states
 `onErrors` definition.
 
 Note that once an error that is propagated to the parallel state from a branch and handled by the 
@@ -2402,7 +2405,7 @@ For more information, see the [Workflow Error Handling](#Workflow-Error-Handling
 | name |State name | string | yes |
 | type |State type | string | yes |
 | waitForCompletion | If workflow execution must wait for sub-workflow to finish before continuing | boolean | yes |
-| workflowId |Sub-workflow unique id | boolean | no |
+| workflowId | Sub-workflow unique id | boolean | no |
 | [repeat](#Repeat-Definition) | SubFlow state repeat exec definition | object | no |
 | [stateDataFilter](#state-data-filter) | State data filter | object | no |
 | [onErrors](#Error-Definition) | States error handling and retries definitions | array | no |
@@ -2456,7 +2459,8 @@ multiple other workflow definitions.
 <img src="media/spec/subflowstateref.png" height="350px" alt="Referencing reusable workflow via SubFlow states"/>
 </p>
 
-Reusable workflow are referenced by their `id` property via the SubFlow states`workflowId` parameter.
+Reusable workflows should be referenced by SubFlow states`workflowId` parameter. The value of the parameter
+should match the `id` parameter of the reusable workflow.
 
 Each referenced workflow receives the SubFlow states data as workflow data input.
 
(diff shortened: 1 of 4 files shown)


specification.md
```diff
@@ -2383,9 +2386,9 @@ If the parallel states branch defines actions, all exceptions that arise from ex
  are propagated to the parallel state 
 and can be handled with the parallel states `onErrors` definition.
 
-If the parallel states defines a `workflowId`, exceptions that occur during execution of the called workflow
-can chose to handle exceptions on their own. All unhandled exceptions from the called workflow
-execution however are propagated back to the parallel state and can be handled with the parallel states
+If the parallel state branch defines a `subflow`, exceptions that occur during execution of the subflow
+can be handlerd by the defined SubFlow state. All unhandled exceptions however should be 
+propagated back to the parallel state and can be handled with the parallel states
```
ricardozanini, line 2391:
```suggestion
If the parallel state branch defines a `subflow`, exceptions that occur during the execution of the subflow
can be handled by the defined SubFlow state. All unhandled exceptions however should be 
propagated back to the parallel state and can be handled with the parallel states
```

specification.md
```diff
@@ -2807,9 +2811,10 @@ It should contain the unique element of the `inputCollection` array and passed a
 
 The `actions` property defines actions to be executed in each state iteration.
 
-If actions are not defined, you can specify the `workflowid` to reference a workflow id which needs to be executed
-for each iteration. Note that `workflowid` should not be the same as the workflow id of the workflow where the foreach state
-is defined.
+If actions are not defined, you can specify the `subflow` parameter. It's value should be a name of 
+ a SubFlow state which should be invoked for each iteration. Note that the referenced SubFlow state
+ should not be part of the core workflow control flow logic (should not have any incoming or outgoing transitions,
+ and should not be able to end workflow execution).
```
ricardozanini, line 2817:
```suggestion
If actions are not defined, you can specify the `subflow` parameter. Its value should be a name of a SubFlow state which should be invoked for each iteration.
Note that the referenced SubFlow state should not be part of the core workflow control flow logic
(should not have any incoming or outgoing transitions, and should not be able to end workflow execution).
```

ricardozanini (approved):

jorgenj:
The description of this PR seems to assert that a sub-flow in a parallel state should be directly/individually compensatable and retryable, and thus introduces a reference to a separate sub-flow state instead of a direct reference to the sub workflow where compensation and retry logic can be defined.  

It seems that the focus is purely on how to get this behavior specifically for the case where a parallel branch is a sub-workflow.   I question why these concerns for compensation and retry are only being addressed for sub-flow branches, and not other types of branches.

Specifically with regard to compensation, I would like to suggest that we set the expectation that compensation applies to the whole state, and in the case of parallel the workflow implementor should have compensation logic that handles the fact that a parallel state might have done many (or even none) of the branches (it's a responsibility of the workflow user to explicitly define exactly how to compensate the branches of a parallel state).   Otherwise it gets incredibly complex to handle compensation for a parallel state.   

Consider the case where we support compensation for a parallel state w/ 3 sub-workflow branches.  When the parallel state fails, in what order should the compensation for those branches be executed?  Must the compensation for all branches be executed in parallel, or is it safe for a runtime to just assume that any order is perfectly safe?

tsurdilo:
@jorgenj 
**a sub-flow in a parallel state should be directly/individually compensatable**: why is that a bad thing?

**Consider the case where we support compensation for a parallel state w/ 3 sub-workflow branches. When the parallel state fails, in what order should the compensation for those branches be executed?**: compensation is always done in the **reverse** order of already executed work. So in case of a parallel state X with branches A, B, C, if all branches execute and then the state executes and compensation is triggered, the compensation order would be A, B, C (in the exact order in reverse they were executed), X. 
This of course depends on the parallel state completionType ( as some branches that we didnt wait on could not have yet completed by the time compensation is triggered so they would not be included)

Hope this helps

jorgenj:
> **a sub-flow in a parallel state should be directly/individually compensatable**: why is that a bad thing? Even if you had workflow written in code that in parallel starts multiple "actions" you would want to be able to compensate each one if needed.

Are you saying that 'action' branches should also be separately compensatable and repeatable?  Because this change completely ignores the fact that you also can't specify compensation or repeat for each action branch.

Imho, if a user wants to compensate a parallel state they should compensate the whole state.  If a customer wants their parallel state to be compensated in parallel, then the compensation state can be a parallel state:

```
states:
  - name: doParallel
    type: parallel
    branches:
    - workflowId: subWf1
    - actions:
      - functionRef: func1
    - workflowId: subWf2
    compensatedBy: doParallelCompensations

 - name: doParallelCompensations
    type: parallel
    branches:
    - workflowId: cleanupSubWf1
    - actions:
      - functionRef: cleanupFunc1
    - workflowId: cleanupSubWf2
```

If a user determines that compensation has to occur in a particular order serially, they have that control to even do fancy stuff like some serial and some parallel:
```
states:
  - name: doParallel
    type: parallel
    branches:
    - workflowId: subWf1
    - actions:
      - functionRef: func1
    - workflowId: subWf2
    compensatedBy: doCompensation1

 - name: doCompensation1
    type: subflow
    workflowId: cleanupSubWf2
    transition: doParallelCompensations
 - name: doParallelCompensations
    type: parallel
    branches:
    - actions:
      - functionRef: cleanupFunc1
    - workflowId: cleanupSubWf1
```

jorgenj:
@tsurdilo =)

tsurdilo:
@jorgenj yeah the actions is an for compensation atm. will need to address that. I would probably suggest that each branch can have compensation logic

jorgenj:
This is adding too much complexity for very little benefit.  The path we're taking here (adding compensation & repeat for branches in parallel) implies that we'll also add compensation at every level of every other state.

Are we then going to add...
* Compensation for individual actions in Operation?
* Compensation for individual loop inputs in ForEach?
* Compensation for individual actions of individual loop inputs in ForEach?
* Compensation for individual conditions of Switch / Event?

This is just one guys opinion, but just because we can think of a use-case for how someone **might** use compensation / repeat, doesn't mean it's a good idea to support that.  I think the bar for adding new features, at least to some degree, should be whether they can accomplish what they need with existing features.

tsurdilo:
@jorgenj no, only parallel state applies here. parallel state can be seen as a

split -> parallel individual business logic exec -> join
each of those parallel executions is not performed as a "unit of work", each branch can have a totally different
set of business instructions to what it should perform and has the right to define its own compensation imo

foreach state is a single unit of work , meaning its functionalities are not independent in nature, same for operation state.

jorgenj:
> each of those parallel executions is not performed as a "unit of work", each branch can have a totally different
> set of business instructions to what it should perform and has the right to define its own compensation imo

What about this can not be accomplished with compensation at the state level only?   Please share an example that requires compensation for a specific branch.

tsurdilo:
@jorgenj 
Regarding **Please share an example that requires compensation for a specific branch.**: 
simplest example could be 2 branches, 1 with a short-running activity and the other is a long-running one. parallel state has completionType n_of_m where n is 1, meaning as soon as the short activity completes the parallel state can transition while the long one is still active. Before the long one completes the workflow starts compensation.

If you model the compensation of those 2 branches on the state level, you would have to check which activity actually completed. this can be very hard to do and imagine for a parallel state with a ton more branches. 
by specifying compensation on the branch level this is much easier as the runtime knows which of the branches completed execution and can be compensated.

PR #272, opened 2021-02-24
Update subflow state repeat

tsurdilo:

- [ x] Specification
- [x ] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

While working on a new example noticed that the subflow "repeat" is missing some features:
1) Added "delay" property which denotes the time delay between repeat executions
2) Added "timeout" property which denotes the max time for repeat executions. This works well with the "max" property as well
3) Fixes the repeat definition required fields in schema

### DIFF
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -1243,8 +1243,8 @@
           "description": "Sub-workflow unique id"
         },
         "repeat": {
-          "$ref": "#/definitions/repeat",
-          "description": "SubFlow state repeat exec definition"
+          "description": "SubFlow state repeat exec definition",
+          "$ref": "#/definitions/repeat"
         },
         "stateDataFilter": {
           "$ref": "#/definitions/statedatafilter"
@@ -1966,8 +1966,28 @@
           "items": {
             "type": "string"
           }
+        },
+        "delay": {
+          "type": "string",
+          "description": "Time delay between repeats (ISO 8601 duration format)"
+        },
+        "timeout": {
+          "type": "string",
+          "description": "Sets the maximum amount of time for repeat executions (ISO 8601 format)"
         }
-      }
+      },
+      "oneOf": [
+        {
+          "required": [
+            "expression"
+          ]
+        },
+        {
+          "required": [
+            "stopOnEvents"
+          ]
+        }
+      ]
     }
   }
 }
--- a/specification.md
+++ b/specification.md
@@ -3047,6 +3047,8 @@ If the defined callback event has not been received during this time period, the
 | max | Sets the maximum amount of repeat executions | integer | no |
 | continueOnError | If set to `true` repeats executions in a case unhandled errors propagate from the sub-workflow to this state | boolean | no |
 | stopOnEvents | List referencing defined consumed workflow events. SubFlow will repeat execution until one of the defined events is consumed, or until the max property count is reached | array | no |
+| delay | Time delay between repeat executions (ISO 8601 duration format) | string | no |
+| timeout | Sets the maximum amount of time for repeat executions (ISO 8601 format) | string | no |
 
 <details><summary><strong>Click to view example definition</strong></summary>
 <p>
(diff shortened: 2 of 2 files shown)


ricardozanini (approved):

schema/workflow.json
```diff
@@ -1243,8 +1243,8 @@
           "description": "Sub-workflow unique id"
         },
         "repeat": {
-          "$ref": "#/definitions/repeat",
-          "description": "SubFlow state repeat exec definition"
+          "description": "SubFlow state repeat exec definition",
+          "$ref": "#/definitions/repeat"
```
jorgenj, line 1247: To be honest, I'm a little confused about the repeat behavior.  It seems to me that if a parent & sub-workflow have a use-case for waiting until some condition is true, then maybe the sub-flow should just do repetition and the parent workflow should just wait for the sub-workflow to complete once.

tsurdilo: Not sure I understand this question - can you show an example please?

jorgenj: Imho, we already have primitives in workflow for doing looping and/or repeating an operation in a workflow.  If users want to repeat something, imho, they should be doing that inside the sub-workflow and not by adding extra complexity to the language for the ways in which a subFlow can be invoked.   

It seems like we're repeating all the functionality that we already have around retrying actions but for sub-workflows.  Perhaps instead a subflow should just be another type of action invocation?

jorgenj: To give an example of what I mean, I think we could do the following and the language would be much more consistent:

* Add 'subFlow' as an optional parameter for an Action definition.  This would be an object with 2 parameters, one is the sub-workflow ID and the other a waitForCompletion flag.
* Drop the SubFlow state from the spec
* Drop the 'workflowId' parameter from parallel-state branch definition

With that change, a customer can use sub-flows like they can invoke any other type of action in parallel, operation, & other states.   Existing retry behavior could now be reused for sub-workflows, looping constructs that are available for actions are also automatically supported.

Really, I see very little difference in behavior between a sub-workflow & the eventRef/functionRef style of action (under the covers all invoke something and wait for a result back).  In the case that a user sets waitForCompletion=false, then the spec should indicate that the parent workflow should behave as if the sub-workflow completed immediately with an output value of an empty json object perhaps.

If you compare the current description of Operation and SubFlow states, there seems to be very little difference in their described behavior.

jorgenj: @tsurdilo =)

tsurdilo: @jorgenj i agree that as we add more and more features and improve spec we will need to start looking at feature duplication..thats unavoidable with growth :( 

but i dont understand how operation and subflow states are similar i nature. operation state performs rest/grpc invocations of functions only, a subflow state has workflow control flow logic . please explain

jorgenj: Saying that one only does rest/grpc and the other only invokes workflows is a completely arbitrary distinction. 

Both invoke an external process and receive a result back (either by rest/grpc, or events, or a rest call to some launch workflow API).  Both allow for handling errors and retrying.  Both have all the other common state functionality like transition/end, compensation, etc.

SubFlow lacks the ability to do multiple, serially or parallel. And subFlow has this odd `repeat` that arguably should go away because it duplicates behavior that can readily be implemented in the invoked workflow instead.

tsurdilo: subflow allows you to define reusable components that you can plug/play into your workflow solutions. they have their use.
as far as things like error handling, transitions compensation, i mean those are  a set of common functionalities that all states have

i mean we could just delete all states and just have a single inject state and be done. i am all for it

tsurdilo: regarding repeat - we can change this if its weird or awkward, but i will just leave this here https://github.com/serverlessworkflow/specification/issues/239#issuecomment-768062283 
if we can come up with a better way that still works lets go for it

jorgenj: I posted a question there, hoping we can get some input from the original requestor.  From that issue, I don't have a clear understanding what they were trying to do, so it's not clear if this could've been solved a different way (or even if it could've been solved with existing functionality).

Issue #300, opened 2021-03-17
Move sub-flow invocation to be an action


jorgenj:
**What would you like to be added**:

Remove subflow state, instead invocation of sub-flows should be an 'action'.

**Why is this needed**:

To give an example of what I mean, I think we could do the following and the language would be much more consistent:

* Add 'subFlow' as an optional parameter for an Action definition. This would be an object with 2 parameters, one is the sub-workflow ID and the other a waitForCompletion flag.
* Drop the SubFlow state from the spec
* Drop the 'workflowId' parameter from parallel-state branch definition

With that change, a customer can use sub-flows like they can invoke any other type of action in parallel, operation, & other states. Existing retry behavior could now be reused for sub-workflows, looping constructs that are available for actions are also automatically supported.

Really, I see very little difference in behavior between a sub-workflow & the eventRef/functionRef style of action (under the covers all invoke something and wait for a result back). In the case that a user sets waitForCompletion=false, then the spec should indicate that the parent workflow should behave as if the sub-workflow completed immediately with an output value of an empty json object perhaps.

A good illustration of my point is that if you compare the current description of Operation and SubFlow states, there seems to be very little difference in their features/behavior.

jorgenj:
This has been discussed in the weekly meeting already, but I realized that an issue was never created yet.  Please go ahead and assign this to me.

PR #301, opened 2021-03-17
Move support for sub-flows to actions

jorgenj:
Move support for sub-flows to actions, so we get all the other features of actions for sub-flows (like looping, retries, error-handling, etc)

- [✅ ] Specification
- [✅ ] Schema
- [✅ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

The intent is to unify the experience for using actions & sub-flows, since sub-flow use-cases have many similar needs as actions.

https://github.com/serverlessworkflow/specification/issues/300

> To give an example of what I mean, I think we could do the following and the language would be much more consistent:
>
> Add 'subFlow' as an optional parameter for an Action definition. This would be an object with 2 parameters, one is the sub-workflow ID and the other a waitForCompletion flag.
Drop the SubFlow state from the spec
Drop the 'workflowId' parameter from parallel-state branch definition
With that change, a customer can use sub-flows like they can invoke any other type of action in parallel, operation, & other states. Existing retry behavior could now be reused for sub-workflows, looping constructs that are available for actions are also automatically supported.
>
> Really, I see very little difference in behavior between a sub-workflow & the eventRef/functionRef style of action (under the covers all invoke something and wait for a result back). In the case that a user sets waitForCompletion=false, then the spec should indicate that the parent workflow should behave as if the sub-workflow completed immediately with an output value of an empty json object perhaps.
>
> A good illustration of my point is that if you compare the current description of Operation and SubFlow states, there seems to be very little difference in their features/behavior.

### DIFF
--- a/comparisons/comparison-bpmn.md
+++ b/comparisons/comparison-bpmn.md
@@ -317,7 +317,8 @@ states:
   inputCollection: "${ .inputsArray }"
   iterationParam: "${ .inputItem }"
   outputCollection: "${ .outputsArray }"
-  workflowId: doSomethingAndWaitForMessage
+  actions:
+  - subFlowRef: doSomethingAndWaitForMessage
   end: true
 ```
 
@@ -349,11 +350,12 @@ name: SubFlow Loop Workflow
 version: '1.0'
 start: SubflowLoop
 states:
-- name: SubflowLoop
-  type: subflow
-  workflowId: checkAndReplyToEmail
-  repeat:
-    max: 100
+- name: SubflowRepeat
+  type: foreach
+  inputCollection: ${ [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, ..., 100 ] }
+  max: 1
+  actions:
+  - subFlow: checkAndReplyToEmail
   end: true
 ```
 
@@ -448,8 +450,9 @@ version: '1.0'
 start: A
 states:
 - name: A
-  type: subflow
-  workflowId: asubflowid
+  type: operation
+  actions:
+    - subFlow: asubflowid
   transition: Event Decision
 - name: Event Decision
   type: switch
--- a/comparisons/comparison-bpmn.md
+++ b/comparisons/comparison-bpmn.md
@@ -348,15 +348,28 @@ a starting "operation" state transitioning to an "event" state which waits for t
 id: subflowloop
 name: SubFlow Loop Workflow
 version: '1.0'
-start: SubflowLoop
+start: StartCount
 states:
+- name: StartCount
+  type: inject
+  data:
+    counter: 0
+  transition: SubflowRepeat
 - name: SubflowRepeat
-  type: foreach
-  inputCollection: ${ [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, ..., 100 ] }
-  max: 1
+  type: operation
   actions:
   - subFlow: checkAndReplyToEmail
-  end: true
+  actionDataFilter:
+    fromStateData: ${ .someInput }
+    toStateData: ${ .someInput }
+  transition: CheckCount
+- name: CheckCount
+  type: Switch
+  dataConditions:
+  - condition: ${ .counter < 100 }
+    transition: SubflowRepeat
+  default:
+    end: true
 ```
 
 </td>
--- a/comparisons/comparison-bpmn.md
+++ b/comparisons/comparison-bpmn.md
@@ -348,27 +348,22 @@ a starting "operation" state transitioning to an "event" state which waits for t
 id: subflowloop
 name: SubFlow Loop Workflow
 version: '1.0'
-start: StartCount
+start: SubflowRepeat
 states:
-- name: StartCount
-  type: inject
-  data:
-    counter: 0
-  transition: SubflowRepeat
 - name: SubflowRepeat
   type: operation
   actions:
-  - subFlow: checkAndReplyToEmail
+  - subFlowRef: checkAndReplyToEmail
   actionDataFilter:
     fromStateData: ${ .someInput }
     toStateData: ${ .someInput }
+  stateDataFilter:
+    output: ${ .maxChecks -= 1 }
   transition: CheckCount
 - name: CheckCount
   type: Switch
-  stateDataFilter:
-    input: ${ .counter += 1 }
   dataConditions:
-  - condition: ${ .counter < 100 }
+  - condition: ${ .maxChecks > 0 }
     transition: SubflowRepeat
   default:
     end: true
@@ -377,6 +372,8 @@ states:
 </td>
 </tr>
 </table>
+
+This workflow assumes that the input to the workflow includes a maxChecks attribute set to an integer value.
 
 * Note: We did not include the `checkAndReplyToEmail` workflow in this example, which would include the 
 control-flow logic to check email and make a decision to reply to it or wait an hour.
@@ -467,7 +464,7 @@ states:
 - name: A
   type: operation
   actions:
-    - subFlow: asubflowid
+    - subFlowRef: asubflowid
   transition: Event Decision
 - name: Event Decision
   type: switch
--- a/comparisons/comparison-bpmn.md
+++ b/comparisons/comparison-bpmn.md
@@ -365,6 +365,8 @@ states:
   transition: CheckCount
 - name: CheckCount
   type: Switch
+  stateDataFilter:
+    input: ${ .counter += 1 }
   dataConditions:
   - condition: ${ .counter < 100 }
     transition: SubflowRepeat
--- a/examples/README.md
+++ b/examples/README.md
@@ -10,7 +10,7 @@ Provides Serverless Workflow language examples
 - [Solving Math Problems (ForEach state)](#Solving-Math-Problems-Example)
 - [Parallel Execution](#Parallel-Execution-Example)
 - [Event Based Transitions (Event-based Switch)](#Event-Based-Transitions-Example)
-- [Applicant Request Decision (Data-based Switch + SubFlow states)](#Applicant-Request-Decision-Example)
+- [Applicant Request Decision (Data-based Switch + SubFlows)](#Applicant-Request-Decision-Example)
 - [Provision Orders (Error Handling)](#Provision-Orders-Example)
 - [Monitor Job for completion (Polling)](#Monitor-Job-Example)
 - [Send CloudEvent on Workflow Completion](#Send-CloudEvent-On-Workfow-Completion-Example)
@@ -24,7 +24,7 @@ Provides Serverless Workflow language examples
 - [New Patient Onboarding (Error checking and Retries)](#New-Patient-Onboarding)
 - [Purchase order deadline (ExecTimeout)](#Purchase-order-deadline)
 - [Accumulate room readings and create timely reports (ExecTimeout and KeepActive)](#Accumulate-room-readings)
-- [Car vitals checks (SubFlow state Repeat)](#Car-Vitals-Checks)
+- [Car vitals checks (SubFlow Repeat)](#Car-Vitals-Checks)
 - [Book Lending Workflow](#Book-Lending)
 - [Filling a glass of water (Expression functions)](#Filling-a-glass-of-water)
 - [Online Food Ordering](#Online-Food-Ordering)
@@ -626,20 +626,32 @@ period, the workflow transitions to the "HandleNoVisaDecision" state.
   },
   {
     "name": "HandleApprovedVisa",
-    "type": "subflow",
-    "workflowId": "handleApprovedVisaWorkflowID",
+    "type": "operation",
+    "actions": [
+      {
+        "subFlow": "handleApprovedVisaWorkflowID" 
+      }
+    ],
     "end": true
   },
   {
       "name": "HandleRejectedVisa",
-      "type": "subflow",
-      "workflowId": "handleRejectedVisaWorkflowID",
+      "type": "operation",
+      "actions": [
+        {
+          "subFlow": "handleRejectedVisaWorkflowID" 
+        }
+      ],
       "end": true
   },
   {
       "name": "HandleNoVisaDecision",
-      "type": "subflow",
-      "workflowId": "handleNoVisaDecisionWorkfowId",
+      "type": "operation",
+      "actions": [
+        {
+          "subFlow": "handleNoVisaDecisionWorkfowId" 
+        }
+      ],
       "end": true
   }
 ]
@@ -674,16 +686,19 @@ states:
   default:
     transition: HandleNoVisaDecision
 - name: HandleApprovedVisa
-  type: subflow
-  workflowId: handleApprovedVisaWorkflowID
+  type: operation
+  actions:
+    - subFlow: handleApprovedVisaWorkflowID
   end: true
 - name: HandleRejectedVisa
-  type: subflow
-  workflowId: handleRejectedVisaWorkflowID
+  type: operation
+    actions:
+      - subFlow: handleRejectedVisaWorkflowID
   end: true
 - name: HandleNoVisaDecision
-  type: subflow
-  workflowId: handleNoVisaDecisionWorkfowId
+  type: operation
+    actions:
+      - subFlow: handleNoVisaDecisionWorkfowId
   end: true
 ```
 
@@ -695,7 +710,7 @@ states:
 
 #### Description
 
-This example shows off the switch state and the subflow state. The workflow is started with application information data as input:
+This example shows off the switch state and the subflow action. The workflow is started with application information data as input:
 
 ```json
     {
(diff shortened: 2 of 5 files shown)


comparisons/comparison-bpmn.md
```diff
@@ -312,10 +312,11 @@ version: '1.0'
 start: SubflowRepeat
 states:
 - name: SubflowRepeat
-  type: subflow
-  workflowId: dosomethingandwaitformessage
-  repeat:
-    max: 10
+  type: foreach
+  inputCollection: ${ [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 ] }
```
tsurdilo, line 316: ForEach state is incorrect here imo. 
Foreach state executes each iteration in parallel - https://github.com/serverlessworkflow/specification/blob/main/specification.md#foreach-state --> "Each iteration of the ForEach state should be executed in parallel."

where "repeat" is a loop one at a time

also in jq you can use "[range(0;9)]" as the expression instead of creating the array manually 

So I think foreach state is not a replacement for repeat

examples/README.md
```diff
@@ -3291,24 +3341,57 @@ We fist define our top-level workflow for this example:
        },
        {
           "name": "DoCarVitalsChecks",
-          "type": "subflow",
+          "type": "operation",
+          "actions": [
+            {
+              "subFlow": {
+                "workflowId": "vitalscheck",
+                "waitForCompletion": false
```
tsurdilo, line 3349: if we allow async exec for subflow actions it should be also possible for functionRef actions

examples/README.md
```diff
@@ -3291,24 +3341,57 @@ We fist define our top-level workflow for this example:
...
+          "transition": "WaitForCarStopped"
+       },
+       {
+          "name": "WaitForCarStopped",
+          "type": "event",
           "workflowId": "vitalscheck",
-          "repeat": {
-             "stopOnEvents": ["CarTurnedOffEvent"]
-          },
+          "onEvents": [
```
tsurdilo, line 3359: Don't understand the logic here. The intend is to repeat execution of the subflow until the car stopped event is consumed. this seems its just doing it once and transitioning to this event state. am i missing something?

tsurdilo: ah ok there are more states below.

tsurdilo:
I think moving subflow invocation in actions is a good idea as discussed in meeting.

Don't think removing repeat without an adequate alternative tho is a good thing (as we had the community request for it).
We can take a look at allowing repeating (loop) in actions as foreach state is not an alternative for it.

Do any of the example images need updates for this ?

comparisons/comparison-bpmn.md
tsurdilo: also i think this implies that input to the inputCollection is an empty object where as the input to the inputCollection is the state data 
(meaning the expression in the pr or using "range" only work on "{ }", they don't just generate an array out of the blue)

tsurdilo: just noticed the bpmn image is incorrect, in that it shows the multi-instance subprocess instead of the loop marker. That probably caused the confusion, sorry. Will update the image asap

ricardozanini:
@tsurdilo @jorgenj, can we leave the `repeat` topic off this PR? So we can discuss it further?

tsurdilo:
@ricardozanini I am not sure if that's possible as this pr removes the subflow state and thus removes the repeat/loop functionality as well. i think we can definitely use this pr with the addition of repeat to actions. Just have to figure out what the best way is to do that

tsurdilo:
updated bpmn2 examples - https://github.com/serverlessworkflow/specification/blob/main/comparisons/comparison-bpmn.md (sorry this one will need a small rebase)

comparisons/comparison-bpmn.md
jorgenj: You ignored that I put `max: 1`.  This makes it serial in nature...

tsurdilo: ok i see, missed it. i guess then the issue left is  that data input for each subflow exec here is the element of the input collection (1, 2, 3, 4 ...) and not state data as intended by business logic in the initial example.

tsurdilo: i think i removed a comment here too somehow :( 
inputCollection expression data input is the state data, so i think it cannot be an expression that somehow creates the data input for the subflow executions. either way the expressions have to operate on some data input and i assume the intend was that they somehow operate on "{ }" maybe, idk. for example it guess just like ${ .orders } or even ${ .inputArray } should be fine for users to undertand

jorgenj: @tsurdilo I'm not sure I follow what you're saying in your last comment.

I can adjust the expression so that it's iterating 10 times and the input to each iteration is the state input with the following, I believe:

```yaml
inputCollection: ${ reduce range(0,10) as $d ([.]; .[$d+1] += .[0] ) }
```

That's probably not very easy to understand at first glance though.   Are you suggesting that I just change the example to say there's an array of 'orders' and the example just iterates over those 'orders'?

jorgenj: Got more clarity, will replace 'repeat' usage in examples with explicit control-flow logic.

tsurdilo:
@jorgenj there are some conflicts that should be resolved before merging. If you want I can fix them, just let me know It's np (I am a rebase pro ;) )

comparisons/comparison-bpmn.md
```diff
@@ -309,14 +309,30 @@ functions:
 id: subflowloop
 name: SubFlow Repeat workflow
 version: '1.0'
-start: SubflowRepeat
+start: StartCount
 states:
+- name: StartCount
+  type: inject
+  data:
+    counter: 0
```
tsurdilo, line 317: I think here you dont have to use the inject state. Just assume that "count" (I would call it "max") is passed as workflow data input.

comparisons/comparison-bpmn.md
```diff
@@ -309,14 +309,30 @@ functions:
...
+  actions:
+    - subFlow: dosomethingandwaitformessage
+  actionDataFilter:
+    fromStateData: ${ .someInput }
+    toStateData: ${ .someInput }
+  transition: CheckCount
+- name: CheckCount
+  type: Switch
+  stateDataFilter:
+    input: ${ .counter += 1 }
```
tsurdilo, line 330: I think it would be clearer if we move this to the above operation state (in stateDataFilter output param)

examples/README.md
```diff
@@ -3409,12 +3509,52 @@ And then our reusable sub-workflow which performs the checking of our car vitals
...
-         "timeDelay": "PT2M",
-         "end": true
+         "type": "event",
+         "timeout": "PT2M",
+         "onEvents": [
+            {
+              "eventRefs": ["StopVitalsCheck"],
+              "eventDataFilter": {
+                "data": "${ true }",
+                "toStateData": "${ .stopRecieved }"
```
tsurdilo, line 3519: Received?

examples/README.md
```diff
@@ -3409,12 +3509,52 @@ And then our reusable sub-workflow which performs the checking of our car vitals
...
-         "type": "delay",
-         "timeDelay": "PT2M",
-         "end": true
+         "type": "event",
+         "timeout": "PT2M",
+         "onEvents": [
+            {
+              "eventRefs": ["StopVitalsCheck"],
+              "eventDataFilter": {
+                "data": "${ true }",
```
tsurdilo, line 3518: What is the ${ true } expression supposed to do?

examples/README.md
```diff
@@ -3409,12 +3509,52 @@ And then our reusable sub-workflow which performs the checking of our car vitals
...
+         ],
+         "transition": "ShouldStopOrContinue"
+      },
+      {
+         "name": "ShouldStopOrContinue",
+         "type": "switch",
+         "dataConditions": [
+            {
+               "name": "Stop Event Received",
+               "condition": "${ .stopRecieved }",
```
tsurdilo, line 3531: Received

specification.md
```diff
@@ -2268,8 +2267,9 @@ instance in case it is an end state without performing any actions.
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
 | name | Unique action name | string | no |
-| [functionRef](#FunctionRef-Definition) | References a reusable function definition | object | yes if `eventRef` is not used |
-| [eventRef](#EventRef-Definition) | References a `trigger` and `result` reusable event definitions | object | yes if `functionRef` is not used |
+| [functionRef](#FunctionRef-Definition) | References a reusable function definition | object | yes if `eventRef` & `subFlow` are not defined |
+| [eventRef](#EventRef-Definition) | References a `trigger` and `result` reusable event definitions | object | yes if `functionRef` & `subFlow` are not defined |
+| subFlow | Either string containing the unique id of the sub-workflow to be invoked or a [subFlowRef](#SubFlowRef-Definition) object | object or string | yes if `eventRef` & `functionRef` are not defined |
```
tsurdilo, line 2272: Should we not call this "subflowRef" instead so it fits nicely with "eventRef and "functionRef"?

specification.md
```diff
@@ -4223,7 +4128,7 @@ their execution followed by a transition another workflow state, given their con
 
 The `terminate` property, if set to `true`, completes the workflow instance execution, this any other active 
 execution paths.
-If a terminate end is reached inside a ForEach, Parallel, or SubFlow state, the entire workflow instance is terminated.
+If a terminate end is reached inside a ForEach, or Parallel, the entire workflow instance is terminated.
```
tsurdilo, line 4131: inside a ForEach or Parallel state
(missing "state")

tsurdilo:
@jorgenj added a couple of small comments. 
The only thing that I think is left is to made a decision on what to do with waitForCompletion in the subflow definition.
As we currently don't have a similar construct for functionRef it's weird to just have it on subflow.

We could address it with this PR or if you want with a separate one, up to you, let me know what you want to do.

tsurdilo:
@jorgenj please take a look at https://github.com/serverlessworkflow/specification/issues/329
i think this is a very valid request. Let me know if you would like to add it to be part of this pr or a separate.

jorgenj:
> @jorgenj please take a look at #329
> i think this is a very valid request. Let me know if you would like to add it to be part of this pr or a separate.

I'd like to get this merged and we can tackle that in a separate PR

jorgenj:
> @jorgenj added a couple of small comments.
> The only thing that I think is left is to made a decision on what to do with waitForCompletion in the subflow definition.
> As we currently don't have a similar construct for functionRef it's weird to just have it on subflow.
> 
> We could address it with this PR or if you want with a separate one, up to you, let me know what you want to do.

I'd prefer if we tackle that in a separate PR, it's not clear to me where that property should reside.  On both functionRef and subflowRef or in some shared location?

comparisons/comparison-bpmn.md
jorgenj: Done

comparisons/comparison-bpmn.md
jorgenj: Done

examples/README.md
jorgenj: It's supposed to replace the results of the event, so that the value assigned to .stopReceived is 'true'.

specification.md
jorgenj: For functions and events, we have other objects in the same document that the functionRef/eventRef are referring to.  For subflow, it's a completely different doc so it seemed odd to me to call it subFlowRef.  It's a minor thing, I'll defer to your judgement and rename it subFlowRef.

examples/README.md
jorgenj: Switched to using has() in the switch state's data-condition.

tsurdilo:
@jorgenj thank you for making the updates and rebasing! I wanted to ask one more thing please
I don't understand how the subflow "waitforcompletion" interacts with the actionMode property on actions which can be "sequential" or "parallel" currently. 
The issue is that when we set actionMode to "sequential" (which is the default) we guarantee that the results of the action will be merged to state data before the next action is executed.
With that and waitForCompletion set to false, I am not sure this fits that model any longer and imo can cause weird behaviour.

Let me know what you think. Thanks.

jorgenj:
> @jorgenj thank you for making the updates and rebasing! I wanted to ask one more thing please
> I don't understand how the subflow "waitforcompletion" interacts with the actionMode property on actions which can be "sequential" or "parallel" currently.
> The issue is that when we set actionMode to "sequential" (which is the default) we guarantee that the results of the action will be merged to state data before the next action is executed.
> With that and waitForCompletion set to false, I am not sure this fits that model any longer and imo can cause weird behaviour.
> 
> Let me know what you think.
> Imo we should remove the waitForCompletion on subflow action for now and it would be nice if it was done alongside this PR please

@tsurdilo 

I'm fine removing it for now in this PR.  

Though, my interpretation is that when waitForCompletion is false, then the workflow should behave as if the sub-workflow (or action) completely immediately with an empty json object as the output of the subflow/action.   In that regard, then as far as actionMode=sequential is concerned it's really no different from a subflow/action that we waited for and which returned an empty json object as it's output.   WDYT?

tsurdilo:
@jorgenj I think what you mentioned is fair, and we should probably make sure it is in the spec doc as well.
Had a chance to think about this for a bit and there are couple of things that are concerning me, some we already mentioned before like:
1) adding consistent loop/repeat ability for actions (which we said we would address in near future)
2) adding consistency to sync/async capabilities of all action types (functionRef, eventRef, and subflowRef). For this if we want to tackle this in a separate PR thats fine but I do think that we should remove waitForCompletion for now and address it across the board. One more thing that we should probably address here is the functionRef to a function definition of type expression. I think that either we should state in spec that referencing expression type functions should not be allowed in actions or if it is allowed to define that async invocation of those if probably not allowed.

3) (this one is new :) ) - if you look at the actions now, both functionRef and eventRef reference function and event definitions which are defined outside of the main "states" control flow logic definition, and allow users to define those as separate json/yaml which can be imported into the main workflow definition.
I believe that this should also be the case for subFlow actions, meaning that I would like to also have a "subflows" top-level definition which just like the others can be an in-line array or an URI to a separate file where those are defined.
This imo would help with consistency across different actions as well. WDYT?

jorgenj:
> @jorgenj I think what you mentioned is fair, and we should probably make sure it is in the spec doc as well.
> Had a chance to think about this for a bit and there are couple of things that are concerning me, some we already mentioned before like:
> 
> 1. adding consistent loop/repeat ability for actions (which we said we would address in near future)

I'm not sure I understand what you'd like to be done in this PR.  Can you expand on this?

> 2. adding consistency to sync/async capabilities of all action types (functionRef, eventRef, and subflowRef). For this if we want to tackle this in a separate PR thats fine but I do think that we should remove waitForCompletion for now and address it across the board. 

Got it, I'll remove it for now then.

> One more thing that we should probably address here is the functionRef to a function definition of type expression. I think that either we should state in spec that referencing expression type functions should not be allowed in actions or if it is allowed to define that async invocation of those if probably not allowed.

I'm not sure I understand the connection to this PR.  This seems more like a future concern for the PR where async support for functions/subFlows is added?

> 3. (this one is new :) ) - if you look at the actions now, both functionRef and eventRef reference function and event definitions which are defined outside of the main "states" control flow logic definition, and allow users to define those as separate json/yaml which can be imported into the main workflow definition.
>    I believe that this should also be the case for subFlow actions, meaning that I would like to also have a "subflows" top-level definition which just like the others can be an in-line array or an URI to a separate file where those are defined.
>    This imo would help with consistency across different actions as well. WDYT?

That's an interesting idea, but I'm not sure I understand the benefit.  The subFlowRef is already a pretty light-weight 'object' referencing an external resource, I'm not sure how much benefit there would be in sharing those externally.

jorgenj:
@tsurdilo

> > 1. adding consistency to sync/async capabilities of all action types (functionRef, eventRef, and subflowRef). For this if we want to tackle this in a separate PR thats fine but I do think that we should remove waitForCompletion for now and address it across the board.
> 
> Got it, I'll remove it for now then.

On second thought, there's a couple examples that would be invalidated and/or hard to reproduce if I remove `waitForCompletion`.  Is this a strong requirement that we *must* remove waitForCompletion for now?  It seems like busy-work if we're just going to be adding it back again later.

specification.md
```diff
@@ -2464,6 +2484,50 @@ to be used as payload of the event referenced by `triggerEventRef`. If it is of
...
+</td>
+</tr>
+</table>
+
+</details>
+
+The `waitForCompletion` property defines if the SubFlow action should wait until the referenced reusable workflow
+has completed its execution. If it's set to "true" (default value), SubFlow action execution must wait until the referenced workflow has completed its execution.
+In this case the workflow data output of the referenced workflow will be used as the result data of the action.
+If it is set to "false" the parent workflow can continue its execution as soon as the referenced sub-workflow
```
jorgenj, line 2527: From the meeting discussion, this section here talks about asynch sub-flow & how the parent workflow should behave in that case.

tsurdilo (approved):

PR #336, opened 2021-04-12
[WIP] Create an ActionDefinition schema to support main action types

cdavernas:

### **WARNING: WIP**

- [] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

Creates a new ActionDefinition, that allows for different activities:

- **Functions**: Invokes a function
- **Async Functions**: Invokes an async function. _(1)_
- **Subflows**: Invokes a subflow
- **Consume Event**: consumes a single cloud event
- **Produce Event**: produces a single cloud event

@tsurdilo Here is a very basic concept illustrating the discussion we had last week.

_(1)_. For there is no support the the AsyncApi spec, the same schema than EventRef is currently used. 
Ideally, should add the possibility to handle AsyncApi spec.


Neuroglia:
@tsurdilo Where should I put the required examples?

tsurdilo:
@Neuroglia sorry i dont understand the question.
Also regarding this pr - it seems as it wants to externalize action definitions to a separate file as its done with function/events/retries definitions. I think this is probably not optimal for actions as they include information which is not really reusable (meaning its tied to a specific workflow implementation). Could you explain to me this pr in case I am missing a point please?

Neuroglia:
@tsurdilo In the PR requirements, there is the Examples task, which I would love to collaborate to. I don't know however where to put those examples. Should I edit the [examples MD file](https://github.com/serverlessworkflow/specification/tree/main/examples) or should I add single yaml/json example files?

> it seems as it wants to externalize action definitions to a separate file as its done with function/events/retries

I do not really wish to externalize actions definitions, putting the code in single, external file sounded easier to read and work on the schema for the time being.

tsurdilo:
@Neuroglia oh ok i get it now :) I think you could add separate json/yaml files for discussion to this pr thats fine.

tsurdilo:
@Neuroglia if it makes it easier we could also try to use the recently added discussions https://github.com/serverlessworkflow/specification/discussions ;)

tsurdilo:
I like this idea very much . 
To describe the issue - we use "waitForCompletion" currently to define sync/async invocation
but our DSL cannot currently say "invoke function async and then wait for its results" ... for that we currently say if async invocation we cannot put its results back into state data.

I think with this we can definitely use "waitForCompletion" the way its intended to be , meaning that it makes things more clear
- if you want to invoke your function async use the "async" action type. If you want to wait for its results before proceeding use the "waitForCompletion" .. to me this makes sense. WDYT?

PR #338, opened 2021-04-12
Clarify the format of Schedule `interval` and its relationship to ISO 8601

### LINKED ISSUE #322: Schedule `interval` field "R/<Duration>" pattern is not sufficiently defined
The [Schedule](https://github.com/serverlessworkflow/specification/blob/main/specification.md#schedule-definition) `interval` field is defined as "ISO 8601 format", but ISO 8601 does not include any pattern like **R/\<Duration>**:
> **5.6.1 Means of specifying recurring time intervals**
> A recurring time interval shall be expressed in one of the following ways.
> 
> a) by a number of recurrences (optional), followed by a start and an end which identify the first time interval [_**R[\<Number>]/\<Start>/\<End>**, which is effectively useless unless the duration is short and within the same calendar day because not every day is the same length_];
> 
> b) by a number of recurrences (optional), followed by a start and a duration which identify the first time interval [_**R[\<Number>]/\<Start>/\<Duration>**_];
> 
> c) by a number of recurrences (optional), followed by a duration and an end which identify the last time interval > [_**R[\<Number>]/\<Duration>/\<End>**_]. 
> 
> In all cases, if the number of recurrences is absent, the number of occurrences is unbounded.
> 
> NOTE 1  In some applications, a recurring time interval is expressed by a number of recurrences (optional) and duration alone, because either the start or end is supplied out-of-band.

The field should be defined more precisely, and if it is important to include such syntax then it should do so clearly (unlike ISO 8601 itself, in which the note _implies_ that such a representation should be possible but doesn't explicitly define how), and specifically define how the start or end time is established.

gibson042:

- [x] Specification
- [ ] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

Fixes #322 by more clearly describing the format of Schedule `interval` (including both restrictions and extensions of ISO 8601) and expectations for how start time is established when it is unspecified.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -2227,7 +2227,7 @@ This is visualized in the diagram below:
 
 The event state timeout period is described in the ISO 8601 data and time format.
 You can specify for example "PT15M" to represent 15 minutes or "P2DT3H4M" to represent 2 days, 3 hours and 4 minutes.
-Timeout values should always be represented as durations and not as time/repeating intervals.
+Timeout values should always be represented as durations and not as specific time intervals.
 
 The timeout property needs to be described in detail as it depends on whether or not the Event state is a workflow starting
 state or not.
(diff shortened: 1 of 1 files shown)


tsurdilo (approved):

tsurdilo:
@gibson042 thank you for the PR!
The DCO action is failing because the commit is not signed. If possible could you sign it? 
(git commit **-s** -m ".....")

ricardozanini (approved):

PR #342, opened 2021-04-13
Adding subflow action version property

tsurdilo:

- [ x] Specification
- [x ] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -1385,7 +1385,7 @@ definition "id" must be a constant value.
 | key | Domain-specific workflow identifier | string | yes if `id` not defined |
 | name | Workflow name | string | yes |
 | description | Workflow description | string | no |
-| version | Workflow version (default value is `"latest"`) | string | no |
+| version | Workflow version | string | no |
 | annotations | List of helpful terms describing the workflows intended purpose, subject areas, or other important qualities | string | no |
 | dataInputSchema | Used to validate the workflow data input against a defined JSON Schema| string or object | no |
 | [start](#Start-Definition) | Workflow start definition | string | yes |
@@ -1465,8 +1465,7 @@ The `name` property is the workflow logical name.
 
 The `description` property can be used to give further information about the workflow.
 
-The `version` property can be used to provide a specific workflow version. The default
-value of this property (if not specified) is `"latest"`.
+The `version` property can be used to provide a specific workflow version.
 
 The `annotations` property defines a list of helpful terms describing the workflows intended purpose, subject areas, or other important qualities,
 for example "machine learning", "monitoring", "networking", etc
--- a/specification.md
+++ b/specification.md
@@ -29,6 +29,7 @@ You can find the specification roadmap [here](roadmap/README.md).
     - [Defining Errors](#Defining-Errors)
     - [Defining Retries](#Defining-Retries)
   - [Workflow Compensation](#Workflow-Compensation)
+  - [Workflow Versioning](#Workflow-Versioning)  
   - [Workflow Metadata](#Workflow-Metadata)
 - [Extensions](#Extensions)
 - [Use Cases](#Use-Cases)
@@ -1386,7 +1387,7 @@ definition "id" must be a constant value.
 | annotations | List of helpful terms describing the workflows intended purpose, subject areas, or other important qualities | string | no |
 | dataInputSchema | Used to validate the workflow data input against a defined JSON Schema| string or object | no |
 | [start](#Start-Definition) | Workflow start definition | string | yes |
-| schemaVersion | Workflow schema version | string | no |
+| schemaVersion | Workflow schema version (default value is `"latest"`) | string | no |
 | expressionLang | Identifies the expression language used for workflow expressions. Default value is "jq" | string | no |
 | [execTimeout](#ExecTimeout-Definition) | Defines the execution timeout for a workflow instance | object | no |
 | keepActive | If "true", workflow instances is not terminated when there are no active execution paths. Instance can be terminated with "terminate end definition" or reaching defined "execTimeout" | boolean | no |
@@ -1490,7 +1491,11 @@ a starting [Event state](#Event-state), it is not used to validate its event pay
 The `start` property defines the workflow starting information. For more information see the [start definition](#Start-Definition) section.
 
 The `schemaVersion` property can be used to set the specific Serverless Workflow schema version to use
-to validate this workflow markup. If not provided the latest released schema version is assumed.
+to validate this workflow markup. The default value of this property is `"latest"`, meaning that 
+if it is not specified, the latest specification release schema version is assumed.
+If `schemaVersion` is defined, it must follow the specification release versions (excluding the leading "v"), meaning that for 
+the [release version v0.6](https://github.com/serverlessworkflow/specification/releases/tag/v0.6) 
+its value should be set to `"0.6"`.
 
 The `expressionLang` property can be used to identify the expression language used for all expressions in
 the workflow definition. The default value of this property is ["jq"](https://stedolan.github.io/jq/). 
@@ -4831,6 +4836,32 @@ States that are marked as `usedForCompensation` can define [error handling](#Wor
 (errors not explicitly handled),
 workflow execution should be stopped, which is the same behavior as when not using compensation as well. 
 
+### Workflow Versioning
+
+In any application, regardless of size or type, one thing is for sure: changes happen.
+Versioning your workflow definitions is an important task to consider. Versions indicate 
+changes or updates to your workflow definitions to the associated execution runtimes. 
+
+There are three places in the [workflow definition](#Workflow-Definition-Structure) where versioning can be applied:
+1. Top level workflow definition `schemaVersion` property.
+2. Top level workflow definition `version` property.
+3. Actions [subflowRef](#SubFlowRef-Definition) `version` property.
+
+The top-level `schemaVersion` property follows the release versioning strategy of the specification (without the leading "v" 
+if the specification release includes it). 
+To give an example, to reference the  specification
+[release version v0.6](https://github.com/serverlessworkflow/specification/releases/tag/v0.6) its value should be set to `"0.6"`.
+The default value of `schemaVersion` is set to "latest", meaning that if this property is not defined, the workflow definition
+conforms to the latest specification release version.
+
+The Serverless Workflow specification does not mandate a specific versioning strategy
+for the top level and actions subflowRef definitions `version` property. It does not mandate the use 
+of a versioning strategy at all. We do recommend however that you do use a versioning strategy 
+for your workflow definitions especially in production environments. 
+
+To enhance portability when using versioning of your workflow and sub-workflow definitions,
+we recommend using an existing versioning standard such as [SemVer](https://semver.org/) for example.
+ 
 ### Workflow Metadata
 
 Metadata enables you to enrich the serverless workflow model with information beyond its core definitions.
--- a/specification.md
+++ b/specification.md
@@ -4840,7 +4840,7 @@ workflow execution should be stopped, which is the same behavior as when not usi
 
 In any application, regardless of size or type, one thing is for sure: changes happen.
 Versioning your workflow definitions is an important task to consider. Versions indicate 
-changes or updates to your workflow definitions to the associated execution runtimes. 
+changes or updates of your workflow definitions to the associated execution runtimes. 
 
 There are three places in the [workflow definition](#Workflow-Definition-Structure) where versioning can be applied:
 1. Top level workflow definition `schemaVersion` property.
--- a/specification.md
+++ b/specification.md
@@ -113,7 +113,7 @@ and the [Workflow Model](#Workflow-Model) It defines a blueprint used by runtime
 
 A business solution can be composed of any number of related workflow definitions.
 Their relationships are explicitly modeled with the Serverless Workflow language (for example
-by using [SubFlow](#SubFlow-Action) actions).
+by using [SubFlowRef Definition](#SubFlowRef-Definition) in actions).
 
 Runtimes can initialize workflow definitions for some particular set of data inputs or events
 which forms [workflow instances](#Workflow-Instance).
@@ -546,6 +546,7 @@ a workflow with a single event state and show how data filters can be combined.
     "id": "GreetCustomersWorkflow",
     "name": "Greet Customers when they arrive",
     "version": "1.0",
+    "specVersion": "0.6",
     "start": "WaitForCustomerToArrive",
     "states":[
          {
@@ -1108,7 +1109,7 @@ Let's take a look at an example of such definitions:
 Here we define two reusable expression functions. Expressions in Serverless Workflow
 can be evaluated against the workflow, or workflow state data. Note that different data filters play a big role as to which parts of the 
 workflow data are being evaluated by the expressions. Reference the 
-[State Data Filtering](#State-Data-Filtering) section for more information on this.
+[State Data Filters](#State-data-filters) section for more information on this.
 
 Our expression function definitions can now be referenced by workflow states when they need to be evaluated. For example:
 
@@ -1367,7 +1368,8 @@ we can use this expression in the workflow "version" parameter:
 {
    "id": "MySampleWorkflow",
    "name": "Sample Workflow",
-   "version": "${ .inputVersion }"
+   "version": "${ .inputVersion }",
+   "specVersion": "0.6"
 }
 ```
 
@@ -1387,7 +1389,7 @@ definition "id" must be a constant value.
 | annotations | List of helpful terms describing the workflows intended purpose, subject areas, or other important qualities | string | no |
 | dataInputSchema | Used to validate the workflow data input against a defined JSON Schema| string or object | no |
 | [start](#Start-Definition) | Workflow start definition | string | yes |
-| specVersion | Serverless Workflow specification release version (default value is `"latest"`) | string | no |
+| specVersion | Serverless Workflow specification release version | string | yes |
 | expressionLang | Identifies the expression language used for workflow expressions. Default value is "jq" | string | no |
 | [execTimeout](#ExecTimeout-Definition) | Defines the execution timeout for a workflow instance | object | no |
 | keepActive | If "true", workflow instances is not terminated when there are no active execution paths. Instance can be terminated with "terminate end definition" or reaching defined "execTimeout" | boolean | no |
@@ -1412,6 +1414,7 @@ definition "id" must be a constant value.
 {  
    "id": "sampleWorkflow",
    "version": "1.0",
+   "specVersion": "0.6",
    "name": "Sample Workflow",
    "description": "Sample Workflow",
    "start": "MyStartingState",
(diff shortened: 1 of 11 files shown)


tsurdilo:
@jorgenj i updated some text around subflow action definition and fixed some property default values just fyi

specification.md
```diff
@@ -2508,10 +2508,25 @@ to the trigger/produced event.
 
 #### SubFlowRef Definition
 
+`subFlowRef` definition can have two types, either `string` or `object`.
+If `string` type, it defines the unique id of the sub-workflow to be invoked. 
+This short-hand definition can be used if sub-workflow lookup is done only by its `id`
+property and not it's `version` property and if the default value of `waitForCompletion` is assumed.
```
mswiderski, line 2514: small type it's` should be replaced with `its`

specification.md
```diff
@@ -2508,10 +2508,25 @@ to the trigger/produced event.
...
+```
+
+If you need to define the `version` and `waitForCompletion` parameters in your `subFlowRef` definition, you can define
+it with its `object` type which has the following properties:
+
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
-| waitForCompletion | If workflow execution must wait for sub-workflow to finish before continuing | boolean | yes |
-| workflowId |Sub-workflow unique id | boolean | no |
+| workflowId | Sub-workflow unique id | boolean | no |
```
mswiderski, line 2527: I believe workflowId is a required attribute so it should be set to yes in the last column

specification.md
```diff
@@ -2543,12 +2565,20 @@ workflowId: handleApprovedVisaWorkflowID
 
 </details>
 
+The `workflowId` property defined the unique ID of the sub-workflow to be invoked.
+
+The `version` property defined the unique version of the sub-workflow to be invoked.
+If this property is defined, runtimes should match both the `id` and the `version` properties
+defined in the sub-workflow definition. If it is not defined, the lookup should be done by
+`workflowId` property alone. In the case of multiple sub-workflows definitions that match this
+lookup, runtimes should report this as errors.
```
jorgenj, line 2574: Nit; can we say that the runtime should define the behavior when the version is ambiguous?  For example, some might have a concept of the latest or 'production' version of a workflow and default to that when no version is specified?

jorgenj (approved):

specification.md
mswiderski: I would still prefer to make it explicit as will make it more portable in my opinion. Leaving the version resolution to the runtimes is fine but still the version should be specified and it could be like meta version as you said (`latest` etc)

The spec could mention use of `latest` or `default` versions but described that they should be explicitly defined in the workflow

Neuroglia: @jorgenj @mswiderski Shouldn't the version, if not specified, default to latest? That's the actual behavior of my runtime. Using 'default or  'latest seems fine to me too, but having this field as required doesn't feel totally right, especially if I compare to, say, Docker or K8s image tags...

mswiderski: in general I'd be fine with it as long as it goes by the spec meaning spec defines version on subflow as latest if not given so it's explicit. Rather if not set it is runtime specific as it makes it less portable. what do you think?

Neuroglia: @mswiderski Yes, you are right, a a default explicit value is IMHO the way to go. It leaves the possibility of not setting anything on legacy subflows, though guaranteeing consistant behavior accross runtimes.

jorgenj: It seems that `latest` might be a potentially dangerous choice.  Consider the case where devs are uploading new versions of their workflow for testing, if they forgot to specify a version in the parent workflow, then their 'production' workflows might start spawning 'dev' (unapproved) versions of the subflow.

I think `default` is reasonable, but I'm not sure exactly how `default`'s behavior would be defined at this point other than to say it can be runtime-specific (unless this spec is now going to get into the topic of managing versions, like 'approving' a new version to be the `default` or something like that).

tsurdilo: @mswiderski @Neuroglia @jorgenj 
I do think that defining "latest" covers us from the spec perspective but as @jorgenj mentioned that could lead to issues due to possible mistakes. Looking at it only from the spec perspective it does seem adding a default value of "latest" is fine for portability.

mswiderski: agreed. Potential user mistakes are unavoidable anyway and I don't think the spec can do much about it. Enforcing version (either to be explicitly given or defined as default) should be enough in my opinion.

jorgenj: Avoiding user mistakes is one thing, but making the default behavior something inherently unsafe like 'latest' (use whatever happens to have been uploaded last for production workflows?) is just guaranteeing that many users will make that mistake.

ricardozanini (approved):

specification.md
mswiderski: isn't that exactly what container images do by default (as pointed by @Neuroglia)? If you don't specify the tag it assumes `latest` .I think people are already used to this behavior and once spec clearly states it there should not be any issues with that.

tsurdilo:
@mswiderski @Neuroglia @jorgenj @ricardozanini et all:

added "latest" default value for subflowref definition.

would like to ask a question -- we also have the top-level workflow "version" property which does not contain a default value. 
given that sub-workflows are also defined as workflows (workflow definition) should we even consider adding a default value to this top-level version property? I would say no but wanted to hear everyones opinion

tsurdilo:
@mswiderski @Neuroglia @jorgenj @ricardozanini et all:

in addition, this type of lookup (with "latest") opens up a question to me about 
1. Function definitions defined in separate json/yaml files
2. Event definitions defined in separate json/yaml files
3. Retry definitions defined in separate json/yaml files

Currently the schemas for those do not include a "version" parameter and it is assume that there is "only one". but if we go this route i feel that we should make this change across the board for all "resource references" and not just sub-workflows.
WDYT?

specification.md
jorgenj: In my experience the generally accepted best-practice is "don't use latest for deployed code" [[1](https://vsupalov.com/docker-latest-tag/)] 

From the [docker blog on best practices](https://docs.docker.com/develop/dev-best-practices/):

> When building images, always tag them with useful tags which codify version information, intended destination (prod or test, for instance), stability, or other information that is useful when deploying the application in different environments. Do not rely on the automatically-created latest tag.

We can also look to Maven's snapshot feature as a case-study with similar results, as far as I'm aware the general advice is "don't use SNAPSHOT for deployed code".

specification.md
```diff
@@ -2508,10 +2508,25 @@ to the trigger/produced event.
...
 #### SubFlowRef Definition
 
+`subFlowRef` definition can have two types, either `string` or `object`.
+If `string` type, it defines the unique id of the sub-workflow to be invoked. 
+This short-hand definition can be used if sub-workflow lookup is done only by its `id`
+property and not it's `version` property and if the default value of `waitForCompletion` is assumed.
+
+```json
+{
+  "subFlowRef": "uniqueSubFlowId"
```
jorgenj, line 2518: Should we consider a mechanism for including the version here in the short-hand?  IMHO, best-practices for production workflows should be to always specify the version (for repeatability), so that would preclude production workflows from using this short-hand if so.

How about allowing the format `<workflowId>:<version>`?  This would obviously require that the workflow id now allow whatever delimiter character is chosen, to avoid confusion.

specification.md
jorgenj:
```suggestion
| workflowId | Sub-workflow unique id | boolean | yes |
```

specification.md
tsurdilo: yeah, since we dont enforce any specific version strategy doing this can be pretty tricky to do

specification.md
mswiderski: I agree that the latest version should not be considered to be used for production but it does make sense to have it in the same way it is for container images - recommended to use concrete versions. So having it defined by the spec will make runtimes more aligned.

specification.md
mswiderski: +1 on having concrete version as recommendation as it makes it explicit and you can directly know what is behind. 

I wouldn’t go for id:version syntax as this is easy to miss. Having it as separate attribute is in my opinion the way to go. 

And lastly about version on top level workflow I’d go with it the same as subflows to be consistent.

mswiderski:
@tsurdilo i am not convinced we want to version functions, events and retries. Sounds to me a bit too much. Personally I think they are easier to evolve and usually won’t have a state (like workflows can have) that makes it harder to deal without versioning.

specification.md
mswiderski: @tsurdilo has this been applied?

mswiderski:
Last question - should the spec be concerned how to resolve latest? Meaning to  define version comparison or it should be left off?

tsurdilo:
@mswiderski @jorgenj et all:
1. applied the small fixes to properties 
2. defined default value for top level "version" and "schemaVersion" properties
3. added a "Workflow Versioning" section to spec doc

let me know...

mswiderski (approved):

specification.md
```diff
@@ -1489,7 +1491,11 @@ a starting [Event state](#Event-state), it is not used to validate its event pay
 The `start` property defines the workflow starting information. For more information see the [start definition](#Start-Definition) section.
 
 The `schemaVersion` property can be used to set the specific Serverless Workflow schema version to use
-to validate this workflow markup. If not provided the latest released schema version is assumed.
+to validate this workflow markup. The default value of this property is `"latest"`, meaning that
```
jorgenj, line 1494: What happens to workflows which don't specify a version when a new version of the schema is released with new required parameters?

specification.md
jorgenj: @mswiderski 
> +1 on having concrete version as recommendation as it makes it explicit and you can directly know what is behind.

I'm not sure if I'm misinterpreting what you're saying, but it seems like you're saying that the version property should be required?  If so, I agree.  I think that's preferable to having it be optional if the default value is expected to be 'latest'.

> And lastly about version on top level workflow I’d go with it the same as subflows to be consistent.

I'm not following what you meant here, could you expand?

specification.md
```diff
@@ -2543,12 +2571,26 @@ workflowId: handleApprovedVisaWorkflowID
...
+an error.
+
 The `waitForCompletion` property defines if the SubFlow action should wait until the referenced reusable workflow
 has completed its execution. If it's set to "true" (default value), SubFlow action execution must wait until the referenced workflow has completed its execution.
 In this case the workflow data output of the referenced workflow will be used as the result data of the action.
-If it is set to "false" the parent workflow can continue its execution as soon as the referenced sub-workflow 
-has been invoked (fire-and-forget). For this case, the referenced (child) workflow data output will be ignored and the result data 
-of the action will be an empty json object (`{}`).
+In the case it's set to "false" the SubFlow action can finish its execution as soon as the referenced sub-workflow 
+has been invoked (fire-and-forget). For this case, the referenced sub-workflow data output should be ignored and the result data
```
jorgenj, line 2592: > output should be ignored

Nit; in RFC's they're very careful about the words 'should' vs. 'must'.  I think we must avoid the word 'should' (in general, but here as a particular example) as it could be construed as something that is a suggestion and not a requirement of the spec.

specification.md
```diff
@@ -4768,6 +4810,32 @@ States that are marked as `usedForCompensation` can define [error handling](#Wor
 (errors not explicitly handled),
 workflow execution should be stopped, which is the same behavior as when not using compensation as well. 
 
+### Workflow Versioning
```
jorgenj, line 4813: WDYT about separating out the bits about schemaVersion from workflow versions?   It seems odd that these are lumped together as one concept.

specification.md
mswiderski: I prefer the spec does not enforce versions to be used as not always there is a need to version your workflows and there should be really good reason to change versions as in case of stateful workflows this can become quite a headache. So having it recommended to be explicitly given if you want to use versions and defaults to `latest` (that in fact will be resolved by runtimes) is the middle ground in my opinion. The same should apply to both top level workflows and sub flows.

jorgenj: @mswiderski 

> defaults to `latest` (that in fact will be resolved by runtimes)

It sounds like you're saying that you're ok with the runtimes determining their definition of 'latest'.  In that case, I'd be fine with that if we could choose a term other than 'latest', given that 'latest' implies a particular strategy for resolving the version.

Perhaps 'live' or 'default'?

specification.md
jorgenj: Or rather, I think we should be careful about the use of words like "should" vs. "must".  In this case, this seems like a "must" rather than a "should".  WDYT?

specification.md
tsurdilo: +1 will update

specification.md
mswiderski: in general we could use something else but I think latest is sort of well established especially for cloud targeting deployments e.g. containers. But I don't have strong preference to keep latest or something else. Maybe except the default which we will have same issue as it is reserver word in many languages.

specification.md
mswiderski: from runtimes point of view version of the spec is more likely not that important as usually they will bind to sdk version is in use. But I am starting to think that maybe schemaVersion should be a required attribute to avoid any confusion.

specification.md
jorgenj: I don't know for sure, but 'default' in this context is a value (as opposed to an attribute name) so I don't think there would be a problem with reserved-word conflicts as this would just be the string "default", right?

mswiderski: yes, you're right, with it as a value is less of an issue.

specification.md
tsurdilo: yeah +1 on making it required, but i would like to change it to "specVersion" so its easier to understand. 
this follows also what CE is doing with their "specversion" context attribute

specification.md
tsurdilo: there is https://tools.ietf.org/html/rfc2119
we are planning to update our documentation to use it

tsurdilo:
* renamed "schemaVersion" to "specVersion"
* updated the workflow versioning section to not include info about specVersion as its already defined

specification.md
```diff
@@ -4768,6 +4810,25 @@ States that are marked as `usedForCompensation` can define [error handling](#Wor
...
+
+1. Top level workflow definition `version` property.
+2. Actions [subflowRef](#SubFlowRef-Definition) `version` property.
+
+The Serverless Workflow specification does not mandate a specific versioning strategy
+for the top level and actions subflowRef definitions `version` properties. It does not mandate the use 
+of a versioning strategy at all. We do recommend however that you do use a versioning strategy 
+for your workflow definitions especially in production environments. 
+
+To enhance portability when using versioning of your workflow and sub-workflow definitions,
```
jorgenj, line 4829: As far as I can tell, so far the spec is avoiding any assumptions on the versioning scheme.   How does the users choice of versioning scheme affect portability if runtimes are supposed to be agnostic to the versioning scheme?

specification.md
```diff
@@ -1488,8 +1490,12 @@ a starting [Event state](#Event-state), it is not used to validate its event pay
 
 The `start` property defines the workflow starting information. For more information see the [start definition](#Start-Definition) section.
 
-The `schemaVersion` property can be used to set the specific Serverless Workflow schema version to use
-to validate this workflow markup. If not provided the latest released schema version is assumed.
+The `specVersion` property can be used to set the specific Serverless Workflow specification release version 
+the workflow markup adheres to. The default value of this property is `"latest"`, meaning that
```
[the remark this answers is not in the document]
tsurdilo: well it means that if you have a workflow definition where you don't specify the specVersion and execute it on runtimes A and B, both would parse it against the latest specification release schemas. On our website we provide already:

https://serverlessworkflow.io/schemas/latest/workflow.json

so that can map the "latest" default keyword

specification.md
tsurdilo: i'm not really sure i get the question.

tsurdilo:
@jorgenj @mswiderski 
after reading the comments on this PR couple of times 
I think it might be the best choice to remove version property from the DSL.
If users want to version their workflows they can use the metadata property (or in @jorgenj case the additional properties I assume). 

Will keep specVersion with the default value of "latest" as that maps to what we are deploying on our website currently anyways: https://serverlessworkflow.io/schemas/latest/workflow.json 
This is updated on each release.

tsurdilo:
So I will update this pr to:
1) remove the top-level "version" property
2) remove the subflowRef "version" property
3) update the text to reflect that from the specification perspective versioning can be defined in the metadata section and is left for runtime impls to figure out 

Thanks

mswiderski:
I still think the version should be defined as first class citizen on the both top level workflow and subflow. If there should be default version or not is less of an issue but I lean towards having it set to latest to be in line with container image tags.

specification.md
jorgenj: I'm trying to understand how the users choice of versioning scheme's affects portability, I was hoping you could enlighten me on this.

tsurdilo:
@jorgenj @mswiderski ok after latest discussions here are the hopefully final updates:

1) no defaults for top-level "version" property
2) no defaults for subflowRef "version" property
3) "schemaVersion" changed to "specVersion" and made it a required parameter with no default value
4) updated all examples and documents to include specVersion: '0.6'

We just need to remember for every release now to update this value too.... :)

tsurdilo:
@Neuroglia @antmendoza: are you ok with these changes? This will affect all our examples in the SDKs as we need to add specVersion to each of the json/yaml and set its value to 0.6. just heads up :)

tsurdilo:
It would make our life as a whole a ton easier if we had a default value for "specVersion" and 
update only that default value in the workflow schema rather than having to:

1) update each sdk every time we release
2) update all examples in each sdk every time we release
3) update specversion on the website schemas each time we release
4) update all examples and specification documents/comparisons/whatever in main spec repo each time we release

Given that I have done releases alone so far...unless someone volunteers to help I would seriously consider 
having a default value for specVersion please :)

antmendoza:
no problem @tsurdilo

specification.md
jorgenj: @tsurdilo Not sure if you missed this question.

tsurdilo: The idea in the text is that if you use a standard such as SemVer rather than some custom versioning scheme, there is a higher probability chance it will be supported by more runtimes

mswiderski:
@tsurdilo you have my vote for the latest proposal

> no defaults for top-level "version" property
> no defaults for subflowRef "version" property
> "schemaVersion" changed to "specVersion" and made it a required parameter with no default value
> updated all examples and documents to include specVersion: '0.6'

jorgenj:
> @jorgenj @mswiderski ok after latest discussions here i made some changes:
> 
> 1. no defaults for top-level "version" property
> 2. no defaults for subflowRef "version" property
> 3. "schemaVersion" changed to "specVersion" and made it a required parameter with no default value
> 4. updated all examples and documents to include specVersion: '0.6'
> 
> We just need to remember for every release now to update this value too.... :)

Apologies for not responding sooner, I thought I had 👍  your comment but don't see it now.  I agree with this approach.

PR #343, opened 2021-04-13
Add explanation for additional properties

tsurdilo:

- [ x] Specification
- [ ] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

adds explanation of top-level additional properties


jorgenj (approved):

specification.md
```diff
@@ -1591,6 +1591,52 @@ This allows you to explicitly model workflows where an instance should be kept a
...
+  "name": "My Test Workflow",
+  "start": "My First State",
+  "loglevel": "Info",
+  "environment": "Production",
+  "category": "Sales",
+  "states": [ ... ]
+}
+```
+
+In this example we specify the "loglevel", "environment", and "category" additional properties.
```
ricardozanini, line 1621:
```suggestion
In this example, we specify the `loglevel`, `environment`, and `category` additional properties.
```

ricardozanini (commented):
Dumb question: if `metadata` is the preferred approach (and does the **same thing**), why supporting additional properties like this? I don't think it's a good idea. How would our SDKs support it? Implementations would have to parse themselves, or will we add them in maps as we do with metadata? 

Anyhow, this can be a can of worms in the future, and I'm curious to see how this would be portable.

Neuroglia:
@ricardozanini I tend to agree with your last comment: I too fail to see the added value of top level properties as opposed to metadata. Actually, IMHO, it even is an invitation to do things the 'wrong way', meaning relying, as you said, on runtime-specific parsers for data that is not even supposed to be there in the first place. 
Am I missing something?

tsurdilo:
@ricardozanini @Neuroglia 
Not a dumb question at all, and in general yes you are correct that this kinda opens up a can of worms. 
The best way I can describe this is following (and @jorgenj please chime in as well as iirc you have a real use case where this is needed):
As we are a declarative dsl our workflows are described in json/yaml which means that typically this language is translated by runtimes into executable code. The workflow definitions themselves are considered resources (similar to let's say CRDs, images, UI, etc) which can be managed by some CMS (content management systems) responsible for all application resources.
Some of these CMS add specific "metadata" to each resource, meaning they kinda "brand" them so it becomes a managed resource for them, and in these scenarios they would have to know our specific language structure in order to push this automatic metadata to our "metadata" object. 
To make things simpler we are adding this ability and clearly stating that this is not the preferred way of doing things as well as that it is not portable/no ability to get to this data from our SDKs. 

On the other hand this is also no something really new. OpenApi for example also has some limited support for additionalProperties, maybe even for these types of use cases, idk.

jorgenj:
That covers it pretty well.  I just want to add that this spec only talks about the workflow language itself, but runtimes won't be implementing just the language they'll also potentially be implementing API's around managing these workflows as resources and there are a lot of small details related to that which aren't generally represented by this spec (since it doesn't cover APIs).

For example, it's pretty common practice to put basic timestamps on resources (createdAt, updatedAt) and if your workflow-definition API is just one in a large number of other APIs you support, you probably need to make that API as consistent as possible with all your other apis (in this case, put those timestamps at the top-level, not nesting them under metadata which none of the rest of your APIs are likely to have).

As stated in the meeting last week, it seems reasonable for the SDK's to just ignore properties they don't know about for the particular version of the spec schema they're parsing.

tsurdilo:
looking at the cloudevents schema they dont seem to care about additionalProperties at all ;) https://github.com/cloudevents/spec/blob/v1.0.1/spec.json
so i guess you can add whatever you want there

ricardozanini:
> For example, it's pretty common practice to put basic timestamps on resources (createdAt, updatedAt) and if your workflow-definition API is just one in a large number of other APIs you support, you probably need to make that API as consistent as possible with all your other apis (in this case, put those timestamps at the top-level, not nesting them under metadata which none of the rest of your APIs are likely to have).

Seems fine to me :+1: 

> As stated in the meeting last week, it seems reasonable for the SDK's to just ignore properties they don't know about for the particular version of the spec schema they're parsing.

From the SDK point of view, if we ignore it, the SDK won't be useful anymore since the workflow itself won't be parsed completely. We can add a `additionaProperties` map to handle such cases.

tsurdilo:
approved in meeting

PR #248, opened 2021-02-08
removing the transition restricting section from spec doc

tsurdilo:
Signed-off-by: Tihomir Surdilovic <tsurdilo@redhat.com>

**Many thanks for submitting your Pull Request :heart:!**

**Please specify parts this PR updates:**

- [ x] Specification
- [ ] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

**What this PR does / why we need it**:
removes the transition restriction section from the spec doc. we will add this back when we have it fully defined
**Special notes for reviewers**:

**Additional information (if needed):**

PR #349, opened 2021-04-16
Update eventDataFilter to call out that the entire event can be accessed

jorgenj:
Signed-off-by: Jorgen Johnson <jorgen.johnson@oracle.com>

- [ ✅  ] Specification
- [ ✅  ] Schema
- [ ✅  ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

Add some clarity on what eventDataFilter can access from the event. 

If you look at the examples that were updated, there were several that were already assuming that the whole event was being added to stateData, hence why there are several examples where I added `data: ${ . }`.

It seems that a good number of the examples were already acting as if this was the case that the whole event would be filtered (not just the payload).  I think, given the number of examples that were assuming this, that we should reconsider whether the default 'data' value should be `${.data}` or just `${ . }`.

### DIFF
--- a/comparisons/comparison-brigade.md
+++ b/comparisons/comparison-brigade.md
@@ -398,6 +398,7 @@ start: LogEventData
 events:
 - name: execEvent
   type: exec
+  dataOnly: false
 functions:
 - name: consoleFunction
   type: console
@@ -408,7 +409,7 @@ states:
   - eventRefs:
     - execEvent
     eventDataFilter:
-      results: "${ .event }"
+      toStateData: "${ .event }"
     actions:
     - name: eventInfoAction
       functionRef:
@@ -490,7 +491,7 @@ states:
   - eventRefs:
     - execEvent
     eventDataFilter:
-      data: "${ .event }"
+      toStateData: "${ .event }"
     actions:
     - name: helloAction
       actionDataFilter:
@@ -565,6 +566,7 @@ start: ExecEventState
 events:
 - name: execEvent
   type: exec
+  dataOnly: false
 - name: nextEvent
   type: next
   kind: produced
@@ -579,7 +581,7 @@ states:
     - execEvent
     actions: []
     eventDataFilter:
-      data: "${ .execEvent }"
+      toStateData: "${ .execEvent }"
   transition:
     nextState: NextEventState
     produceEvents:
@@ -597,7 +599,7 @@ states:
   - eventRefs:
     - nextEvent
     eventDataFilter:
-      data: "${ .nextEvent }"
+      toStateData: "${ .nextEvent }"
     actions:
     - name: consoleLogAction
       functionRef:
--- a/examples/README.md
+++ b/examples/README.md
@@ -317,7 +317,8 @@ filters what is selected to be the state data output which then becomes the work
      "onEvents": [{
          "eventRefs": ["GreetingEvent"],
          "eventDataFilter": {
-            "data": "${ .data.greet }"
+            "data": "${ .greet }",
+            "toStateData": "${ .greet }"
          },
          "actions":[  
             {  
@@ -363,7 +364,8 @@ states:
   - eventRefs:
     - GreetingEvent
     eventDataFilter:
-      data: "${ .data.greet }"
+      data: "${ .greet }"
+      toStateData: "${ .greet }"
     actions:
     - functionRef:
         refName: greetingFunction
@@ -3224,7 +3226,7 @@ the data for an hour, send report, and so on.
             }
           ],
           "eventDataFilter": {
-            "data": "${ .readings }"
+            "toStateData": "${ .readings }"
           }
         }
       ],
@@ -3307,7 +3309,7 @@ states:
     - functionRef:
         refName: LogReading
     eventDataFilter:
-      data: "${ .readings }"
+      toStateData: "${ .readings }"
   end: true
 - name: GenerateReport
   type: operation
--- a/roadmap/README.md
+++ b/roadmap/README.md
@@ -31,6 +31,7 @@ _Status description:_
 | ✔️| Add subflow actions `version` property | [spec doc](../specification.md) |
 | ✔️| Renamed `schemaVersion` to `specVersion` and it is now a required parameter | [spec doc](../specification.md) |
 | ✔️| Add GraphQL support for function definitions | [spec doc](../specification.md) |
+| ✔️| Allow eventDataFilter to access entire event | [spec doc](../specification.md) |
 | 🚩 | Workflow invocation bindings |  |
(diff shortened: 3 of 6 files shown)


jorgenj:
Related to #248

jorgenj:
As discussed in the community meeting, I'll update this to say that the default is the whole event (rather than just the payload).

tsurdilo:
@jorgenj i cant think of a better name for it :) we can ask in meet if anyone has a good one

ricardozanini (approved):

tsurdilo:
I am still kinda weird about this change.

In our correlation rules we limit the expressions to context attributes only (no data). there you can specify that 
for example you only want events with the starting id of like "abc-*" for example. (or should at least we can add if not possible" as well as define the source and the specversion etc. 

Now let's say that you run the workflow on 2 different runtimes , which support 2 different release versions of cloudEvents.
It is MUCH easier to just have to then modify the reusable event definitions, than having to update all your workflow control flow logic expressions.
The fact is is that if we do allow CE context attributes to be used inside control flow logic, people will use it and that to me lowers the portability.

I still don't have a single concrete example workflow where you would care about the CE context attributes and not just the payload. All the non-domain-specific things should be defined in your event definition, and the control-flow logic which executes your business logic should only care about the data part. 

@mswiderski wdyt about this ?

Neuroglia:
@tsurdilo I totally agree with you, and too fail to see the added value of the proposed changed. In addition, Cloud Event is to my understanding a simple enveloppe, something that is related to the transfer/routing/correlation aspect of the data, just like an HTTP request. As the spec does not allow explicitly accessing the context of any kind of other application protocols (HTTP, GRPC, ...), IMHO the context attributes should not be made available for data-related operation.

mswiderski:
I also agree that data passed to the workflow instance (to its state) should only contain payload of the event. Event itself should be available for processing of the event like matching on attributes etc. regardless of the protocol used then can be accessed and used but should not be part of the state of workflow instance. With that it should not be required to prefix access to state data with `data`

jorgenj:
> @tsurdilo I totally agree with you, and I too fail to see the added value of the proposed changed. In addition, Cloud Event is to my understanding a simple enveloppe, something that is related to the transfer/routing/correlation aspect of the data, just like an HTTP request. As the spec does not allow explicitly accessing the context of any kind of other application protocols (HTTP, GRPC, ...), IMHO the context attributes should not be made available for data-related operation.

This was discussed in the meeting, it's not as simple as viewing the envelope as a 'transport protocol', for example the ID is a unique identifier of the event from the source system, which is extremely useful for workflows that may be themselves distributing events (for example the SAGA pattern) as that can be used as an idempotency token in the downstream services.   The counter-argument is that workflow users should just add their own arbitrary idempotency token to the event payload instead, which completely ignores the case where the user doesn't have control over what the event source adds to the payload of the event.

tsurdilo:
@mswiderski @Neuroglia -- @jorgenj  made good points about being able to in some cases add the entire event.  We talked in meet and will add a flag to allow users a choice if they want to include context attributes + payload info or not. 

@jorgenj and i promise won't change mind again :)

mswiderski:
Are these points described somewhere? Sorry couldn’t make it to the meeting

mswiderski:
having this extra parameter that will control if entire event or just payload should be given to the workflow sounds bit weird to me. Especially that when you then get data out of the workflow instance you get the entire event which will certainly not be desired in most of the cases. 

At the same time I do see the value in having access to some of that data but I am not convinced the workflow definition should be concerned with that. We should keep in mind that cloud events have different bindings and these bindings control how the metadata are given. Sometimes they are in protocol specific headers (http) or some message specific header (kafka) and so on   So it won’t be always as single json object and constructing that from various bindings might not be optimal to say the least.  

maybe instead we should define some kind of execution context that would provide access to that kind of data that are sort of request scoped. That way workflow can have access to that data at the time event is being consumed but won’t store it in the instance state. 

Very rough idea but what’s your thoughts on this approach?

jorgenj:
> We should keep in mind that cloud events have different bindings and these bindings control how the metadata are given. Sometimes they are in protocol specific headers (http) or some message specific header (kafka) and so on So it won’t be always as single json object and constructing that from various bindings might not be optimal to say the least.

This doesn't seem to have much bearing on how the event is exposed by the runtime to the workflow definition, if you look back at the original issue (#248), I believe @tsurdilo consulted with a cloud-events expert to confirm that in fact any cloud-event should be capable of being expressed as a JSON object regardless of what protocol it was transmitted via.  

"Not optimal" from what perspective?  The runtime will already be dealing with the potentially large payload, optionally making the relatively smaller envelope available as well seems like a trivial concern.

> maybe instead we should define some kind of execution context that would provide access to that kind of data that are sort of request scoped. That way workflow can have access to that data at the time event is being consumed but won’t store it in the instance state.

Isn't this exactly what this PR is proposing, that the whole event be made available at the time of event consumption (and that it's the workflow users prerogative which parts of the entire event be saved, as they see fit for their use-case)?

tsurdilo:
> Isn't this exactly what this PR is proposing, that the whole event be made available at the time of event consumption (and that it's the workflow users prerogative which parts of the entire event be saved, as they see fit for their use-case)?

yeah but it comes at a price which i think can be avoided:
1) yes ce can be transformed to json but now we have to enforce it
2) we are enforcing for every runtime to make both the envelope and the payload available. 

I am not saying to avoid this by not having this functionality but to think about how this can be solved while not enforcing 1) and 2) via the DSL

one thing that comes to mind is to force eventfilter definitions to use let's say an "envelope" and "payload" prefix to access each. That way a runtime could during parse time know that:
1) envelope info is needed
2) transform CE to json 

if that runtime is not capable of doing this it would be able to enforce the users during design/compile time of this error rather than at runtime.

tsurdilo:
What can we do to move this forward? Really would like to get this resolved soon if possible.

jorgenj:
> What can we do to move this forward? Really would like to get this resolved soon if possible.

@tsurdilo I posted an update with what was requested in the last community meeting but it's not clear to me if the goal-posts have been moved again since then or not.

manuelstein (approved):
lgtm

tsurdilo:
approved in team meet 5/10 . needs rebase

tsurdilo:
@jorgenj rebased and updated some text/descriptions so they are more clear.

tsurdilo:
@jorgenj ok relax :)
updated some text and fixed the readme

and by the way, typically you dont turn off something in boolean flags, its not natural. 
i think it would be better to do like "includeEventContext" or something and force users to set it to true to enable it, than but I will open a new PR on this don't worry :)

PR #364, opened 2021-05-01
Add GraphQL service support

lynxnathan:

- [x] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

Adds GraphQL service definition support.

For GraphQL, `arguments` have a more specific semantic, requiring a `selectionSet` property in order to respect the GraphQL specification that requires a selection of fields to always be present. What would usually be passed as the arguments goes into `inputs`.

Authentication is missing overall in the spec so I didn't include that. For GraphQL's case adding a third field to `arguments` such as `headers` should do the trick most of the time as well but in combination with mTLS and things like that this should be probably be further discussed.

Due to the current workflow model a GraphQL feature known as subscription (server push/keep-alive connection) is also not present.


specification.md
```diff
@@ -1075,9 +1077,167 @@ Note that the referenced function definition type in this case must be `rpc`.
 
 For more information about functions, reference the [Functions definitions](#Function-Definition) section.
 
+### Using Functions For GraphQL Service Invocations
+
+If you want to use GraphQL services, you can also invoke them using a similar syntax to the above methods with the caveat that your `functionRef`'s `arguments` property must have the following properties:
```
tsurdilo, line 1082: Nitpick - don't use "caveat" as it seems like a warning. Just describe what needs to be done :)

specification.md
```diff
@@ -1075,9 +1077,167 @@ Note that the referenced function definition type in this case must be `rpc`.
...
+
+| Parameter | Description | Type | Required |
+| --- | --- | --- | --- |
+| inputs | object whose keys are the field's arguments name, and its value that argument value | object | no |
```
tsurdilo, line 1086: Should we change maybe to:

"Input object containing field argument names and their values"

specification.md
```diff
@@ -1075,9 +1077,167 @@ Note that the referenced function definition type in this case must be `rpc`.
...
+| selectionSet | A string containing a valid GraphQL [selection set](https://spec.graphql.org/June2018/#sec-Selection-Sets) | string | yes |
```
tsurdilo, line 1087: Is this field really required? If not then please change last column to "no"

specification.md
```diff
@@ -1075,9 +1077,167 @@ Note that the referenced function definition type in this case must be `rpc`.
...
+
+And the outer `data` container object must be omitted from the response of successful queries and mutations, leaving only the selection set's specified fields. For exceptions the response must be returned in its entirety.
+
+We'll use the following [schema definition](https://graphql.org/learn/schema/) to show how that would work with both a query and a mutation:
```
tsurdilo, line 1091: following GraphQL schema definition 
(add GraphQL)

specification.md
```diff
@@ -1075,9 +1077,167 @@ Note that the referenced function definition type in this case must be `rpc`.
...
+  "arguments": {
+    "inputs": {
+      "id": 42
+    },
+    "selectionSet": "{ id, name, favoriteTreat { id } }"
+  }
+}
+```
+
+Which would return the following action state:
```
tsurdilo, line 1156: Could we add "GraphQL" in this sentence so that its not confused with SW "actions" and SW workflow "state"? 
Maybe just use "following results"?

specification.md
```diff
@@ -1075,9 +1077,167 @@ Note that the referenced function definition type in this case must be `rpc`.
...
+}
+
+input PetInput {
+    id: Int!
+    name: String!
+    favoriteTreatId: Int
+}
+```
+
+#### Executing a `Query`
```
tsurdilo, line 1120: Maybe use "Invoking a GraphQL Query"

specification.md
```diff
@@ -1075,9 +1077,167 @@ Note that the referenced function definition type in this case must be `rpc`.
...
+    "id": 42,
+    "name": "Snuffles",
+    "favoriteTreat": {
+      "id": 9001
+    }
+  }
+}
+```
+
+#### Executing a `Mutation`
```
tsurdilo, line 1170: Maybe use "Invoking a GraphQL Mutation"

specification.md
```diff
@@ -1075,9 +1077,167 @@ Note that the referenced function definition type in this case must be `rpc`.
...
+      "id": "${ .petId }"
+    },
+    "selectionSet": "{ id, name, age(useDogYears: ${ .isPetADog }) { dateOfBirth, years } }"
+  }
+}
+```
+
+Expressions must be evaluated before executing the operation.
+
+Also note that at this time subscriptions are not supported due to the workflow model.
```
tsurdilo, line 1234: Maybe use something like

Note that GraphQL Subscriptions are not supported at this time.

tsurdilo (changes requested):
This is really nice - thank you!
Left some minor comments. 

One thing that is missing from this PR 
are updates to the workflow json schema that corresponds to these changes.

I can add that in if you want, no problem.

tsurdilo (commented):
From what I see this is missing addition of "selectionSet" to the functionRef definition in workflow.js schema.

tsurdilo:
Regarding "selectionSet" i wonder if we can add that this property should only be used when the function type is graphql?

specification.md
lynxnathan: Yes, as you can see in the [spec](https://spec.graphql.org/June2018/#sec-Language.Operations) the `SelectionSet` is not optional, and neither is an empty `SelectionSet` valid.

specification.md
lynxnathan: Oh I meant result indeed!

lynxnathan:
Perhaps you mean something like this? https://json-schema.org/understanding-json-schema/reference/conditionals.html

tsurdilo:
nvm :) i missed that selectionSet is a property of arguments and not its own . sry

specification.md
```diff
@@ -2446,7 +2608,7 @@ it with its `object` type which has the following properties:
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
 | refName | Name of the referenced [function](#Function-Definition) | string | yes |
-| arguments | Arguments to be passed to the referenced function | object | no |
+| arguments | Arguments to be passed to the referenced function | object | only for GraphQL functions |
```
tsurdilo, line 2611: could you please change
"only for GraphQL functions"
to 
"yes if function type is graphql, otherwise no"

specification.md
tsurdilo: added selectionSet to be a property of the functionDef

tsurdilo:
Made some small changes, hope you don't mind, if its wrong I will revert them asap:

1) made "selectionSet" a property of the functionRef definition (it is defines as only usable if funtion type is "graphql"
2) removed the need to do arguments->inputs, so "arguments" would be the inputs 

Updated examples you provided , let me know what you think.

lynxnathan:
@tsurdilo looks cleaner, no objections from me.

tsurdilo (approved):

manuelstein (approved):
lgtm

tsurdilo:
approved in team meet 5/10 . needs rebase

PR #382, opened 2021-05-16
Adding support for secrets and globals

tsurdilo:

- [ x] Specification
- [x ] Schema
- [ ] Examples
- [ ] Extensions
- [x ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

Adds support for secrets and globals in workflow expressions


tsurdilo:
Runtime implementors, look at the jq "arguments" that you can pass to expressions..should be easy to impl it w/ that..just fyi

specification.md
```diff
@@ -1642,6 +1646,30 @@ In this case the `failOnValidationErrors` default value of `true` is assumed.
...
+which contains an array of names of the secrets, for example:
+
+```json
+"secrets": "file://workflowsecrets.json"
+```
+
+If `array` type, it defines an array (of string types) which contains the names of the secretes, for example:
+
+```json
+"secrets": ["MY_PASSWORD", "MY_STORAGE_KEY", "MY_ACCOUNT"]
```
cdavernas, line 1663: Secrets

cdavernas (approved):
@tsurdilo aside from some 'secretes' tips, this looks awesome! WP! X)

tsurdilo:
@cdavernas fixed. thanks for the review

schema/globals.json
```diff
@@ -0,0 +1,77 @@
...
+    ]
+  },
+  "required": [
+    "globals"
+  ],
+  "definitions": {
+    "global": {
+      "type": "object",
+      "properties": {
+        "nameSpace": {
```
ricardozanini, line 32:
```suggestion
        "namespace": {
```

schema/globals.json
```diff
@@ -0,0 +1,77 @@
...
+            {
+              "type": "object",
+              "description": "Global object value"
+            }
+          ]
+        }
+      },
+      "additionalProperties": false,
+      "required": [
+        "nameSpace",
```
ricardozanini, line 71:
```suggestion
        "namespace",
```

specification.md
```diff
@@ -1642,6 +1646,50 @@ In this case the `failOnValidationErrors` default value of `true` is assumed.
...
+```json
+"globals": "file://workflowglobals.json"
+```
+
+If `array` type, it defines an array which contains the global definitions, for example:
+
+```json
+"globals": [
+ {
+   "nameSpace": "AGE",
```
ricardozanini, line 1684:
```suggestion
   "namespace": "AGE",
```

specification.md
```diff
@@ -5026,7 +5072,139 @@ for your workflow definitions especially in production environments.
...
+
+Workflow globals are used to define static data which is available to [Workflow Expressions](#Workflow-Expressions).
+
+Globals can be defined via the [Workflow top-level "globals" property](#Workflow-Definition-Structure),
+for example:
+
+```json
+"globals": [
+{
+    "nameSpace": "MYNAMESPACE",
```
ricardozanini, line 5086:
```suggestion
    "namespace": "MYNAMESPACE",
```

specification.md
```diff
@@ -5026,7 +5072,139 @@ for your workflow definitions especially in production environments.
...
+    "name": "MYNAME",
+    "value": "myvalue"
+},
+{
+  "nameSpace": "OTHERNAMESPACE",
```
ricardozanini, line 5091:
```suggestion
  "namespace": "OTHERNAMESPACE",
```

specification.md
```diff
@@ -5026,7 +5072,139 @@ for your workflow definitions especially in production environments.
...
+Runtimes must make globals available to expressions.
+
+Here is an example of using globals in Workflow expressions:
+
+```json
+{
+...,
+"globals": [
+  {
+    "nameSpace": "AGE",
```
ricardozanini, line 5118:
```suggestion
    "namespace": "AGE",
```

ricardozanini (commented):
Hi @tsurdilo @cdavernas 

Can we consider using a `Configuration Map` instead of `Globals`? I think it can potentially correlate with Kubernetes `ConfigMap`, and implementations based on Kubernetes will have this object "free".

So, instead of having:

```json
{
    globals: [{
      namespace: AGE
      name: MAX
      value: 18
    }]
}
```

We would have something like:

```json
{
    configMap: [{
        age.max: 18
    }]
}
```

A plain key/value map and compatible with [Kubernetes ConfigMap](https://kubernetes.io/docs/concepts/configuration/configmap/).

tsurdilo:
@ricardozanini would that be too kubernetes-specific? it should work also w/o as in just setting some static values no matter the runtime deployment. 

i am not expert with configmap but looks to me you could map the namespace to the name of the configmap and 
then map the keys and values to the data section in the configmap.

you could map object types to string if kubernetes configmap only allows string/number data ?

tsurdilo:
@ricardozanini i guess we could add "fromConfigMap" option maybe to tell runtimes to read in configmap and convert to this json structure but i really think having a non-specific option like this could be used by more than just kubernetes runtimes? wdyt?

ricardozanini:
>  @ricardozanini would that be too kubernetes-specific? it should work also w/o as in just setting some static values no matter the runtime deployment.

I don't think so. I'm just suggesting that instead of a complex object, we just use a key/value pair for this globals/configuration structure in the DSL. The k8s ConfigMap has also the notion of holding "files" as part of their definition. We would just use a plain key/value definition.

> i am not expert with configmap but looks to me you could map the namespace to the name of the configmap and then map the keys and values to the data section in the configmap.

We could, but I don't see why we would do that. Just a plain `configuration` object map should do. 

> you could map object types to string if kubernetes configmap only allows string/number data?

No, you could not. Only any data rather than complex objects.

> I  guess we could add "fromConfigMap" option maybe to tell runtimes to read in configmap and convert to this json structure but i really think having a non-specific option like this could be used by more than just kubernetes runtimes? wdyt?

I think we should keep this super simple. Only a set of key/value pairs should do. It would be pretty similar to what we already have in many languages and also on k8s.

tsurdilo:
@ricardozanini 

>I don't think so. I'm just suggesting that instead of a complex object, we just use a key/value pair for this globals/configuration structure in the DSL.

why??? thats too limiting imo. these are passed as arguments to jq expressions which evaluate against json data...so i don't see a reason for this limitation honestly

tsurdilo:
@ricardozanini 

>I think we should keep this super simple. Only a set of key/value pairs should do. It would be pretty similar to what we already have in many languages and also on k8s.

>EDIT: for the configuration: "file:myurl.json" could work. For k8s implementations, one runtime could use configuration: kuberbetes://myConfigMap. All the keys in there would be read in runtime.

I don't think this will work as any json/yaml you want to read in would have to validate against the added "globals.json" JSON Schema. If you want to do this runtime would have to convert the structure of the k8s configmap first which then you can use just the key-value pairs and not use the object types. I don't see issue w/ that.

ricardozanini:
> @ricardozanini
> 
> > I don't think so. I'm just suggesting that instead of a complex object, we just use a key/value pair for this globals/configuration structure in the DSL.
> 
> why??? thats too limiting imo. these are passed as arguments to jq expressions which evaluate against json data...so i don't see a reason for this limitation honestly

hmm so you say that we would have a complex structure to a configuration parameter? Like static JSON data? I don't think this is a good practice. But either way, since JSON is a string and this is just configuration data AFAIK, you can add your JSON there. Won't be validated either way by the schema definition.

tsurdilo:
@ricardozanini 

whats the problem with doing 

```json
"globals": [  {    
"nameSpace": "TRANSLATIONS",
"name": "DOG",
"value": {
    "serbian": "pas",
   "spanish": "perro",
   "french": "chien"
}
]
```

and then use it in your expressions with like "${ $TRANSLATIONS.DOG.spanish }" ?

To me this is actually really useful.

specification.md
tsurdilo: +100 ty

specification.md
```diff
@@ -1642,6 +1646,50 @@ In this case the `failOnValidationErrors` default value of `true` is assumed.
...
+```json
+"globals": "file://workflowglobals.json"
+```
+
+If `array` type, it defines an array which contains the global definitions, for example:
+
+```json
+"globals": [
+ {
+   "namespace": "AGE",
```
jorgenj, line 1684: Is there a functional purpose for the namespace other than organizing similar globals into the same namespace?

This seems like overkill to have "namespaces", do we really think that a single workflow definition will have multiple namespaces?   If it's just for organizination, then users that want to group values could just prefix the names of their globals (ie; `AGE.MIN_ADULT`) instead?

tsurdilo: for large global data it is a nice way of grouping things. also for editors that can provide auto-completion
its much easier to do like 

$TRANSLATIONS.  --> and get a list of globals that fall under that namespace than doing

$ --> and get a list of a million strings to scroll a mile through

tsurdilo: we could make namespace optional (blank by default) which then would be the top-level namespace? that would be an option but im not for removing it

manuelstein: we're accessing all state data by their path. I think introducing namespacing in json/yaml like it's available in XML is interesting, but I think this shouldn't be a globals-specific thing. Objects are fine for grouping data, no?

manuelstein:
I think the globals in this case should be:
```json
{
  "TRANSLATIONS": {
    "DOG": {
      "serbian": "pas",
      "spanish": "perro",
      "french": "chien"
    }
  }
}
```

ricardozanini:
Fine by me if you guys prefer to have to deal with complex objects. To me, a simple key/pair would do. Adding tons of globals can be considered a bad practice, and having a simple `translations.dog.*` key should do.

specification.md
cdavernas: @manuelstein Yeah, agreed, and it might even be easier to integrate with jq, using --argjson flag

tsurdilo:
@manuelstein 

>I think the globals in this case should be:
{
  "TRANSLATIONS": {
    "DOG": {
      "serbian": "pas",
      "spanish": "perro",
      "french": "chien"
    }
  }
}


agree but problem with this how do we describe this structure in json schema. any inputs appreciated

jorgenj:
> agree but problem with this how do we describe this structure in json schema. any inputs appreciated
> or do we not bother with describing it and say any valid JSON object will do. up to you guys

It seems like "any valid json object" is reasonable.  Similar to how the the spec currently handle function arguments, right?

tsurdilo:
ok sounds like just json object is good. will update the pr when i can

specification.md
tsurdilo: ok will update for it to just be json object

tsurdilo:
@jorgenj @manuelstein @cdavernas 
One thing I noticed is that making "globals" just an object can clash with the secrets where we define the $SECRETES namespace as reserved. 
Are you OK with us just documenting this in the spec, because there is no way I think we can restrict that in the schema.

tsurdilo:
Updated to make globals object type.
In order not to interfere with $SECRETS, I made globals required to use the $GLOBALS namespace. 

Let me know if this is ok. I think this is safer than just telling users not to have an element in globals call "SECRETS" as they might not have control over that especially if they reference a file where globals are defined.

jorgenj:
> Updated to make globals object type.
> In order not to interfere with $SECRETS, I made globals required to use the $GLOBALS namespace.
> 
> Let me know if this is ok. I think this is safer than just telling users not to have an element in globals call "SECRETS" as they might not have control over that especially if they reference a file where globals are defined.

Agreed, this seems reasonable.

specification.md
```diff
@@ -1642,6 +1647,48 @@ In this case the `failOnValidationErrors` default value of `true` is assumed.
...
+
+It has two possible types, `string` or `object`.
+If `string` type, it is an URI pointing to a JSON or YAML document
+which contains an object of global definitions, for example:
+
+```json
+"globals": "file://workflowglobals.json"
+```
+
+If `object` type, it defines an array which contains the global definitions, for example:
```
jorgenj, line 1680: I'm blanking on what this should say instead right now, but I think with the latest updates this needs to be updated to not say "array" here, right?

specification.md
```diff
@@ -1642,6 +1647,48 @@ In this case the `failOnValidationErrors` default value of `true` is assumed.
...
+```
+
+For more information about Workflow secrets, reference the [Workflow Secrets section](#Workflow-Secrets).
+
+The `globals` property can be used to define Workflow global values
+which are accessible in [Workflow Expressions](#Workflow-Expressions).
+
+It has two possible types, `string` or `object`.
+If `string` type, it is an URI pointing to a JSON or YAML document
+which contains an object of global definitions, for example:
```
jorgenj, line 1674: Not saying I see a purpose to explicitly support this, but technically could the file contain just any valid json/yaml?  I think object is the most desirable use-case, but if the file contains an array, it seems like that'd be valid to just reference that data via `$GLOBALS[0]`, etc?

Should that be supported and described?  Or should we say that a runtime should consider that an error if the top-level content of the file is anything but an 'object'?

specification.md
tsurdilo: it was array before. now it was changed to object type with the latest updates

specification.md
tsurdilo: good point. up to you guys really. since globals can also be defined in-line we should probably restrict its type however. 
we can keep it as object type (which means the file should be a json object) or we can expand it to be object or array types (which then would restrict the read in files to that as well). 

either way, if the globals are read from file, they can only contain json which is also possible to define inline in the workflow definition

specification.md
jorgenj: Hmmm, sorry my question must not have been very clear.  

I was asking if you still want this to say "it defines an *array* which contains the global definitions"

tsurdilo: oops sry good point! will fix

cdavernas:
Anyone as an idea on how to actually map, say K8s config maps/secrets, to the 'secrets' or 'global' object?
An idea is to force the runtime to retrieve all the secrets it has access to (mounted on its container at given paths), and to merge them all into one fat object. Would that be the recommended way to do it?

ricardozanini:
@cdavernas in this case I would use this approach:

```json
{
    "globals": "kubernetes:/configMap/namespace/name" 
}
```
The string is a valid URI reference, `kubernetes` protocol (I just invented). The runtime would auto-discover and bind the file to it.

specification.md
tsurdilo: fixed

specification.md
```diff
@@ -5026,7 +5071,124 @@ for your workflow definitions especially in production environments.
...
+    "operation": ".applicant | .age < $GLOBALS.AGE.MIN_ADULT",
+    "type": "expression"
+  }
+]
+}
+```
+
+Workflow globals values should only contain static data, meaning that their value should not 
+contain Workflow expressions. 
+Workflow globals should not have access to [Workflow secrets definitions](#Workflow-Secrets).
```
manuelstein, line 5152: IIUC, "their value should not contain Workflow expressions" means I can't use expressions in defining their content. How could globals have access to secrets? Is the second sentence needed, i.e. can I somehow refer to secrets without an expression?

specification.md
```diff
@@ -5026,7 +5071,124 @@ for your workflow definitions especially in production environments.
...
+
```
manuelstein, line 5153: Are globals mutable? Can I change them in an expression and if I do, are the new contents also transferred to the next state or synchronized across parallel states when I modify them?

specification.md
```diff
@@ -5026,7 +5071,124 @@ for your workflow definitions especially in production environments.
...
+      "account": "${ $SECRETS.AZURE_STORAGE_ACCOUNT }",
+      "account-key": "${ $SECRETS.AZURE_STORAGE_KEY }",
+      ...
+    }
+  
+}
+```
+
+Note that secrets can also be used in [expression functions](#Using-Functions-for-Expression-Evaluation).
+
```
manuelstein, line 5191: Are secrets mutable, i.e. can I use an action to obtain a secret (token) and use it subsequently?
How are secrets different from globals?

manuelstein (commented):
So, this is the first time we're adding additional _documents_ to the workflow data which can be addressed by a dot `.` (in jq). I understand globals and secrets are provided in variables to the expression language. Is there any difference to use them compared to the workflow data? I think there may be (mutability) and I missed it, could you point it out pls?

specification.md
ricardozanini: This is a good question. I don't think globals should be mutable. Hence maybe we can change to `constants` instead?

specification.md
ricardozanini: I don't think secrets should be mutable.

specification.md
tsurdilo: good point :) yeah its obvious but do you think we should remove the second part or leave it there just in case someone does not get the obvious

ricardozanini (approved):

tsurdilo:
1) added immutable aspect to secretes
2) changed "globals" to "constants" as per suggestion
3) added immutable aspect to constants

ricardozanini (approved):

PR #385, opened 2021-05-18
Fix required params def for usedForCompensation

tsurdilo:

- [ ] Specification
- [x ] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

Fix for the required param definition of usedForCompensation

### DIFF
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -561,9 +561,11 @@
       },
       "additionalProperties": false,
       "if": {
-        "properties": {
-          "usedForCompensation": {
-            "const": true
+        "not": {
+          "properties": {
+            "usedForCompensation": {
+              "enum": [ false ]
+            }
           }
         }
       },
@@ -753,9 +755,11 @@
       },
       "additionalProperties": false,
       "if": {
-        "properties": {
-          "usedForCompensation": {
-            "const": true
+        "not": {
+          "properties": {
+            "usedForCompensation": {
+              "enum": [ false ]
+            }
           }
         }
       },
@@ -877,9 +881,11 @@
       },
       "additionalProperties": false,
       "if": {
-        "properties": {
-          "usedForCompensation": {
-            "const": true
+        "not": {
+          "properties": {
+            "usedForCompensation": {
+              "enum": [ false ]
+            }
           }
         }
       },
@@ -1262,9 +1268,11 @@
       },
       "additionalProperties": false,
       "if": {
-        "properties": {
-          "usedForCompensation": {
-            "const": true
+        "not": {
+          "properties": {
+            "usedForCompensation": {
+              "enum": [ false ]
+            }
           }
         }
       },
@@ -1381,9 +1389,11 @@
       },
       "additionalProperties": false,
       "if": {
-        "properties": {
-          "usedForCompensation": {
-            "const": true
+        "not": {
+          "properties": {
+            "usedForCompensation": {
+              "enum": [ false ]
+            }
           }
         }
       },
@@ -1492,9 +1502,11 @@
       },
       "additionalProperties": false,
       "if": {
-        "properties": {
-          "usedForCompensation": {
-            "const": true
+        "not": {
+          "properties": {
+            "usedForCompensation": {
+              "enum": [ false ]
+            }
           }
         }
       },
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -561,13 +561,12 @@
       },
       "additionalProperties": false,
       "if": {
-        "not": {
-          "properties": {
-            "usedForCompensation": {
-              "enum": [ false ]
-            }
+        "properties": {
+          "usedForCompensation": {
+            "const": true
           }
-        }
+        },
+        "required": ["usedForCompensation"]
       },
       "then": {
         "required": [
@@ -755,13 +754,12 @@
       },
       "additionalProperties": false,
       "if": {
-        "not": {
-          "properties": {
-            "usedForCompensation": {
-              "enum": [ false ]
-            }
+        "properties": {
+          "usedForCompensation": {
+            "const": true
           }
-        }
+        },
+        "required": ["usedForCompensation"]
       },
       "then": {
         "required": [
@@ -881,13 +879,12 @@
       },
       "additionalProperties": false,
       "if": {
-        "not": {
-          "properties": {
-            "usedForCompensation": {
-              "enum": [ false ]
-            }
+        "properties": {
+          "usedForCompensation": {
+            "const": true
           }
-        }
+        },
+        "required": ["usedForCompensation"]
       },
       "then": {
         "required": [
@@ -1268,13 +1265,12 @@
       },
       "additionalProperties": false,
       "if": {
-        "not": {
-          "properties": {
-            "usedForCompensation": {
-              "enum": [ false ]
-            }
+        "properties": {
+          "usedForCompensation": {
+            "const": true
           }
-        }
+        },
+        "required": ["usedForCompensation"]
       },
       "then": {
         "required": [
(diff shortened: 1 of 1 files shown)


JBBianchi:
Why use `"enum": [ false ]` instead of `"const": false` like it was with the "positive" version?

ricardozanini (approved):

tsurdilo:
@JBBianchi my impression was that this was causing the issue in sdk. If not we can just close this. Either approach work it seems.

JBBianchi:
I'm not sure you're fixing the issue. The goal is to avoid "falling into" `then` if `usedForCompensation` is not set.

Here, you did just changed the expression of the `if`, meaning that you said `if (!false) ...` instead of `if(true) ...`, which doesn't change a lot in the end. If `usedForCompensation` is not set, if will be ignored and it will fall into `then` anyways.

Either you have to change the if/then/else to invert the logic (`if (false) ...`) which means interverting then and else as well, or you can add a constrain to if with a require on `usedForCompensation` .

tsurdilo:
@JBBianchi i am testing this inside VSCode but they might have a different parsing approach for this.
We cannot make "usedForCompensation" a required field unfortunately.

JBBianchi:
> 
> 
> @JBBianchi i am testing this inside VSCode but they might have a different parsing approach for this.
> We cannot make "usedForCompensation" a required field unfortunately.

In the if block:
```json
      "if": {
        "properties": {
          "usedForCompensation": {
            "const": true
          }
        },
        "required": ["usedForCompensation"]
      }
```

JBBianchi:
> 
> 
> @JBBianchi i am testing this inside VSCode but they might have a different parsing approach for this.
> We cannot make "usedForCompensation" a required field unfortunately.
> OK I'll revert the logic then as suggested.

Btw, from my understanding, it is "kind of" required as it has a default value of "false"... isn't it?

tsurdilo:
@JBBianchi updated let me know if this works now

JBBianchi:
> 
> 
> @JBBianchi updated let me know if this works now

I tried with the sample provided in the [related SDK issue](https://github.com/serverlessworkflow/sdk-typescript/issues/99).
The workflow is not valid until we add an `end` property for instance. I think that's what we wanted?

```typescript
import { DefinedError } from 'ajv';
import { WorkflowValidator, Specification } from '../src';

describe('missing usedForCompensation', () => {
  it('should be invalid', () => {
    const workflow: Specification.Workflow = {
      id: 'helloworld',
      version: '1.0',
      name: 'Hello World Workflow',
      description: 'Inject Hello World',
      start: 'Hello State',
      states: [
        {
          type: 'inject',
          name: 'Hello State',
          data: {
            result: "Hello World!"
          }
        } as Specification.Injectstate
      ]
    };
    const workflowValidator: WorkflowValidator = new WorkflowValidator(workflow);
    const isValid = workflowValidator.validate();
    if (!isValid) {
      workflowValidator.errors.forEach((error: DefinedError) => console.error(error.message));
    }
    expect(isValid).toBe(false);
  });
});
```

outputs:
```
must have required property 'timeDelay'
must have required property 'timeDelay'
must match exactly one schema in oneOf
must have required property 'onEvents'
must have required property 'onEvents'
must have required property 'onEvents'
must match exactly one schema in oneOf
must have required property 'actions'
must have required property 'actions'
must have required property 'actions'
must match exactly one schema in oneOf
must have required property 'branches'
must have required property 'branches'
must have required property 'branches'
must match exactly one schema in oneOf
must have required property 'dataConditions'
must have required property 'eventConditions'
must match exactly one schema in oneOf
must have required property 'workflowId'
must have required property 'workflowId'
must match exactly one schema in oneOf
must have required property 'end'                      <----
must have required property 'transition'
must match exactly one schema in oneOf
must have required property 'inputCollection'
must have required property 'inputCollection'
must have required property 'inputCollection'
must have required property 'inputCollection'
must match exactly one schema in oneOf
must have required property 'action'
must have required property 'action'
must match exactly one schema in oneOf
must match a schema in anyOf
```

and is compliant to the test (workflow is invalid).

By adding `end: true` to the Injectstate, the workflow becomes valid.

tsurdilo:
@JBBianchi yeah this seems correct now - if usedForCompensation is not defined you need either "end" or "transition" defined.

One question regarding the above validation output - it seems it pulls in validation of all the states, not just inject state which is defined.

JBBianchi:
> 
> 
> @JBBianchi yeah this seems correct now - if usedForCompensation is not defined you need either "end" or "transition" defined.
> 
> One question regarding the above validation output - it seems it pulls in validation of all the states, not just inject state which is defined.

Yes, it validates the `states` property against all the union types even if the `type` property of each one of them is a const and it should be able to filter it down to `injectstate`. That might be something to improve on [ajv ](https://ajv.js.org/) side, not sure...

tsurdilo:
@JBBianchi ok thats fair, we can take a look at that at some point. 
@JBBianchi @antmendoza I'll go ahead and merge this then as its a fix - do you guys need this in 0.6.x as well ?

JBBianchi:
> 
> 
> @JBBianchi ok thats fair, we can take a look at that at some point.
> @JBBianchi @antmendoza I'll go ahead and merge this then as its a fix - do you guys need this in 0.6.x as well ?

Preferably <3

antmendoza:
late to the party, thank you @tsurdilo @JBBianchi ;)

antmendoza:
@tsurdilo are you going to release a new version of the 0.6 as well ?

tsurdilo:
@antmendoza thats a good question :) idk the answer now honestly, let's discuss. 
From your perspective, is it needed for the typescript sdk?

antmendoza:
If it is an inconvenience we can life without it

the actual implementation is fetching the json schemas from https://api.github.com/repos/serverlessworkflow/specification/contents/schema?ref=v0.6 but we can just copy them

tsurdilo:
@antmendoza is it possible that you pull the schema from https://serverlessworkflow.io/schemas/0.6/workflow.json ?
it would be much easier for me to push the 0.6.x updates we make there than cut a whole new release. let me know

JBBianchi:
> 
> 
> @antmendoza is it possible that you pull the schema from https://serverlessworkflow.io/schemas/0.6/workflow.json ?
> it would be much easier for me to push the 0.6.x updates we make there than cut a whole new release. let me know

Atm it's using the GitHub API to list and download everything from `schema` of the specified `ref` (pending PR changes the defaut ref from v{version} to {version}.x).

It's not a big deal to change and get the schemas from `https://serverlessworkflow.io/schemas/0.6/workflow.json` (only unreferenced schemas like extensions/kpi.json won't be fetched as there is no pointer to it for instance, once again, no big deal).

One thing thought, it would be nice to fix the schemas `$id` because right now it's ...serverlessworkflow.org/~~core~~/0.6/workflow.json

tsurdilo:
> One thing thought, it would be nice to fix the schemas `$id` because right now it's ...serverlessworkflow.org/~core~/0.6/workflow.json

@JBBianchi what should we change it to ?

JBBianchi:
I think the [$id](https://github.com/serverlessworkflow/specification/blob/0.6.x/schema/workflow.json#L2) should be `"$id": "https://serverlessworkflow.io/schemas/0.6/workflow.json"` so it's a valid URL.

tsurdilo:
> I think the [$id](https://github.com/serverlessworkflow/specification/blob/0.6.x/schema/workflow.json#L2) should be `"$id": "https://serverlessworkflow.io/schemas/0.6/workflow.json"` so it's a valid URL.

+1 will update over weekend

tsurdilo:
@JBBianchi 
https://github.com/serverlessworkflow/specification/pull/387 (master)
https://github.com/serverlessworkflow/specification/pull/388 (0.6.x)

I already updated this on the website by the way :)

manuelstein (approved):

tsurdilo:
@JBBianchi https://github.com/serverlessworkflow/serverlessworkflow.github.io/commit/aad78c58d1ccbe5299b636628f6f37dd530ce09d

PR #399, opened 2021-06-07
Update to workflow Timeouts

tsurdilo:

- [ x] Specification
- [x ] Schema
- [ x] Examples
- [ ] Extensions
- [x ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

* Updates workflow timeouts
* Formatting fixes

### DIFF
--- a/comparisons/comparison-bpmn.md
+++ b/comparisons/comparison-bpmn.md
@@ -1,25 +1,25 @@
 # Comparisons - BPMN2
 
 The [Business Process Model and Notation (BPMN)](https://www.omg.org/spec/BPMN/2.0/PDF) defines a flowchart-based
-DSL for workflows. It is maintained by the [Object Management Group (OMG)](https://www.omg.org/). 
+DSL for workflows. It is maintained by the [Object Management Group (OMG)](https://www.omg.org/).
 The latest BPMN version is [2.0.2](https://www.omg.org/spec/BPMN/2.0.2/), published in 2014.
 
 BPMN2 defines a graphical notation to specify workflows. This notation can then be shared between tooling and organizations.
 The graphical notation is translated into XML, which then can be used for runtime execution.
 
 For this comparison, we will compare the Serverless Workflow language with the graphical representation of BPMN2,
-and not its underlying XML representation. The BPMN2 XML is very difficult to understand, quite large for even the smallest workflows, and often not portable between runtimes. 
+and not its underlying XML representation. The BPMN2 XML is very difficult to understand, quite large for even the smallest workflows, and often not portable between runtimes.
 It makes more sense to use its portable graphical notation for comparisons.
 
-Serverless Workflow is a declarative workflow language, represented with JSON or YAML. It currently does not define a graphical notation. However, it can be graphically represented using different flowcharting techniques such as 
-UML activity diagrams. The [Serverless Workflow Java SDK](https://github.com/serverlessworkflow/sdk-java#building-workflow-diagram) as well as its [VSCode Extension](https://github.com/serverlessworkflow/vscode-extension) provide means to generate SVG diagrams based on the workflow JSON/YAML. 
+Serverless Workflow is a declarative workflow language, represented with JSON or YAML. It currently does not define a graphical notation. However, it can be graphically represented using different flowcharting techniques such as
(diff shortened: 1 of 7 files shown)


tsurdilo:
Adds ability to set top level timeout settings:

```json
"timeouts": {
      "type": "object",
      "description": "Workflow default timeouts",
      "properties": {
        "workflowExecTimeout": {
          "$ref": "#/definitions/workflowExecTimeout"
        },
        "stateExecTimeout": {
          "$ref": "#/definitions/stateExecTimeout"
        },
        "actionExecTimeout": {
          "$ref": "#/definitions/actionExecTimeout"
        },
        "branchExecTimeout": {
          "$ref": "#/definitions/branchExecTimeout"
        },
        "eventTimeout": {
          "$ref": "#/definitions/eventTimeout"
        }
      },
      "additionalProperties": false,
      "required": []
    }
```

States can also set timeouts property which includes properties relevant to the state type.
For parallel state its branches can also set the timeouts property which can contain the branchExecTimeout only

The specification doc describes how timeout settings can be inherited from "parent" definition, meaning that for a branch for example if not specified branchExecTimeout would be inherited from the parallel state timeouts setting, if not specified there would inherit from the top-level workflow timeouts branchExecTimeout setting...

tsurdilo:
i understand the the formatting changes make this a little harder to review, but cannot undo that now...sry.

tsurdilo:
best way to review is to see the parsed markdown of the Workflow Timeouts section: https://github.com/tsurdilo/specification/blob/tout2/specification.md#Workflow-Timeouts

ricardozanini (approved):

cdavernas (commented):
I see that `stateExecTimeout`, `actionExecTimeout`, `branchExecTimeout` and `eventTimeout` have the exact same definition. As a matter of consequence, shouldn't we find a mechanism to use one single definition for those?
IMHO, having so much different components is sure accurate, but might heavily burden a workflow definition. Having to declare 4 different timeouts for my states, events, actions and branches, even if I wanna use the same config for all three of them, seems a bit heavy to me. WDYT?

manuelstein:
Could it work to make this a timeout section and then allow one `default` as well as specific overwriting?
E.g.
```
"timeouts": {
    "state": 30, # state exec timeout 30sec
    "default": 10, # 10sec default used for others, i.e. workflow, action, event
}
```
Should we have a branch exec timeout? The reason I ask is that thinking about cases, I wonder if there's a case where a default branch exec timeout isn't doing the same as the default state exec timeout?

manuelstein:
I also realized that a default may not be that handy, e.g. if you specify one default that works on state and action, then a series of consecutive actions will always be limited by the state timeout.

tsurdilo:
Yeah that is for the user to determine. The state timeout expiration does have effect and should have precedence over actions execution timeout.

tsurdilo:
@cdavernas I would like to get this in as just an overall improvement, then will work on integration with error handling...once we got that in let's take a look at how this looks on a large workflow and figure out improvements to make things shorter/easier to define. wdyt?

tsurdilo:
@cdavernas @manuelstein is this good to get in? I know we wanted to do improvements as far as this being too verbose, but I'd like to get this in for 0.7 and then see to define a number of workflow defs (small and large) and be able to then take a look at concrete defs and see what kind of improvement needs to be made.

This is blocking work on updating the workflow error handling for timeouts ;)

tsurdilo:
> Should we have a branch exec timeout? The reason I ask is that thinking about cases, I wonder if there's a case where a default branch exec timeout isn't doing the same as the default state exec timeout?

I think there is an use for it. 
For defaults yeah users should not set a branch timeout that is larger than the state timeout.
This can be also presented to users during validation as its in the workflow definition

cdavernas:
@tsurdilo Fine by me! We can get it in and polish it later. Can't approve or review though, so you'd have to take my word on it x)

tsurdilo:
@cdavernas @manuelstein ok sounds good. we can go over it in meet again to see if there is any inputs.
I do think that just this alone is a huge upgrade over whats there now and even if its not perfect in terms of verbosity it unlocks us to really have clear set timeouts and then improve error handling for it

PR #402, opened 2021-06-09
[WIP] AsyncApi workflow extension

tsurdilo:

- [ ] Specification
- [ ] Schema
- [ ] Examples
- [x ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

Work in progress!!

Ads an AsyncApi Workflow extension for SW
This is an actual extension for AsyncApi and should also be published here: https://github.com/asyncapi/extensions-catalog

### DIFF
--- a/schema/extensions/asyncapiworkflow.yml
+++ b/schema/extensions/asyncapiworkflow.yml
@@ -1,0 +1,48 @@
+id: workflow
+type: generic
+title: CNCF Serverless Workflow Extension
+description: Associate a Serverless Workflow and its event definitions
+version: 0.1.0
+author: Serverless Workflow Authors
+contact:
+  email: serverlessworkflow@gmail.com
+  url: https://serverlessworkflow.io
+  name: Serverless Workflow Project
+tags:
+  - workflow
+  - serverless
+definitions:
+- hooks:
+  - info
+  - channels.*.publish
+  - channels.*.subscribe
+  schema:
+    type: object
+    required: []
+    properties:
+      id:
+        type: string
+      version:
+        type: string
+      event:
+        type: string
+  examples:
+  - description: Example configurations for both info and publis/subscript
+    associateWorkflow:
+      info:
+        title: My workflow application
+        x-workflow:
+          id: myworkflowid
+          version: myworkflowversion
+    assignPublishEvent:
+      channels:
+        hello:
+          publish:
+            x-workflow:
+              event: 'My Consumed Event'
+    assignSubscribeEvent:
+      channels:
+        goodbye:
+          subscribe:
+            x-workflow:
+              event: 'My Produced Event'


tsurdilo:
@cdavernas @mswiderski et all:

this is work in progress just wanted to get some opinions before moving forward with it.

it defines a real AsyncApi extension which can be used to:
* Associate an AsyncApi definition with a workflow definition (via id and version)
* Allows to define publish workflow event, meaning that for event you can use the CE definition of the "consumed" event defined in the workflow def
* Allows to define a subscribe workflow event, meaning that for event you can use the CE definition of the "produced" event defined in the workflow def

So a sample AsyncApi def could be (this is really simple):

```yaml
asyncapi: 2.0.0
info:
  title: Hello world application
  version: 0.1.0
  x-workflow:
    id: myworkflowid
    version: '1.0'
channels:
  hello:
    publish:
      x-workflow:
        event: HelloEvent
  goodbye:
    subscribe:
      x-workflow:
        event: GoodByeEvent
```

The associated simple workflow can be:

```yaml
id: myworkflowid
version: '1.0'
specVersion: '0.7'
name: My Workflow
start: Greet
events:
- name: HelloEvent
  type: helloEventType
  source: "/messages/hello"
  kind: consumed
- name: GoodbyeEvent
  type: goodbyeEventType
  source: "/messages/goodbye"
  kind: produced
states:
- name: Greet
  type: event
  onEvents:
  - eventRefs:
    - HelloEvent
    actions: []
  end:
    produceEvents:
    - eventRef: GoodbyeEvent
```

From this you can generate an application which can trigger workflow execution (workflows that start via event(s))
by publishing events, and also you can define for apps to subscribe to the events that the workflow publishes.

WDYT?

mswiderski:
@tsurdilo looking into it first question that I have - is there really a need for an extension? I mean why would someone consuming AsyncAPI need to know there is workflow behind it?

min other words won’t the standard spec constructs of AsyncAPI be enough? I think most important to know what is being produced and what users can subscribe to regardless if the logic behind is based on workflow or not. 
One important aspect here is the messages - structure of them but in serverless workflow spec data are not defined at all so this might be quite challenging to achieve. Though this is not related to AsyncAPI only but to OpenAPI too which runtimes can decide to expose as a way to easily guide consumers on what to expect from given service.

tsurdilo:
@mswiderski i think we might be looking at this from 2 different points of view
the "consumer" of this would be the runtime impls and not end users. end users would define this alongside their workflows.

Via the workflow dsl you do not define how the runtimes expose it. you have to depend on the runtime rules (for example the workflow id becomes the endpoint name the runtime exposes it as" etc. In addition the workflow DSL does not include things like how user wants to protect the workflow definition once a runtime makes it into a "service" on their rules. You have to then use the runtimes way to define auth. 

This allows you to define your workflow definition and use AsyncApi to define exactly how you want the runtime to expose it without some proprietary rules. 

In addition this allows you to offload specific definitions of event sources from the dsl into asyncapi standard. It allows you to not have to define if you are using kafka or mqtt inside your workflow definition (source) which makes it even more portable imo.

At least that was the idea behind it ;)

tsurdilo:
Note to self - this actually makes perfect sense for workflow queries ;)

mswiderski:
I admit we are talking about different things indeed. I am bit confused as it seems AsyncAPI should become a extension to workflow definition which sounds weird to me at this point. Not sure why it should be defined as an extra asset to workflow definition. 

To me AsyncAPI is like OpenAPI define the contract between service provider and service consumer. So it should provide all the info that especially consumers should be capable of using that service. And service provider could decide that given workflow to be exposed via certain channels e.g. Kafka or MQTT but that should be given by the service provider and not really user building the workflow. 

Workflow should be self contained meaning it should have all the info that runtime needs to make it executable and AsyncAPI or OpenAPI are ways to describe how that workflow can be used by consumers. 

Yeah, it will be runtime specific as some might support Kafka others MQTT and another with expose it only over http. So this part I don’t think it makes sense to make it portable as I don’t think it’s even possible to have that level of interoperability. Frankly speaking not sure what would be the value….

tsurdilo:
> Workflow should be self contained meaning it should have all the info that runtime needs to make it executable and AsyncAPI or OpenAPI are ways to describe how that workflow can be used by consumers.

I think this is the problem that this tries to solve and ideas on what alternatives exist are welcome:
 
If you develop a service you can use OpenApi to define how consumers can access it via OpenApi and in workflow def you can use it when making function invocations. 

When you create your workflow definition, you have to rely as mentioned currently on the runtime rules on how it is exposed.
Using workflow "id" is in a way a standard but not always something user might want. Ok, then we add the "business key" but again user has to change the workflow definition in order to change this, so not optimal.

This is even worse for message-driven workflows (starting event state / start message events). To move from mqtt to kafka you have to change your workflow definition (or rely on the frameworks used by the runtime). 
That is low-level and having them self contained for this and then try to use as much as domain specific features as possible on others is maybe a little bit weird.

Last thing I wanted to mention is when you develop a "service" you develop the OpenApi def along with it. 
When you develop a workflow why is it a bad idea to create OpenApi / AsyncApi definitions where you can describe it with?

mswiderski:
> If you develop a service you can use OpenApi to define how consumers can access it via OpenApi and in workflow def you can use it when making function invocations.

this is my point that you have the OpenAPI definition that describes how the service is accessible rather than how it should be deployed. As if I got it right that’s what you suggesting here - use AsyncAPI as deployment descriptor that instructs runtime what should be done. 

In addition that way we have like a dual meaning of the AsyncAPI 

- deployment instructions 
- consumer contract

personally I always looked at both OpenAPI and AsyncAPI as the later - consumer contract. Never thought about using it as deployment instructions which might be an interesting thing to explore.

> When you create your workflow definition, you have to rely as mentioned currently on the runtime rules on how it is exposed.
Using workflow "id" is in a way a standard but not always something user might want. Ok, then we add the "business key" but again user has to change the workflow definition in order to change this, so not optimal.

but does this really matter? I mean it does not make much sense to try to enforce runtimes how they expose workflows. With that said what’s why we should rely on OpenAPI and AsyncAPI definitions to provide the contract on how to consume this workflow. I would not recommend to try to unify how runtimes use OpenAPI or AsyncAPI as this will be very much limiting for the runtimes. 

> This is even worse for message-driven workflows (starting event state / start message events). To move from mqtt to kafka for consumed events you have to change your workflow definition (or rely on the frameworks used by the runtime). If you want to change events that you produce (lets say again mqtt to kafka) you have to use the runtime specific ways to do that, and again most likely have to change your workflow definition.

while I understand the point here I don’t think it as simple as that. There will be needs to change how things are deployed as well and some runtimes will make it specific to given protocol/environment/libraries in use. I don’t think we can escape from that to be honest. Trying to model workflows completely agnostic from how it is going to be used does not sound like usual practice. We might be taking it a step to far as in the end workflows are built to solve given problem and it is usually very well known what it will use and where it will be deployed and by that will want to take advantage of these details as much as possible. 

> This in end means that you have to rely on versioning and specific features provided by runtimes for anything pretty much.
With this approach you could do this without having any down time on your workflows or having idk rolling restarts or having to move your workflows to different cluster etc etc

I don’t see how this is solved to be honest. And yes you will rely on runtimes for pretty much anything - that’s the role of the runtime :)

> Last thing I wanted to mention is when you develop a "service" you develop the OpenApi def along with it.
When you develop a workflow why is it a bad idea to create OpenApi / AsyncApi definitions where you can describe it with? In the end its just code anyways ;)

first of all I don’t think you develop both service and OpenAPI it’s more you either do top to bottom (spec first and generate code from it) or bottom up (implement service and generate the spec out of it). I see it exactly the same for AsyncAPI as well. 
the different with workflows is that the workflow is actually your service description so no need for any extra spec to define it again. That being said workflow brings all the details to solve a given business problem as would any custom written service do and then runtimes can provide contracts how to interact with the workflow at runtime and that can be OpenAPI or AsyncAPI based. That way this workflow can be used by other workflows as well since it provides standard based contract and it is less important if this contract will differ between runtimes. 

So before giving alternative I’d like to better understand what we want to address with it. Are we looking at providing sort of deployment instructions based on AsyncAPI for runtimes to use on deploying workflows?

tsurdilo:
> And yes you will rely on runtimes for pretty much anything - that’s the role of the runtime :)
Imo true only for runtimes that enforce how workflows are invoked only by looking at the workflow definition and based on their technologies/programming languages used.
There are also runtimes where you have full control on how you want to deploy your workflows and how to expose them. Now given that this is probably not possible to enforce in the end I guess maybe yeah that statement is true for majority of cases/runtimes now.

The main idea is to provide a way using AsyncApi spec for users not to have to use the workflow definition to define their produced and consumed events, their sources etc. It is to allow to externalize this out of the definitions so it can change w/o updates to the core deployed definition.

It's similar approach as to why we added OpenApi for service invocations and not added custom properties to DSL.

cdavernas:
@tsurdilo I tend to agree with @mswiderski here. I don't see the use for describing how the worklows events are produced or can be consumed from runtimes. IMO it is not the responsibility of SW to do that, and it cuts through multiple boundaries. As a matter of fact, if you want to listen to events produced by the worklow, you should either connect to the channels they have been published onto, or, in case of runtime specific lifetime events, should connect onto whatever mechanism the runtime offers.

> The main idea is to provide a way using AsyncApi spec for users not to have to use the workflow definition to define their produced and consumed events, their sources etc

Why would we impose to users the need to declare events in AsyncAPI, even though most may not even be coming from an AsyncAPI compatible endpoint? Why would we impose to runtimes the need to parse two different DSLs to make one work? Why would the user need to learn AsyncAPI to declare events that we can so far describe directly in SW? I just really can't see the added value here. Actually, I can even see potentially huge problems, because AsyncAPI is used to describe servers and channels, not just events. Why would you define a channel you have no control on, just to declare an event? 
For example, let's say I connect to operation `https://asyncapi.test.api/asyncapi.json#lighton`, described by the AsyncAPI at the specified location. That operation is on, say, a Kafka channel. Now, I've declared that event in my own AsyncAPI document too, attached to my SW. Which doc is the right one? Mine or the server's? The server, of course, making mine totally useless. Plus, mine declares servers, channels and many other concepts that, at best, I will be ignoring. At worst, I could get confused, and think that using my local AsyncAPI document is the way to go. In that case,things can work out if my document is kinduv in sync with the server's. If the server's changes, for example by migrating from Kafka to NATSS, my local AsyncAPI document becomes, in addition of useless, totally wrong.
Plus, how could you possibly describe the server and impose a channel config onto runtimes? Well, you really shouldn't, this is up to runtimes exclusively IMO (and might vary at anytime), the same goes with their potential lifetime events channels.
Finally, you were speaking of cloud events (which is IMO the best way to produce lifetime events)... they are not supported out of the box by AsyncAPI, which does offer a (very crappy) way to represent the schema with an [http channel](https://www.asyncapi.com/blog/asyncapi-cloud-events). That fact alone kills it for me.

tsurdilo:
agree :) lets then archive this pr

Issue #287, opened 2021-03-08
Replace "jq syntax" with something better-defined


gibson042:
#216 proposed interpreting Workflow Expressions as "jq syntax", and #249 implemented it. But jq syntax<a href="#user-content-note1">¹</a> and semantics<a href="#user-content-note2">²</a> have no formal specification, and vary across versions<a href="#user-content-note3">³</a>, across non-official implementations<a href="#user-content-note4">⁴</a> and even across the _same_ implementation on different hosts<a href="#user-content-note5">⁵</a>. `jq` is a great utility, but it should not be relied upon in an interop specification.

**What would you like to be added**:
The "standard" expressions should have fully-specified syntax and semantics, and therefore should not be "jq". [JMESPath](https://jmespath.org/) and [CEL](https://github.com/google/cel-spec) come to mind as solid replacement candidates.

**Why is this needed**:
A specification should not depend upon the happenstance of particular implementation choices, because that risks fracturing the ecosystem built upon it. Undefined/implementation-defined behavior should be minimized, and—where not possible to eliminate completely—should be _explicitly_ documented in order to warn practitioners about areas of potential incompatibility.

--------
<a id="note1">¹</a> Even something as simple as `. []` is ambiguous because https://stedolan.github.io/jq/manual/ does not define tokenization.

<a id="note2">²</a> https://stedolan.github.io/jq/manual/ does not define the behavior of [`length`](https://stedolan.github.io/jq/manual/#length) with numeric input (_although the version 1.6 `jq` binary returns the number itself, even if it is non-integer_), the syntax accepted by [`tonumber`](https://stedolan.github.io/jq/manual/#tonumber) (_although the version 1.6 `jq` binary accepts at least leading ASCII whitespace, leading +/- signs, leading zeroes, leading/trailing decimal points, and exponential notation, and rejects hexadecimal notation_), the complete behavior of arithmetic operators (_e.g., `1e999 + 1e999`, `0 / 0`, `infinite * 0`, `nan == nan`, etc._), or the behavior of `… as $__loc__` (_which the version 1.6 `jq` binary rejects_).

<a id="note3">³</a> For example, `test("\\p{Alphabetic}")` is not valid in the version 1.5 `jq` binary but is valid in version 1.6 because of an update to the embedded [Oniguruma](https://github.com/kkos/oniguruma) library.

[...]

PR #403, opened 2021-06-09
Consistently reference jq version 1.6

gibson042:
I'm still opposed to default expression language being `jq` for a [number of reasons](https://github.com/serverlessworkflow/specification/issues/287), but if it must be referenced then it should be referenced with a specific version.

- [x] Specification
- [ ] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

To avoid implying that post-1.6 `jq` expressions are portable.

n/a

n/a

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -1362,7 +1362,7 @@ evaluated. Reference the
 [State Data Filtering](#State-data-filters) section for more information about state data filters.
 
 By default, all workflow expressions should be defined using the [jq](https://stedolan.github.io/jq/) [version 1.6](https://github.com/stedolan/jq/releases/tag/jq-1.6) syntax.
-You can find more information on jq in its [manual](https://stedolan.github.io/jq/manual/).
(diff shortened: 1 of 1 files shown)


manuelstein (approved):
LGTM

ricardozanini (approved):

tsurdilo (changes requested):
Does this mean we do no support 1.5 or 1.4? 
We should do this on the global level which includes 
specific versions (with reasons why not to support older ones) for OpenAPI, GraphQL, gRPC, and soon AsyncAPI. 

So having this alone just for jq and not others is not something we want to do. We should do a global change for versioning inside a table or a reference document rather than the spec document.

tsurdilo:
Thank you for the PR. We will address this as a global update before our 1.0 release. Have a great day.

Issue #285, opened 2021-03-05
Expressions vs. string interpolation (<strike>paths</strike>)


jorgenj:
**What would you like to be added**:

There are really 3 concerns at hand in this issue:

1. In situations where a field expects an expression (ie; actionDataFilter.result) is it necessary for the expression text to be surrounded by `${}`. 
2. In situations where the expected type of result of an expression is unknown, a user should be able to clearly indicate whether they expect string interpolation or json data to be returned (for example; specifying arguments to a function invocation)
3. Some data-filter fields return data, and others return paths where other data should be merged (actionDataFilter.result vs. actionDataFilter.toStateData).  These different concepts should be clarified, and in particular the "path" type fields should have some examples talking about the complex ways in which it's expected to work. 

~~Fix all the places where an 'expression' is used to specify a location in a json structure where data should be inserted/merged.~~

**Why is this needed**:

Regarding string interpolation vs. workflow expressions, if both use-cases are basically a jq expression in a string surrounded by `${}`, then there are situations where it's unclear which is intended.   Imho, for fields where the purpose of the field is to accept an expression, the DSL should not require the `${}`, and that `${}` should be reserved to indicate string interpolation is expected.

~~Recent changes to actionDataFilter confusingly use 'expression' for 'toStateData' attribute, and there are presumably many other places in the spec that also need to be fixed.~~

~~Expressions (denoted by a string with `${...}`) typically extract or transform some values.  On the other hand, a property like actionDataFilter.toStateData is expecting the user to indicate a 'path' or location in the state data where the results should be inserted/merged.   It doesn't make sense to call this an 'expression', or to have the user style it like an expression.   The proposal is to fix the descriptions, and remove the surrounding `${}` from all of the examples.~~

tsurdilo:
i dont really get this at all, but others maybe do so please chime in. for me its much less confusing just to say "if a string is supposed to be evaluated as an expression by the runtime put it inside ${ ... }".

jorgenj:
See my comments here on 'interpolation':

https://github.com/serverlessworkflow/specification/issues/286#issuecomment-792939516

jorgenj:
Reposting this here for clarity:


Consider this example from bash:

```
var FOO=BAR
var $FOO=BAR

env
```

What do you expect the output of env to include?

Now consider the following example with what you're proposing:
```yaml
toStateData: ${ .foo }
```

Given the following action results:
```yaml
baz: 1
```

Given the existing state data of:
[...]

PR #404, opened 2021-06-12
[WIP] Expressions Markup

manuelstein:

Update expression markup:
- resolve string interpolation conflict by leaving string interpolation to the expression language as implemented in Synapse
- resolve function naming conflict by separating expression function references from expressions to facilitate implementation in Synapse
- clarify expression language support requirements for secrets and constants, so it can simplify implementation in Synapse
- clarify use of expression paths to be assignments (as already implemented in Synapse), neither deep or shallow merges
- reorder expression language examples (move expression function declaration closer to the example that uses the expression function) to make it easier to read
- separate JSONSchema expression typing from pure JSON data types to allow explicit declaration of parameters that support expressions as currently only the processing of few parameters in Synapse actually check for expression typing

The issues this PR addresses apply to any implementation. All solutions have been discussed extensively and chosen carefully to cause the least friction.

The presentation of this PR is to collect preliminary agreement. If successful, necessary search/replace work on examples would be added.

### DIFF
--- a/schema/common.json
+++ b/schema/common.json
@@ -4,6 +4,100 @@
   "description": "Serverless Workflow specification - common schema",
   "type": "object",
   "definitions": {
+    "expressionMarkup": {
+      "type": "string",
+      "pattern": "^\\$\\{.*\\}$"
+    },
+    "functionMarkup": {
+      "type": "string",
+      "pattern": "^\\$fn\\{.*\\}$"
+    },
+    "markup": {
+      "oneOf": [
+        {
+          "$ref": "#/definitions/expressionMarkup"
+        },
+        {
+          "$ref": "#/definitions/functionMarkup"
+        }
+      ]
+    },
+    "booleanOrMarkup": {
+      "oneOf": [
+        {
+          "$ref": "#/definitions/markup"
+        },
+        {
+          "type": "boolean"
+        }
+      ]
+    },
+    "numberOrMarkup": {
+      "oneOf": [
+        {
+          "$ref": "#/definitions/markup"
+        },
+        {
+          "type": "number"
+        }
+      ]
+    },
+    "stringOrMarkup": {
+      "oneOf": [
+        {
+          "$ref": "#/definitions/markup"
+        },
+        {
+          "type": "string",
+          "pattern": "^(?!\\$(fn)?\\{.*\\}$)"
+        }
+      ]
+    },
+    "arrayOrMarkup": {
+      "oneOf": [
+        {
+          "$ref": "#/definitions/markup"
+        },
+        {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/anyOrMarkup"
+          }          
+        }
+      ]
+    },
+    "objectOrMarkup": {
+      "oneOf": [
+        {
+          "$ref": "#/definitions/markup"
+        },
+        {
+          "type": "object",
+          "additionalProperties": {
+            "$ref": "#/definitions/anyOrMarkup"
+          }
+        }
+      ]
+    },
+    "anyOrMarkup": {
+      "oneOf": [
+        {
+          "$ref": "#/definitions/booleanOrMarkup"
+        },
+        {
+          "$ref": "#/definitions/numberOrMarkup"
+        },
+        {
+          "$ref": "#/definitions/stringOrMarkup"
+        },
+        {
+          "$ref": "#/definitions/arrayOrMarkup"
+        },
+        {
+          "$ref": "#/definitions/objectOrMarkup"
+        }
+      ]
+    },
     "metadata": {
       "type": "object",
       "description": "Metadata information",
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -391,12 +391,12 @@
               "description": "Name of the referenced function"
             },
             "arguments": {
-              "type": "object",
+              "$ref": "common.json#/definitions/objectOrMarkup",
               "description": "Function arguments/inputs"
             },
             "selectionSet": {
-              "type": "string",
-              "description": "Only used if function type is 'graphql'. A string containing a valid GraphQL selection set"
+              "$ref": "common.json#/definitions/stringOrMarkup",
+              "description": "Only used if function type is 'graphql'. A string or an expression returning a valid GraphQL selection set"
             }
           },
           "additionalProperties": false,
@@ -419,17 +419,14 @@
           "description": "Reference to the unique name of a 'consumed' event definition"
         },
         "data": {
-          "type": [
-            "string",
-            "object"
-          ],
-          "description": "If string type, an expression which selects parts of the states data output to become the data (payload) of the event referenced by 'triggerEventRef'. If object type, a custom object to become the data (payload) of the event referenced by 'triggerEventRef'."
+          "$ref": "common.json#/definitions/anyOrMarkup",
+          "description": "The data (payload) of the produced event referenced by `triggerEventRef`. Workflow expressions can be used to create values."
         },
         "contextAttributes": {
           "type": "object",
-          "description": "Add additional extension context attributes to the produced event",
+          "description": "Add additional event extension context attributes to the trigger/produced event. Workflow expressions can be used to create values.",
           "additionalProperties": {
-            "type": "string"
+            "$ref": "common.json#/definitions/objectOrMarkup"
           }
         }
       },
@@ -1243,8 +1240,8 @@
           "description": "Data condition name"
         },
         "condition": {
-          "type": "string",
-          "description": "Workflow expression evaluated against state data. Must evaluate to true or false"
+          "$ref": "common.json#/definitions/booleanOrMarkup",
+          "description": "Boolean or a workflow expression evaluated against state data. Must evaluate to true or false"
         },
         "transition": {
           "description": "Workflow transition if condition is evaluated to true",
@@ -1270,7 +1267,7 @@
         },
         "condition": {
           "type": "string",
-          "description": "Workflow expression evaluated against state data. Must evaluate to true or false"
+          "description": "Boolean or a workflow expression evaluated against state data. Must evaluate to true or false"
         },
         "end": {
           "$ref": "#/definitions/end",
@@ -1406,16 +1403,16 @@
           "description": "State end definition"
         },
         "inputCollection": {
-          "type": "string",
-          "description": "Workflow expression selecting an array element of the states data"
+          "$ref": "common.json#/definitions/arrayOrMarkup",
+          "description": "Array or a workflow expression returning an array"
         },
         "outputCollection": {
-          "type": "string",
-          "description": "Workflow expression specifying an array element of the states data to add the results of each iteration"
+          "$ref": "common.json#/definitions/jsonpointer",
+          "description": "Workflow expression merging the `$RESULT` of each iteration into states data"
         },
         "iterationParam": {
           "type": "string",
-          "description": "Name of the iteration parameter that can be referenced in actions/workflow. For each parallel iteration, this param should contain an unique element of the inputCollection array"
+          "description": "Name of the iteration parameter that can be referenced in actions/workflow. For each parallel iteration, this param should contain a unique element of the inputCollection array"
         },
         "max": {
           "type": [
@@ -1761,17 +1758,14 @@
           "description": "References a name of a defined event"
         },
         "data": {
-          "type": [
-            "string",
-            "object"
-          ],
-          "description": "If String, expression which selects parts of the states data output to become the data of the produced event. If object a custom object to become the data of produced event."
+          "$ref": "common.json#/definitions/anyOrMarkup",
+          "description": "The data (payload) of the produced event. Workflow expressions can be used."
         },
         "contextAttributes": {
           "type": "object",
-          "description": "Add additional event extension context attributes",
+          "description": "An object containing additional event extension context attributes. Workflow expressions can be used.",
           "additionalProperties": {
-            "type": "string"
+            "$ref": "common.json#/definitions/objectOrMarkup"
           }
         }
       },
@@ -1784,11 +1778,11 @@
       "type": "object",
       "properties": {
         "input": {
-          "type": "string",
+          "$ref": "common.json#/definitions/markup",
           "description": "Workflow expression to filter the state data input"
         },
         "output": {
-          "type": "string",
+          "$ref": "common.json#/definitions/markup",
           "description": "Workflow expression that filters the state data output"
         }
       },
@@ -1798,13 +1792,9 @@
     "eventdatafilter": {
       "type": "object",
       "properties": {
-        "data": {
-          "type": "string",
-          "description": "Workflow expression that filters the received event/payload (default: '${ . }')"
-        },
         "toStateData": {
-          "type": "string",
-          "description": " Workflow expression that selects a state data element to which the filtered event should be added/merged into. If not specified, denotes, the top-level state data element."
+          "$ref": "common.json#/definitions/markup",
+          "description": "Workflow expression to save the result of the event reception provided in expression variable `$RESULT`. If not specified, the event is merged at the top-level state data (default `${ . *= $RESULT }`)"
         }
       },
       "additionalProperties": false,
@@ -1814,16 +1804,12 @@
       "type": "object",
       "properties": {
         "fromStateData": {
-          "type": "string",
-          "description": "Workflow expression that selects state data that the state action can use"
-        },
-        "results": {
-          "type": "string",
-          "description": "Workflow expression that filters the actions data results"
+          "$ref": "common.json#/definitions/markup",
+          "description": "Workflow expression that selects state data that the action can use"
         },
         "toStateData": {
-          "type": "string",
-          "description": "Workflow expression that selects a state data element to which the action results should be added/merged into. If not specified, denote, the top-level state data element"
+          "$ref": "common.json#/definitions/jsonpointer",
+          "description": "Workflow expression to save the result of the operation provided in expression variable `$RESULT`. If not specified, the action result is merged at the top-level state data (default `${ . *= $RESULT }`)"
         }
       },
       "additionalProperties": false,
--- a/specification.md
+++ b/specification.md
@@ -288,8 +288,8 @@ The workflow data output is the data output of the last executed workflow state.
 
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
-| input | Workflow expression to filter the states data input | string | no |
-| output | Workflow expression that filters the states data output | string | no |
+| input | Workflow expression to filter the states data input | [expression](#workflow-expressions) | no |
+| output | Workflow expression that filters the states data output | [expression](#workflow-expressions) | no |
 
 <details><summary><strong>Click to view example definition</strong></summary>
 <p>
@@ -419,9 +419,8 @@ The second way would be to directly filter only the "veggie like" vegetables wit
 
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
-| fromStateData | Workflow expression that filters state data that can be used by the action | string | no |
-| results | Workflow expression that filters the actions data results | string | no |
-| toStateData | Workflow expression that selects a state data element to which the action results should be added/merged into. If not specified denotes the top-level state data element | string | no |
+| fromStateData | Workflow expression that selects state data that the action can use | [expression](#workflow-expressions) | no |
+| toStateData | Workflow expression to save the result of the operation provided in expression variable `$RESULT`. If not specified, the action result is merged at the top-level state data (default `${ . *= $RESULT }`) | [expression](#workflow-expressions) | no |
 
 <details><summary><strong>Click to view example definition</strong></summary>
 <p>
@@ -437,9 +436,8 @@ The second way would be to directly filter only the "veggie like" vegetables wit
 ```json
 {
   "actionDataFilter": {
-    "fromStateData": "${ .language }",
-    "results": "${ .results.greeting }",
-    "toStateData": "${ .finalgreeting }"
+    "fromStateData": "${ .language }", 
+    "toStateData": "${ .finalgreeting = $RESULT.greeting }"
   }
 }
 ```
@@ -450,8 +448,7 @@ The second way would be to directly filter only the "veggie like" vegetables wit
 ```yaml
 actionDataFilter:
   fromStateData: "${ .language }"
-  results: "${ .results.greeting }"
-  toStateData: "${ .finalgreeting }"
+  toStateData: "${ .finalgreeting = $RESULT.greeting }"
 ```
 
 </td>
@@ -460,16 +457,22 @@ actionDataFilter:
 
 </details>
 
-Action data filters can be used inside [Action definitions.](#Action-Definition)
-Each action can define this filter which can:
+Action data filters can be used inside [Action definitions.](
+#Action-Definition).
 
-* Filter the state data to select only the data that can be used within function definition arguments using its `fromStateData` property.
-* Filter the action results to select only the result data that should be added/merged back into the state data
-  using its `results` property.
-* Select the part of state data which the action data results should be added/merged to
-  using the `toStateData` property.
+The `fromStateData` property filters the state data to select only the data that
+can be used within function definition arguments.
+By default, the workflow expressions are evaluated against the entire state
+data.
 
-To give an example, let's say we have an action which returns a list of breads and pasta types.
+The `toStateData` property can customize how the action's result will be
+embedded into the state data.
+The expression can access the action result through the `$RESULT` variable.
+If not set to a custom expression, the action result is merged at the top-level
+of state data (equivalent to `${ . *= $RESULT }`).
+
+To give an example, let's say we have an action which returns a list of breads
+and pasta types.
 For our workflow, we are only interested into breads and not the pasta.
 
 Action results:
@@ -481,7 +484,7 @@ Action results:
 }
 ```
 
-We can use an action data filter to filter only the breads data:
+We can use an action data filter to save only the breads:
 
 ```json
 {
@@ -489,7 +492,7 @@ We can use an action data filter to filter only the breads data:
     {
        "functionRef": "breadAndPastaTypesFunction",
        "actionDataFilter": {
-          "results": "${ {breads: .breads} }"
+          "toStateData": "${ .breads = $RESULT.breads} }"
        }
     }
  ]
@@ -525,8 +528,7 @@ and have the following action definition:
     {
        "functionRef": "breadAndPastaTypesFunction",
        "actionDataFilter": {
-          "results": "${ [ .breads[0], .pasta[1] ] }",
-          "toStateData": "${ .itemsToBuyAtStore }"
+          "toStateData": "${ .itemsToBuyAtStore = [ $RESULT.breads[0], $RESULT.pasta[1] ] }"
        }
     }
  ]
@@ -550,9 +552,7 @@ into. With this, after our action executes the state data would be:
 
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
-| data | Workflow expression that filters the event data (payload) | string | no |
-| toStateData | Workflow expression that selects a state data element to which the action results should be added/merged into. If not specified denotes the top-level state data element | string | no |
-
+| toStateData | Workflow expression to save the result of the event reception provided in expression variable `$RESULT`. If not specified, the event is merged at the top-level state data (default `${ . *= $RESULT }`) | [expression](#workflow-expressions) | no |
 <details><summary><strong>Click to view example definition</strong></summary>
 <p>
 
@@ -567,7 +567,7 @@ into. With this, after our action executes the state data would be:
 ```json
 {
     "eventDataFilter": {
-       "data": "${ .data.results }"
+       "toStateData": "${ .receivedEvent = $RESULT }"
     }
 }
 ```
@@ -577,7 +577,7 @@ into. With this, after our action executes the state data would be:
 
 ```yaml
 eventDataFilter:
-  data: "${ .data.results }"
+  toStateData: "${ .receivedEvent = $RESULT }"
 ```
 
 </td>
@@ -586,16 +586,16 @@ eventDataFilter:
 
 </details>
 
-Event data filters can be used to filter consumed event payloads.
-They can be used to:
+Event data filters can be used to customize how the consumed event will be
+embedded into the state data. An expression in the `toStateData` property can
+address the event through the `$RESULT` variable.
 
-* Filter the event payload to select only the data that should be added/merged into the state data
-  using its `data` property.
-* Select the part of state data into which the event payload should be added/merged into
-  using the `toStateData` property.
+When `toStateData` is not set to a custom expression, the event is merged at the
+top-level state data (equivalent to `${ . *= $RESULT }`).
 
-Allows event data to be filtered and added to or merged with the state data. All events have to be in the CloudEvents format
-and event data filters can filter both context attributes and the event payload (data) using the `data` property.
+All events have to be in the CloudEvents format. The [Event definition](#Event-Definition)'s `dataOnly` property decides whether `$RESULT` carries
+only the `data` attribute of the event (default, `dataOnly = true`) or whether
+it contains the entire event envelope (`dataOnly = false`).
 
 Here is an example using an event filter:
 
@@ -603,9 +603,11 @@ Here is an example using an event filter:
 <img src="media/spec/event-data-filter-example1.png" height="400px" alt="Event Data Filter Example"/>
 </p>
 
-Note that the data input to the Event data filters depends on the `dataOnly` property of the associated [Event definition](#Event-Definition).
-If this property is not defined (has default value of `true`), Event data filter expressions are evaluated against the event payload (the CloudEvents `data` attribute only). If it is set to
-`false`, the expressions should be evaluated against the entire CloudEvent (including its context attributes).
+If the [Event definition](#Event-Definition)'s `dataOnly` property is not
+defined (has default value of `true`), Event data filter expression will be
+provided only the event payload (the CloudEvents `data` attribute only) through
+the `$RESULT` variable. If the property is set to `false`, the expression
+should be given the entire CloudEvent (including its context attributes).
 
 #### Using multiple data filters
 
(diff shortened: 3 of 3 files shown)


specification.md
```diff
@@ -5157,43 +5225,37 @@ Workflow constants should not have access to [Workflow secrets definitions](#Wor
...
+      "pass": "${ $SECRETS.MY_PASSWORD }",
+      "token": "${ $SECRETS.\"urn:example:tenant:accessToken\" }"
     }
-  
 }
 ```
 
 Note that secrets can also be used in [expression functions](#Using-Functions-for-Expression-Evaluation).
 
-Secrets are immutable, meaning that workflow expressions are not allowed to change their values.
```
tsurdilo, line 5196: I think this should maybe be a separate PR if you wish to remove it. 
We decided in our meeting that this should be added when we added secretes/constants. 

I would suggest open a separate pr and define reason for removal please.

tsurdilo:
I assume this is missing the updates to examples and comparisons docs expressions until a decision is made right ?
That's fine really just wondering if that will be updated as well.

There is also all the tests in SDKs where the DSL should be updated for this after.

schema/common.json
```diff
@@ -4,6 +4,103 @@
...
+    "expressionMarkup": {
+      "type": "string",
+      "pattern": "^\\${.*}$"
+    },
+    "jsonpointer": {
+      "$ref": "#/definitions/expressionMarkup"
+    },
+    "functionMarkup": {
+      "type": "string",
+      "pattern": "^\\$fn{.*}$"
```
tsurdilo, line 16: This is confusing to me honestly. We have an expression definition which is assumed its contents is an expression then we have a function markup which assumes its content is reference to expression function. 

the diff from what i can tell is 
"${ expresion}"
"$fn{ string }"

i think just doing the way we have now:
"${ expression }" 
"${ fn:string }"

is same thing really no reason to change and it makes more sense to me at least.

tsurdilo: i do like the use of pattern tho.we can add patterns to existing stuff to make it more clear when writing DSL in  IDEs

schema/common.json
```diff
@@ -4,6 +4,103 @@
...
+        },
+        {
+          "type": "object",
+          "additionalProperties": {
+            "$ref": "#/definitions/anyOrMarkup"
+          }
+        }
+      ]
+    },
+    "anyOrMarkup": {
```
tsurdilo, line 85: I think that this is here only for runtimes, not the spec
there is no reason imo to define this inside DSL where runtimes can determine the result type from the expression alone.

specification.md
```diff
@@ -1212,15 +1212,17 @@ Which would execute the mutation, creating the object and returning the followin
...
 
 ```json
 {
   "refName": "getOnePet",
   "arguments": {
-    "id": "${ .petId }"
+    "id": "${.petId}"
   },
-  "selectionSet": "{ id, name, age(useDogYears: ${ .isPetADog }) { dateOfBirth, years } }"
+  "selectionSet": "${ '{ id, name, age(useDogYears: \\(.isPetADog)) { dateOfBirth, years }' }"
```
tsurdilo, line 1225: why should be make this so hard for users to define (with \\ and ' etc)? its much easier to define the way its before

schema/workflow.json
```diff
@@ -442,17 +442,14 @@
           "description": "Reference to the unique name of a 'consumed' event definition"
         },
         "data": {
-          "type": [
-            "string",
-            "object"
-          ],
-          "description": "If string type, an expression which selects parts of the states data output to become the data (payload) of the event referenced by 'triggerEventRef'. If object type, a custom object to become the data (payload) of the event referenced by 'triggerEventRef'."
+          "$ref": "common.json#/definitions/objectOrMarkup",
```
tsurdilo, line 445: why does it have to be object or markup restriction? sorry i don't understand

schema/workflow.json
```diff
@@ -1188,8 +1185,8 @@
           "description": "Data condition name"
         },
         "condition": {
-          "type": "string",
-          "description": "Workflow expression evaluated against state data. Must evaluate to true or false"
+          "$ref": "common.json#/definitions/booleanOrMarkup",
+          "description": "Boolean or workflow expression that evaluates to boolean against state data"
```
tsurdilo, line 1189: why "boolean or workflow expression" you would never just put "false" or "true" as the value here. its always an expression which must return boolean value

schema/workflow.json
```diff
@@ -1339,16 +1336,16 @@
           "description": "State end definition"
         },
         "inputCollection": {
-          "type": "string",
-          "description": "Workflow expression selecting an array element of the states data"
+          "$ref": "common.json#/definitions/arrayOrMarkup",
```
tsurdilo, line 1339: why array or markup. the expression never returns an array. it just selects an array.

schema/common.json
```diff
@@ -4,6 +4,103 @@
   "description": "Serverless Workflow specification - common schema",
   "type": "object",
   "definitions": {
+    "expressionMarkup": {
+      "type": "string",
+      "pattern": "^\\${.*}$"
+    },
+    "jsonpointer": {
```
tsurdilo, line 11: https://datatracker.ietf.org/doc/html/rfc6901 ?
for this i think we need to decide do we want to use this in spec first ? :)

schema/workflow.json
```diff
@@ -414,12 +414,12 @@
...
             "arguments": {
-              "type": "object",
+              "$ref": "common.json#/definitions/objectOrMarkup",
               "description": "Function arguments/inputs"
             },
             "selectionSet": {
-              "type": "string",
-              "description": "Only used if function type is 'graphql'. A string containing a valid GraphQL selection set"
+              "$ref": "common.json#/definitions/stringOrMarkup",
+              "description": "Only used if function type is 'graphql'. A string or an expression returning a string formatted as a valid GraphQL selection set"
```
tsurdilo, line 422: " A string or an expression returning a string formatted as a valid GraphQL selection set"
i really have 0 clue what this means as a reader. does it have to be a string just cause its a string value property? 
any object return by expression can be expressed in its string format here imo. but could be wrong.

schema/workflow.json
```diff
@@ -442,17 +442,14 @@
...
-            "object"
-          ],
-          "description": "If string type, an expression which selects parts of the states data output to become the data (payload) of the event referenced by 'triggerEventRef'. If object type, a custom object to become the data (payload) of the event referenced by 'triggerEventRef'."
+          "$ref": "common.json#/definitions/objectOrMarkup",
+          "description": "Object structure that can use expressions to generate the data of the produced event."
         },
         "contextAttributes": {
           "type": "object",
-          "description": "Add additional extension context attributes to the produced event",
+          "description": "An object carrying additional extension context attributes to the produced event that may each use expresions",
```
tsurdilo, line 450: again, i really do not understand this sentence

specification.md
```diff
@@ -223,8 +223,8 @@ The workflow data output is the data output of the last executed workflow state.
 
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
-| input | Workflow expression to filter the states data input | string | no |
-| output | Workflow expression that filters the states data output | string | no |
+| input | Workflow expression to filter the states data input | expression | no |
```
tsurdilo, line 226: the type should be a valid JSON type. not a made up one please

tsurdilo: same for all other changes from "string" type to "expression" type

specification.md
```diff
@@ -301,7 +301,7 @@ we can define a state filter:
 ```json
 {
   "stateDataFilter": {
-    "input": "${ {fruits: .fruits} }"
+    "input": "${fruits: .fruits}"
```
tsurdilo, line 304: this is wrong imo. the original expression its is clear that the expression returns an object
in changex expression its not even an expression that can be evaluated by an expression parser
it defines really its own expression syntax imo 

if .fruits returns an array in some case then what you create an element fruits which is of array type? but thats not the user wanted he wants to just have the array. confusing imo.

specification.md
```diff
@@ -345,7 +345,7 @@ The second way would be to directly filter only the "veggie like" vegetables wit
 ```json
 {
   "stateDataFilter": {
-    "input": "${ {vegetables: .vegetables[] | select(.veggieLike == true)} }"
+    "input": "${vegetables: .vegetables[] | select(.veggieLike == true)}"
```
tsurdilo, line 348: the expression that was removed is a clear jq expression you can cut/paste into jq play for example
you are taking away users ability to use 1/2 of the jq spec like this imo

specification.md
```diff
@@ -424,7 +424,7 @@ We can use an action data filter to filter only the breads data:
     {  
        "functionRef": "breadAndPastaTypesFunction",
        "actionDataFilter": {
-          "results": "${ {breads: .breads} }"
+          "results": "${breads: .breads}"
```
tsurdilo, line 427: {breads: .breads}  is a jq expression
"breads: .breads" is not

specification.md
```diff
@@ -460,8 +460,8 @@ and have the following action definition:
     {  
        "functionRef": "breadAndPastaTypesFunction",
        "actionDataFilter": {
-          "results": "${ [ .breads[0], .pasta[1] ] }",
-          "toStateData": "${ .itemsToBuyAtStore }"
+          "results": "${[ .breads[0], .pasta[1] ]}",
```
tsurdilo, line 463: this keeps the expression the same (just removes the leading trailing spaces). to keep it consistent to the changes you made for queries that return object type (removal of { } from valid expressions) and then keeping the array ones the same is confusing imo

specification.md
```diff
@@ -485,8 +485,8 @@ into. With this, after our action executes the state data would be:
 
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
-| data | Workflow expression that filters the event data (payload) | string | no |
-| toStateData | Workflow expression that selects a state data element to which the action results should be added/merged into. If not specified denotes the top-level state data element | string | no |
+| data | Workflow expression that filters the event data (payload) | expression | no |
+| toStateData | Workflow expression that selects a state data element to which the action results should be added/merged into. If not specified denotes the top-level state data element | path expression | no |
```
tsurdilo, line 489: please define what is a path expression in its own section so its clear and link to it here

specification.md
```diff
@@ -1251,7 +1253,7 @@ Let's take a look at an example of such definitions:
     "type": "expression"
   },
   {
-    "name": "isMinor",
+    "name": " is a minor ?",
```
tsurdilo, line 1256: space between minor and ?

specification.md
```diff
@@ -1357,44 +1359,86 @@ For more information about workflow expressions, reference the [Workflow Express
...
-
-Serverless Workflow does not mandate the use of jq and it's possible to use an expression language 
-of your choice with the restriction that a single one must be used for all expressions
-in a workflow definition. If a different expression language needs to be used, make sure to set the workflow 
-`expressionLang` property to identify it to runtime implementations.
+Expressions are evaluated during execution of the workflow and can be used for the following purposes:
+- Transform state data
+  - upon entering and leaving a state, using [State Data Filter](#State-data-filters) `input` and `output`
+  - entering and leaving an action, using [Action Data Filter](#Action-data-filters) `fromStateData`, `result`
+  - when consuming an event, using [Event Data Filter](#Event-data-filters) `data`
```
tsurdilo, line 1366: i think it might be missing producing events here

specification.md
```diff
@@ -1357,44 +1359,86 @@ For more information about workflow expressions, reference the [Workflow Express
...
+- Select a path in the workflow data
+  - to set the result of an action to, using [Action Data Filter](#Action-data-filters) `toStateData`
+  - to set the result of consuming an event to, using [Event Data Filter](#Event-data-filters) `toStateData`
+- Define functions that can be executed as an action [using functions for expression evaluation](#Using-Functions-For-Expression-Evaluation)
+- Define any state parameter of the workflow definition.
+  Basically any leaf parameter in states can be set by evaluating an expression against the workflow state data at runtime. However, structures like the `actions` array in an [operation state](#Operation-state) can not be generated at runtime, i.e. the workflow definition has to conform to the JSON specification and contain all required elements. The JSON specification explicitly allows expressions where applicable.
```
tsurdilo, line 1372: don't use "state" alone and then "workflow state" . probably best to keep it consistent. i know this should be done throughout document so it can be done later

specification.md
```diff
@@ -1357,44 +1359,86 @@ For more information about workflow expressions, reference the [Workflow Express
...
+
+A workflow expression can be defined where it is applied as an _expression string_ 
+or it can be defined as an [expression function definition](#Using-Functions-For-Expression-Evaluation) 
+and then applied using an _expression reference string_.
+
+An _expression string_ starts with the `${` token and ends on `}`. All text 
+between the curly brackets is evaluated using the workflow's specified 
+[`expressionLang`](#Workflow-Definition-Structure).
+Note, that without the closing bracket at the very end of the string, it won't
```
tsurdilo, line 1381: trimming spaces is available in every programming language..why do we make this restriction it sounds like we dont trust people..even in IDE you can set "trim spaces on save" :)

specification.md
```diff
@@ -1490,46 +1570,34 @@ This would set the data output of the particular state to:
...
 }
 ```
 
-we can use this expression in the workflow "version" parameter:
+The function name `IsAdultApplicant` can be referenced anywhere where expressions are allowed.
+Its return type is boolean, so it makes a useful condition:
 
 ```json
 {
-   "id": "MySampleWorkflow",
```
tsurdilo, line 1522: why are we removing this example w/o some sort of clear description . maybe im looking at diff wrong but if its removed its another ghost removal

specification.md
```diff
@@ -3258,7 +3326,7 @@ If events defined in event-based conditions do not arrive before the states `eve
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
 | name | Data condition name | string | no |
-| [condition](#Workflow-Expressions) | Workflow expression evaluated against state data. Must evaluate to true or false | string | yes |
+| [condition](#Workflow-Expressions) | Boolean or workflow expression that evaluates to boolean against state data | boolean or expression | yes |
```
tsurdilo, line 3329: again there is no "boolean or workflow expression" its just an expression that must return boolean

specification.md
```diff
@@ -3770,7 +3838,7 @@ You can also use the filter property to filter the state data after data is inje
         ]
      },
      "stateDataFilter": {
-        "output": "${ {people: [.people[] | select(.age < 40)]} }"
+        "output": "${people: [.people[] | select(.age < 40)]}"
```
tsurdilo, line 3841: invalid expression

specification.md
```diff
@@ -3797,7 +3865,7 @@ You can also use the filter property to filter the state data after data is inje
       address: 1234 SomeStreet
       age: 30
   stateDataFilter:
-    output: "${ {people: [.people[] | select(.age < 40)]} }"
+    output: "${people: [.people[] | select(.age < 40)]}"
```
tsurdilo, line 3868: invalid expression

specification.md
```diff
@@ -3829,7 +3897,7 @@ In which case the states data output would include only people whose age is less
 You can change your output path easily during testing, for example change the expression to:
 
 ```text
-${ {people: [.people[] | select(.age >= 40)]} }
+${people: [.people[] | select(.age >= 40)]}
```
tsurdilo, line 3900: invalid expresssion

specification.md
```diff
@@ -3841,9 +3909,9 @@ This allows you to test if your workflow behaves properly for cases when there a
 | id | Unique state id | string | no |
 | name | State name | string | yes |
 | type | State type | string | yes |
-| inputCollection | Workflow expression selecting an array element of the states data | string | yes |
-| outputCollection | Workflow expression specifying an array element of the states data to add the results of each iteration | string | no |
-| iterationParam | Name of the iteration parameter that can be referenced in actions/workflow. For each parallel iteration, this param should contain an unique element of the inputCollection array | string | yes |
+| inputCollection | Array or a workflow expression creating an array | array or expression | yes |
```
tsurdilo, line 3912: what is a "workflow expression creating an array" ? creating it where? why? how?

specification.md
```diff
@@ -3841,9 +3909,9 @@ This allows you to test if your workflow behaves properly for cases when there a
...
+| outputCollection | Workflow expression specifying a path where an array is created that collects the results of each iteration | path expression | no |
```
tsurdilo, line 3913: whats a "path" in this sence. 
also outputcollection does not have to be created..it can exist already

specification.md
```diff
@@ -5157,43 +5225,37 @@ Workflow constants should not have access to [Workflow secrets definitions](#Wor
...
-inside your [Workflow Expressions](#Workflow-Expressions). 
+Secrets allow you access sensitive information, such as passwords, OAuth tokens, ssh keys, etc.
+They can be used in [Workflow Expressions](#Workflow-Expressions).
 
 You can define the names of secrets via the [Workflow top-level "secrets" property](#Workflow-Definition-Structure), 
 for example:
 
 ```json
-"secrets": ["MY_PASSWORD", "MY_STORAGE_KEY", "MY_ACCOUNT"]
+"secrets": ["MY_ACCOUNT", "MY_PASSWORD", "urn:example:tenant:accessToken"]
```
tsurdilo, line 5235: cant this be done as a separate pr for secrets and not ghost added to this one please? if you would like this change then lets just create a separate pr for it as idk what its got to do w/ expressions

specification.md
```diff
@@ -5157,43 +5225,37 @@ Workflow constants should not have access to [Workflow secrets definitions](#Wor
...
+If secrets are only named in a Workflow definition, runtimes must assure to 
+provide their values during execution.
 
-Secrets can be used only in [Workflow expressions](#Workflow-Expressions) under the `SECRETS` namespace. 
-This is reserved namespace that should only be allowed for values defined by the `secrets` property.
-
-Here is an example on how to use secrets and pass them as arguments to a function invocation:
+Secrets can be used only in [workflow expressions](#Workflow-Expressions).
+The default expression language JQ provides secrets through the reserved 
+`$SECRETS` variable. Here is an example on how to use secrets and pass them as
```
tsurdilo, line 5243: i like namespace rather than "variable" please

specification.md
```diff
@@ -5157,43 +5225,37 @@ Workflow constants should not have access to [Workflow secrets definitions](#Wor
...
-        
 {
   "refName": "uploadToAzure",
     "arguments": {
-      "account": "${ $SECRETS.AZURE_STORAGE_ACCOUNT }",
-      "account-key": "${ $SECRETS.AZURE_STORAGE_KEY }",
-      ...
+      "user": "${ $SECRETS.MY_ACCOUNT }",
+      "pass": "${ $SECRETS.MY_PASSWORD }",
+      "token": "${ $SECRETS.\"urn:example:tenant:accessToken\" }"
```
tsurdilo, line 5252: this is so ugly to me :( i dont think there is a reason for it really @cdavernas am i missing something?

tsurdilo:
Thanks for the PR, left some comments. 
Overall I think that leaving the things the way there are is probably best than taking this but thats just my opinion.
@cdavernas do yo think these changes are a must have in order to get this working in Synapse? 
From what i see actually this makes the parsing work for runtimes almost impossible to do (or much much harder) than currently..but idk. 

I would like to see the updates to secretes in a separate pr please if you have the time. Thanks

specification.md
cdavernas: @tsurdilo What is ugly? What is done with the `token` property? If yes, I agree. From where I stand I do not see any need for supporting such ugly (and dangerous) practices.

tsurdilo: @cdavernas yes, the "token" property value

cdavernas:
@tsurdilo as for now, the only missing parts regarding the spec, IMHO, are a some kind of auth component and maybe additional actions (produce event, consume event, throw error, ...).

Aside from that, from my understanding of the current state of the spec, nothing is missing, all's good, I do not personally have the need for anything else.

Regarding expressions, my understanding of how it works, as currently implemented in both Synapse and in a private implementation for a client, I do not have any problem either. 

Here's what I'm doing (nb: not saying it's the/a right way to do it):
- if a string starts with `${` and ends with `}` then I consider the string[2..^1] as a SINGLE expression, meaning that `${ { .greetings }  }` would work, but `${ .greetings } to my ${ .relative }` would not.
- I then check that jq expression string for function expressions, which I then replace with their invocation result
- check for constants, then replace as is by the value (string or otherwise) at the specified path
- check for secrets, then replace as is by the value (string or otherwise) at the specified path
- run a jq process with the state/branch/iteration/action data and the resulting expression as is
- parse the json output of jq as is
- apply filtering (actiondatafilters, etc)

schema/common.json
manuelstein: when we markup expressions with `${ }` and allow `${ fn:Function }`, then we first say that everything inside `${}` should be interpreted by the expression language, which may be JQ or something else, meaning you don't know how to interpret what's inside `${}` and it's up to the interpreter to compute that. But then we say that if fn:Function appears inside the markup, we don't care what the interpreter would do with it but instead, we invoke a named expression function.

schema/common.json
manuelstein: This is not about the result type of the expression. It's about declaring fields that can also be defined using an expression. Let's say we want to define a parameter that's supposed to be a number, e.g. number of repeats. If you also want to allow an expression, it can be either a string with format `${ }` or it can be a number, but it can neither be a random string nor an object. This is about getting the workflow definition parsed with useful types.

schema/workflow.json
manuelstein: Agree, it is easier to read when keeping "contains" like it was before. So "A string (or an expression returning a string) containing a valid GraphQL selection set" is easier to read.
The stringOrMarkup is to allow the parser to already detect whether it's an expression that should be fed to the interpreter or whether it's a string that can remain what it is.

schema/workflow.json
manuelstein: I think we want to allow expressions or objects. "objectOrMarkup" is pretty much that, it allows to either specify an object or an expression. But also, it allows to specify objects that is built from several expressions:
```json
"data": {
  "foo": "bar",
  "dynamic": "${ .someValue }"
}
```

schema/workflow.json
manuelstein: tried to clarify above. maybe there's a less technical description that is easier to understand. any suggestions?

schema/workflow.json
manuelstein: Have you ever altered your code to `if ( true )` or `if ( false )` for testing? Maybe it's bad style, but I thought that only the result counts, i.e. whatever is in there should eventually be boolean.

schema/workflow.json
manuelstein: Maybe you want to construct your array:
```
inputCollection: [
  "alwaysThis",
  "${ .someMoreElements }",
  "${ .different.path.to.some.moreElements }",
  "andAlsoThis"
]
```

specification.md
manuelstein: a link is a good idea, thx!

specification.md
```diff
@@ -251,8 +251,8 @@ The workflow data output is the data output of the last executed workflow state.
 
 ```yaml
 stateDataFilter:
-  input: "${ .orders }"
-  output: "${ .provisionedOrders }"
+  input: "${.orders}"
```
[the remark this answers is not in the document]
manuelstein: that was just coincidental, because we discussed different markups while in the making. I changed it in bulk so it got rid of the spaces. They're okay to keep if you prefer. I'll have to redo all changes on the latest main anyways and will keep the spaces.

specification.md
manuelstein: thx, that is wrong

specification.md
```diff
@@ -327,8 +327,8 @@ The first way would be to use both "input", and "output":
 ```json
 {
   "stateDataFilter": {
-    "input": "${ {vegetables: .vegetables} }",
-    "output": "${ {vegetables: .vegetables[] | select(.veggieLike == true)} }"
+    "input": "${vegetables: .vegetables}",
+    "output": "${vegetables: .vegetables[] | select(.veggieLike == true)}"
```
manuelstein, line 331: also wrong

specification.md
manuelstein: alo wrong, yes. replacement error

specification.md
manuelstein: same again

specification.md
manuelstein: that is mainly for discussion because we're not explaining that the expression used in toStateData is auto-completed by a runtime, e.g. `${ .path.to.element }` will be executed as
`jq ' .path.to.element  = $result' --argjson result '{...}' /statedatafile`
I agree we need to be clear about it. If, however, we want to pass the result to the expression, e.g. through $action.result, we could avoid this `path expression` type and make the default expression `${ $action.result }`

specification.md
manuelstein: well, this is the expression discussion all over again. I agree it would be simpler, but this doesn't compute without a clean specification for how string interpolation works.

specification.md
manuelstein: The expression function `name` property type is string, so the name of the expression function can be any arbitrary string including spaces, dots, colons, and question marks. It can end with a space and it can start with a space. When it does, it should be different from when it doesn't, e.g. `"aFunction"` is different from `" aFunction"`

specification.md
manuelstein: Well, we could say that every string starting with ${ is evaluated as expression. At some point during the discussion, $: was proposed and everything following that would be the expression, so one doesn't have to bother closing the statement.
The closing bracket is just for style, really. So, are you proposing we allow whitespaces after `}`? Those won't appear in the result...

specification.md
manuelstein: The version field, or any workflow header, can't be set by an expression. It's not within any state.

specification.md
manuelstein: again, can be set to json true/false if eventually we only care that what is in here is boolean.

specification.md
manuelstein: thx, same as the other examples that got reduced once too often. will fix

specification.md
manuelstein: the result of the expression has to be an array, doesn't it? Or did I miss something? Can we have a single object as inputCollection?

specification.md
manuelstein: that's the complication we have from not passing the output to the expression.

specification.md
manuelstein: secrets can be an array of strings. Maybe I missed something, but I think it can be an arbitrary string, no? A urn is just a string, so nothing is being ghost-added

specification.md
manuelstein: Well, there's no namespacing concept in JQ and I tried to explain what the secrets definition becomes inside JQ. In its own terms its a predefined variable.

specification.md
manuelstein: That's basic JQ. There's currently no restriction on the string - neither in the secrets definition nor in JQ. Not sure what it is you don't want to support? Do you want to restrict secret names to be capital letter snake case?

specification.md
manuelstein: Sorry, I thought I had moved the text about immutability elsewhere. Or probably because they are passed to the JQ execution as a predefined variable that is not extracted after the interpreter has finished, it can only be immutable. I'll add it back, so that it's clear to the user that secrets can't be set.

Well, technically, the variable can be overwritten during the JQ execution, so it can be "set", it only doesn't propagate back. hmm, maybe neglecting the details and just saying it's immutable is enough for the user to know

specification.md
```diff
@@ -1357,44 +1359,86 @@ For more information about workflow expressions, reference the [Workflow Express
...
+parameters if you want to set them using an expression.
+
+The default expression language is [jq](https://stedolan.github.io/jq/).
+This version of the workflow language supports jq [version 1.6](https://github.com/stedolan/jq/releases/tag/jq-1.6) syntax.
+You can find more information on jq v1.6 in its [manual](https://stedolan.github.io/jq/manual/v1.6/).
+
+JQ expressions provide the state data as the root document (`.`). 
+Global [secrets](#Workflow-secrets) and [constants](#Workflow-constants) can be 
+accessed through the jq variables `$SECRETS` and `$CONST` respectively, which 
+makes them reserved variable names in JQ expressions.
```
gibson042, line 1425: There's probably a chance of wanting to expose _other_ data in the future as well, so it would make more sense to reserve either a pattern (e.g., a leading underscore with `$_SECRETS` and `$_CONST` being the first two uses) or a single object-valued variable (e.g., `$SPECIAL` with `SECRETS` and `CONST` keys).

specification.md
```diff
@@ -1357,44 +1359,86 @@ For more information about workflow expressions, reference the [Workflow Express
...
+  - to set the result of consuming an event to, using [Event Data Filter](#Event-data-filters) `toStateData`
+- Define functions that can be executed as an action [using functions for expression evaluation](#Using-Functions-For-Expression-Evaluation)
+- Define any state parameter of the workflow definition.
+  Basically any leaf parameter in states can be set by evaluating an expression against the workflow state data at runtime. However, structures like the `actions` array in an [operation state](#Operation-state) can not be generated at runtime, i.e. the workflow definition has to conform to the JSON specification and contain all required elements. The JSON specification explicitly allows expressions where applicable.
+
+A workflow expression can be defined where it is applied as an _expression string_ 
+or it can be defined as an [expression function definition](#Using-Functions-For-Expression-Evaluation) 
+and then applied using an _expression reference string_.
+
+An _expression string_ starts with the `${` token and ends on `}`. All text
```
gibson042, line 1378: This reads like `}` terminates the expression string, which implies that `${greeting}, world` would be valid. But for that matter, I think it is a mistake to consume everything between a start-of-string `${` and an end-of-string `}`, because that precludes _ever_ supporting interpolation without backwards incompatibility (because `${greeting}, ${noun}` is already valid and interpreted as "results of evaluating `greeting}, ${noun` as an expression"). Sorry to [repeat myself](https://github.com/serverlessworkflow/specification/issues/285#issuecomment-853431146), but total-replacement behavior would be communicated **so** much better by a prefix rather than a syntax that looks like it can be embedded.

specification.md
```diff
@@ -1357,44 +1359,86 @@ For more information about workflow expressions, reference the [Workflow Express
...
+starts with the `$fn{` token and ends with `}`.
+The _expression reference string_ has the format `$fn{<function name>}`, where 
+`<function name>` has to match the user-specified `name` of the defined 
+expression function of `type` `"expression"`.
+
+On the rare occasion that one needs to create a string that literally looks like 
+an _expression string_ or an _expression reference string_, the dollar sign 
+should be repeated. Only if the remainder matches a workflow expression markup,
+the runtime reduces the double dollar signs `$$` to a single dollar sign `$` and 
+refrains from evaluating it as a workflow expression.
```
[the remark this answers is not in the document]
manuelstein: Thanks, personally I wouldn't replace any occurence of a `$` at the start of an expression-capable string. Only the ones that could actually be interpreted as an expression (`${...}` or `$fn{...}`). I agree any markup like `$xyz{ }` could be included (e.g. regex `^\$.+\{.*\}`) that would allow for more than just the function reference. I was hoping to get a discussion on $fn{}. The existing markup `fn:` causes confusion for jq and maybe causes problems in other expression languages, too.

specification.md
manuelstein: so, as we said to have no separate interpolation markup, that's the way strings can be created from jq.

schema/common.json
manuelstein: First of all, this is not ghost-adding it. You've specified that the toStateData would point to elements in state data that the results should be written to. So, you've already added pointers, which are not jq expressions.
But I agree, so I removed the jsonpointer thingy. I think the outcome of merging results into state data would be much clearer if it's done in the expression language providing the result in a variable. That way, we can resolve:
- @jorgenj's issue on whether deep merge or flat merge is done
- the error in the merge description in the spec, which does not match what jq does (neither `*` nor `+` operations would neither unique-merge nor concatenate nested arrays/lists)

specification.md
manuelstein: thx, same search/replace error.

specification.md
manuelstein: removed `path expression` in favor of not having pointer-like fields in the spec.

specification.md
manuelstein: yes, thx, the search/replace struck back again.

specification.md
manuelstein: search/replace hit me again, sry

specification.md
manuelstein: Ok, as with jsonpointer, removed the path expression thingy to not have them in the spec

schema/common.json
```diff
@@ -4,6 +4,103 @@
   "description": "Serverless Workflow specification - common schema",
   "type": "object",
   "definitions": {
+    "expressionMarkup": {
+      "type": "string",
+      "pattern": "^\\${.*}$"
```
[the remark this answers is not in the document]
manuelstein: thx, changed in update.

specification.md
```diff
@@ -1357,44 +1359,86 @@ For more information about workflow expressions, reference the [Workflow Express
...
+expression can only access data that has been passed to that state, inside an action, 
+only the data filtered by the [Action Data Filter](#Action-data-filters) `fromStateData` 
+parameter are available to generate arguments to an action, etc.
+Please refer to the specification of filters for more inforamtion, 
+e.g. [State Data Filter](#State-data-filters), [Action Data Filter](#Action-data-filters), 
+[Event Data Filter](#Event-data-filters) and see the specification of individual 
+parameters if you want to set them using an expression.
+
+The default expression language is [jq](https://stedolan.github.io/jq/).
+This version of the workflow language supports jq [version 1.6](https://github.com/stedolan/jq/releases/tag/jq-1.6) syntax.
```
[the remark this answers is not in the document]
manuelstein: Yes, it's meant to be jq 1.6. Previous versions of jq have issues and I think it is valid for a spec to fix it to a specific version. Evaluating what jq adds in future versions, we can assess if we want to update this to be only a minimum requirement.

specification.md
manuelstein: Hmm, I tried to make it clear that the string has to end on `}` and that all in between is an expression, no interpolation. Can you suggest a better wording that makes this clear? Doesn't read like interpolation to me, maybe I've grown blind to it.

gibson042: I guess the "closing bracket at the very end of the string" clears it up. But what about the other point, regarding prefix-based syntax being more future-compatible and less misleading than the not-actually-embeddable open–close syntax?

specification.md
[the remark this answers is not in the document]
gibson042: What would such a spec change look like, and how would a workflow specify jq version 1.6 after the spec updates to allow support for later versions? I think it would be more accurate to update these lines to something like
> The default expression language is [jq version 1.6](https://github.com/stedolan/jq/releases/tag/jq-1.6). A future revision may update this default, requiring workflows that require continued use of that exact language to specify it explicitly.

tsurdilo:
this issue has not received any action in quite some time. we still want to address this at some point or continue the work? 
if not we should close the pr and anyone can feel free to reopen it when they feel it should be active again

specification.md
manuelstein: I think I would agree but I have to defer this proposal to become its own PR.

specification.md
manuelstein: Prefix is sufficient, but when we've jointly discussed as preparation to this PR, we want 2 things: least friction and clarity. So, here I'm trading a bit of clarity to go with less friction to the existing mark-up.

manuelstein:
I had tried to address your comments, @tsurdilo. But there are some comments that are rather "blocking" in nature, i.e. the ones where you're saying you see no reason. I've tried to provide reason and that's where the conversion ended.

You had requested to do a PR instead of discussing it as an issue. I was hoping that it would be easier to collect approval for this bugfix once the PR is on the table. Adopting business practice, I'd take no answer for an agreement.

So, considering summer vacation etc. I believe two months have been plenty of time to follow-up. So let me put a reasonable **deadline for requests: Monday, 4th October 2021**. After the deadline, I'll redo the work once again on the latest spec.

PR #417, opened 2021-07-01
OData function

cdavernas:
Signed-off-by: charles.davernas <charles.davernas@neuroglia.io>

**Many thanks for submitting your Pull Request :heart:!**

**Please specify parts this PR updates:**

- [x] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

**What this PR does / why we need it**:

Adds OData support to Functions

PR #419, opened 2021-07-07
Async API Operation Support

cdavernas:

- [x] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

Adds a new `asyncapi` operation type, which enables support for the [Async API Specification](https://www.asyncapi.com/docs/specifications/v2.1.0). 

Async API support means that, at runtime's discretion, users could connect to a myriad of defined channels in an abstract and agnostic fashion.

### DIFF
--- a/schema/odata.json
+++ b/schema/odata.json
@@ -1,0 +1,81 @@
+{
+    "$id": "https://serverlessworkflow.io/schemas/0.7/odata.json",
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "description": "Serverless Workflow specification - OData arguments schema",
+    "type": "object",
+    "id": {
+        "type": "string",
+        "description": "The unique identifier of the single entry to query",
+        "minLength": 1
+    },
+    "queryOptions":{
+        "$ref": "#/definitions/queryoptions"
+    },
+    "definitions": {
+      "queryoptions": {
+        "type": "object",
+        "properties": {
+          "filter": {
+            "type": "string",
+            "description": "The $filter system query option allows clients to filter the set of resources that are addressed by a request URL. $filter specifies conditions that MUST be met by a resource for it to be returned in the set of matching resources",
+            "minLength": 1
+          },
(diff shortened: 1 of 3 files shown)


schema/odata.json
```diff
@@ -0,0 +1,81 @@
+{
```
tsurdilo, line 1: Are any of these properties required?

For integer types i think you should not use minLength but for example:

"minimum": 0,

Consider setting default value for boolean type fields (if you think its needed)

tsurdilo (approved):

tsurdilo:
@cdavernas very nice! ty. Left a small comment regarding the schema def.

schema/odata.json
cdavernas: Oh no 😠...
It seems I made that `asyncapi` branch from the `odata` in #417

tsurdilo: would it be easier then to do them together in this one ?

cdavernas: Yeah, maybe, if that's not a problem for anyone

cdavernas:
@tsurdilo So far, my PR would support the following:
1. A single AsyncApi PUB request
2. A single AsyncApi SUB request: would actually wait for a single message

I'd we add a property such as `ReadMode` (ReadOne, ReadN or ReadToEnd) or whatever to the AsyncApi `FunctionDef`, we'd have a function able to basically observe a subscription, the RX way. From there to streaming, the step is very small...

tsurdilo:
needs rebase

schema/odata.json
ricardozanini: not from my side, we are expecting the async.json file, right? :sweat_smile:

tsurdilo:
Replaced with https://github.com/serverlessworkflow/specification/pull/434
Will close this one

PR #435, opened 2021-08-02
Error and Retry handling update

tsurdilo:

- [x ] Specification
- [ x] Schema
- [x ] Examples
- [ ] Extensions
- [ x] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

Updates the workflow error and retry handling . To be added to 0.7 spec release.

* Big update (backwards compat change): state actions now are retried by default. You can define specific retry definition, if not defined runtimes can define the default retry strategy.
* (backwards compat change): state actions can now define errors for which they are NOT retried
* state onErrors do no longer define retries
* added top-level "errors" array which can be in-line array or reference to a json/yaml external/reusable file. All known errors that can occur during workflow execution have to be defined here and then referenced in workflow control flow logic

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -71,7 +71,9 @@
       - [Additional Properties](#additional-properties)
   * [Workflow Error Handling](#workflow-error-handling)
     + [Defining Errors](#defining-errors)
-    + [Defining Retries](#defining-retries)
+  * [Action Retries](#action-retries)
+    + [Retry actions on known errors](#retry-actions-on-known-errors)
+    + [Automatic retries on known and unknown errors](#automatic-retries-on-known-and-unknown-errors)
   * [Workflow Timeouts](#workflow-timeouts)
     + [Workflow Timeout Definition](#workflow-timeout-definition)
       - [WorkflowExecTimeout Definition](#workflowexectimeout-definition)
@@ -1760,6 +1762,7 @@ definition "id" must be a constant value.
 | [auth](#Auth-Definition) | Workflow authentication definitions | array or string | no |
 | [events](#Event-Definition) | Workflow event definitions.  | array or string | no |
 | [functions](#Function-Definition) | Workflow function definitions. Can be either inline function definitions (if array) or URI pointing to a resource containing json/yaml function definitions (if string) | array or string| no |
+| autoRetries | If set to true, [actions](#Action-Definition) should automatically be retried on unchecked errors. Default is false | boolean| no |
 | [retries](#Retry-Definition) | Workflow retries definitions. Can be either inline retries definitions (if array) or URI pointing to a resource containing json/yaml retry definitions (if string) | array or string| no |
 | [states](#Workflow-States) | Workflow states | array | yes |
 | [metadata](#Workflow-Metadata) | Metadata information | object | no |
@@ -3643,7 +3646,8 @@ This is visualized in the diagram below:
 | [eventRef](#EventRef-Definition) | References a `trigger` and `result` reusable event definitions | object | yes if `functionRef` & `subFlowRef` are not defined |
 | [subFlowRef](#SubFlowRef-Definition) | References a workflow to be invoked | object or string | yes if `eventRef` & `functionRef` are not defined |
 | [retryRef](#Defining-Retries) | References a defined workflow retry definition. If not defined uses the default runtime retry definition | string | no |
-| nonRetryableErrors | List of references to defined [workflow errors](#Defining Errors) for which the action should not be retried | array | no |
+| nonRetryableErrors | List of references to defined [workflow errors](#Defining Errors) for which the action should not be retried. Used only when `autoRetries` is set to `true` | array | no |
+| retryableErrors | List of references to defined [workflow errors](#Defining Errors) for which the action should be retried. Used only when `autoRetries` is set to `false` | array | no |
 | [actionDataFilter](#Action-data-filters) | Action data filter definition | object | no |
 | sleep | Defines time periods workflow execution should sleep before / after function execution | object | no |
 
@@ -3711,7 +3715,12 @@ should be retried according to the default retry policy of the runtime implement
 retries reference [this section](#defining-retries).
 
 The `nonRetryableErrors` property is a list that references one or more unique names of workflow error definitions. 
-This is the list of known errors for which the action should not be retried for.
+This is the list of known errors for which the action should not be retried for. 
+It should be used only when the workflow top-level `autoRetries` property is set to `true`.
+
+The `retryableErrors` property is a list that references one or more unique names of workflow error definitions.
+This is the list of known errors for which the action should be retried for.
+It should be used only when the workflow top-level `autoRetries` property is set to `false`.
 
 ##### Subflow Action
 
@@ -4022,7 +4031,7 @@ jitter: PT0.001S
 </details>
 
 Defines the states retry policy (strategy). This is an explicit definition and can be reused across multiple
-defined workflow state errors.
+defined state [actions](#Action-Definition).
 
 The `name` property specifies the unique name of the retry definition (strategy). This unique name
 can be referred by workflow states [error definitions](#Error-Definition).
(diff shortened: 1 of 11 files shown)


tsurdilo:
This is still Work in Progress - still need to update the specification documentation 
and all of the samples json/yaml

specification.md
```diff
@@ -4890,40 +4899,29 @@ actions `nonRetryableErrors` property.
...
-a workflow action definition.
-
-TODO - Finish this section!
-
+Runtime implementations can define their own default retry policy. Serverless Workflow recommends the following settings:
 
-TODO - Rewrite this section!
+* `maxAttempts` to be `unlimited`, meaning that the action should be retried indefinitely until successful.
+* `delay` to be set to one second, meaning that there is a one second delay between action retries.
+* `multiplier` to be set to two meaning that the delay should be multiplied by two for each retry attempt.
```
ricardozanini, line 4916:
```suggestion
* `maxAttempts` to be `unlimited`, meaning that the action should be retried indefinitely until successful.
* `delay` to be set to one second, meaning that there is a one second delay between action retries.
* `multiplier` to be set to two, meaning that the delay should be multiplied by two for each retry attempt.
```

specification.md
```diff
@@ -4977,30 +4984,65 @@ Different states now can use the defined retry strategy. For example:
...
 
 </td>
 </tr>
 </table>
 
-In this example we say that if the  "Inventory service timeout" error is encountered, the workflow should transition to the next state
-as defined by the error definitions `transition` property.
+In our example the first action named `MyFirstFunction` is going to be retried according to the `FirstRetryStrategy`
+retry policy
+for all errors except `SomeErrorOne` and `SomeErrorTwo`.
```
ricardozanini, line 5036:
```suggestion
In our example, the first action named `MyFirstFunction` is going to be retried according to the `FirstRetryStrategy`
retry policy
for all errors except `SomeErrorOne` and `SomeErrorTwo`.
```

specification.md
```diff
@@ -4977,30 +4984,65 @@ Different states now can use the defined retry strategy. For example:
...
+
+The seconds action named `MySecondFunction` is going to be retried according to the `SecondRetryStrategy`
+retry policy 
+for all errors except `SomeErrorTwo` and `SomeErrorThree`.
```
ricardozanini, line 5040:
```suggestion
The second action named `MySecondFunction` is going to be retried according to the `SecondRetryStrategy`
retry policy 
for all errors except `SomeErrorTwo` and `SomeErrorThree`.
```

specification.md
```diff
@@ -4977,30 +4984,65 @@ Different states now can use the defined retry strategy. For example:
...
+
+The third action named `MyThirdFunction` is going to retried according to the default runtime retry policy.
```
ricardozanini, line 5042:
```suggestion
The third action named `MyThirdFunction` is going to be retried according to the default runtime retry policy.
```

ricardozanini (approved):

cdavernas (commented):
@tsurdilo Nice work, cheers!

However, as discussed with you yesterday, IMHO we should take the opposite approach concerning retries. 

As a matter of fact, I believe retries should be explicitly defined for specific errors. I understand your point of view concerning extremely verbose definitions in case you want to define retry strategies for multiple kinds of error, but that could very well be solved by allowing the use of wildcards/regexes. With explicitly defined retry strategies, runtimes can easily implement extremely powerfull features, leveraging libraries such as [Polly](http://www.thepollyproject.org/) (which we should IMO base ourselves on, bringing features such as circuit breakers and the like).

Here are my concerns with the way you propose retry strategies to be implemented:

1. Counter-intuitive on most modern programing languages, which use explicit try/catch/finally statements.
2. Possible DDOS on services that throw unhandled exceptions because of implementation errors.
3. Good programing practice suggest that processes should fail fast, so that they can be fixed fast too. Your proposal goes against that concept, because you consider that workflows should ideally never fail. I personnaly have no problem with failing workflows, that can be fixed/compensated with explicit user interaction, like our private workflow dashboard implementation allows. It provides users with a far greater control over flows, and allows them to cherry pick fixes.
4. Near impossible to implement reliable circuit breakers.
5. Incompatibility with idempotency, distributed transactions and sagas
6. No reason to retry calling a service that returns 4** or 5** status codes, for example: if such codes are returned, the workflow is faulted, not the services. Retrying infinitly won't change a thing: you'll have to fix the workflow. Plus, I'd rather not explicitly say: do not retry for 400, 401, 403, ... => it is far too obvious no action (but potentially compensation) should be undertaken in those cases

@tsurdilo @ricardozanini @manuelstein WDYT?

tsurdilo:
@cdavernas I understand what you are saying, but the fact is that the "traditional" way of thinking or using a programming approach for workflows is imo not optimal.

With workflows you *do not* want to fail on any random / unchecked exception. This is the "fault tolerance" part of workflows that so far has been completely ignored by most workflow runtimes. I'll try to explain:
* workflows should be written / designed so they do not fail on unchecked/unknown exceptions. this is the core part that we need to embrace, especially with DSL workflows. DSL workflows are typically used for very short-running, pipe-line type executions. we are with our DSL redefining this, allowing for DSL workflows to take on very large and long-running, mission critical executions. As such we *have* to think about fault tolerance. Workflows are built / used to complete, and this "traditional" approach you mention yes its been used so far, but does not apply here.

To go through your points:
1. Tried to clarify above, yes workflows have to take a different approach. try/catch approach as in programming language concepts does not provide failure tolerance and possible recovery and this cannot be easily expressed in any currently existing DSLs (or flow chart based workflow langs).
2. Yes and this is handled by rate limiting on actions. Runtimes provide rate limiting anyways, we will be adding this as part of our dsl (similar to what we did with timeouts) for actions for the next release. Agree this is an issue, but it is also an issues regardless of this change/update
3. Yes this is a "philosophical" question, but in the end it is the "services" that need to fail fast, workflows on the other hand should not just fail or be retried. That is the what I am trying to "preach" and I think is a big thing that really we need to look at and change our mind sets about.
4. I think it is possible, let's work on an example together on this
5. Idempotency of actions? Not seeing that point, can you please explain? It is perfectly working fine along side saga/compensation. This actually has nothing to do with compensation at all. I can explain if needed.
6. Yes that's why we added the "nonRetryableErrors" property. That's the best way to assure that you explicitly don't retry on checked/known errors, and allow for workflow onErrors to handle them properly
7. I don't agree with that sorry :) Unknown erros are a bad thing and workflows should again not fail on them. Failing on unknown errors introduces a TON more issues again with fault tolerance and recovery.

I think really that in the end we should have a small meet where I can properly introduce this PR as it really is very important on my personal side that we are on the same page.

JBBianchi:
I don't have an in depth understanding of the DSL, but intuitively, as a developer, my mind would rather follow @cdavernas point of view. 

For me, it "kind of" makes sense that if an error occurred (other than service availability error, but an actual error returned by a service), retrying will produce the same error again and again (at least if your service is "pure", the same input should always produce the same output). 

But it might indeed be a good idea @tsurdilo to have a little meet to exchange on the matter, maybe it's just a matter of understanding better the how and why, we might not be seing the objective here from the same angle.

cdavernas:
@tsurdilo First of all, thanks for your quick answer, as usual ;)

Second, allow me to go over some of your answers, leaving the other open for later discussions:

3. IMO, if the workflow itself contains errors, by for instance passing incorrect data to actions, the workflow should fail fast too, for it is reponsible for things no going as expected.
5. Idempotency of the commands that are sent to services via actions, yes. Retrying an idempotent action will have exactly the same outcome than first time, thus failling over and over.
6. I know, and that's the problem IMO. `nonRetryableErrors` would force workflow authors to always specify a countless amount of errors for which not to retry. Typically, for all 400 and 500. Whereas explicit retry strategies would usually be implemented for 503, for example. So, IMHO, defining one retry strategy for a 503 is less of a pain than declaring potentially dozens of `nonRetryableErrors`.
7. Ok, I see where you are coming from on that. I too agree that the DSL should be as fault tolerant as possible, which is doable in both cases anyways. My problem is rather that your way of seeing things implies that one should develop against unknown errors, rather than against well known, well documented ones.

Agreed with the meet idea! I am, and remain, at your disposition. I just think this is a big bang kind of PR, and that we need to really get to the bottom of all use-cases before pushing it in. I'd really like to discuss it with other implementers, such as @ricardozanini, @manuelstein and @jorgenj, too, and see what's their take on it. Slack me anytime this week if you want to tackle the issue ASAP: I'd be happy, as usual, to talk with you, as it is always both a pleasure and usefull ;)

tsurdilo:
@cdavernas nice thank you!

regarding the further points:
3. Yes, the workflow still can fail on unknown errors (and fast). You can always set maxAttempts for action retries to 1 to say "dont retry this action". If you want resilience/tolerance then use either default retry option or set a custom one.
4. Actions are still idempotent. Retries just allow you to try to resolve any issues. Remember for some long running workflows that run for a year+ retrying for a week is "nothing" really.
5. Good point, this however imo can be solved by grouping errors, being able to set a number of "codes" per error, etc. So this is a DSL improvement imo more than an issue
6. This way it works for **both** the only thing that this changes is that we can handle unknown/unexpected failures much much much better :)

ricardozanini:
> I think really that in the end we should have a small meet where I can properly introduce this PR as it really is very important on my personal side that we are on the same page.

Let's do it before this PR got too verbose to follow :)

tsurdilo:
@cdavernas note that if this really is some issue (after we discuss stuff more) it would be easy to have a flag that sets the "automatic retries" option. If then set to false, this pr would behave the "traditional" way and would be still a major improvement imo :)

tsurdilo:
@ricardozanini yeah i agree, lets talk on Slack to figure out a good time for whoever would like to attend in the next couple days.

tsurdilo:
@cdavernas per discussion today added the autoRetries prop (defaults to false) so we support the explicit as well as auto retries options. Documented those in spec doc and updated samples.

PR #460, opened 2021-08-19
Add support for custom function type definitions

ricardozanini:

**Please specify parts of this PR updates:**

- [x] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

In this PR, we add the possibility for implementations/runtimes to offer their own function `type` for users, albeit respecting the required attributes. This way, runtimes won't be limited to the set of functions supported by the specification.

Runtimes are encouraged to add support for all function types defined in the specification.

SDKs could also offer a level of extensibility for this attribute. I'll send a PR to Go and Java ones to support this approach.


tsurdilo:
@ricardozanini i played with this a little bit and made some changes :) Hope you don't mind.
Updated the text a little and also I think the option of string(enum)/object types is probably best to make it clear
that a type used is a custom type.

With my changes the type can either be string with the enum (supported by spec) values or object type where its clear that its a custom type:

          "type": {
            "custom": "mycustomtype"
          }

we can change that "custom" property name to whatever you see fit..maybe "name" but i wanted to emphasize that a custom object type is used so that parsers for example can see it is that.

Wdyt?

tsurdilo:
Another option that i think could work is to have two exclusive properties "type" and "customType" 
which would maybe even work better and be easier to deal with for SDKs...if you are ok with that let's go for it.

ricardozanini:
@tsurdilo, I agree that from an SDK/parser point of view, having the `object` type or even `customType` would be easier. Although from the user's perspective, I believe it's a little bit clunky. It feels way more natural having `type`, since it will be the same thing in the end.

If you strongly disagree, I would say to stick with `customType` or any other attribute name as a **string**, not object. Runtime implementations can be very creative if we allow any object. Additionally, the schema should validate these attributes as either `type` or `customType`, not both (I think we covered this already, right?).

Having `customType`, runtimes would have to add in their docs notes like "since this is not a type supported by the spec, use the field `customType` when declaring these functions." Which feels like they are transgressors. :smiling_imp:

schema/functions.json
```diff
@@ -40,17 +40,30 @@
...
-            "rest",
-            "asyncapi",
-            "rpc",
-            "graphql",
-            "odata",
-            "expression"
-          ],
-          "default": "rest"
+          "description": "Defines the function type. If string type, can be either `rest`, `asyncapi`, `rpc`, `graphql`, `odata`, `expression`. If object type it can define a custom type defined by specific runtime. Default value if not specified should be string type with value `rest`",
+          "oneOf": [
```
ricardozanini, line 44: The IDE IntelliSense understood this? We can still use the provided enums?

tsurdilo:
so we are ok with type/customtype? i feel bad for messing with your pr so will so all updates for this of its ok.

cdavernas:
I agree with @ricardozanini and do think that 2 distinct properties look and feel like a quick dirty fix. IMHO, just allowing any unsupported values in the `type` property is good enough. Runtime implementations could easily overlook - or not - any non-spec types

tsurdilo:
@cdavernas @ricardozanini i honestly rather put all the "customer" runtimes into the spec so we can promote it as having support for them than just putting anything in there user wants. 
having support for like drl, dmn, etc is nice to have as it promotes what the spec supports and we can even define it in the documentation as to what the format is of the "operation" property that way
issue with just type imo is that it forces our documentation of the "operation" prop to become irrelevant

tsurdilo:
i would like to enforce that in the drl that a "type" is custom and put in the docs that if you develop a custom type hey bring it into the spec via pr. i feel that just using "type" for everything promotes more of "do what you want we dont care" type of thing :)

tsurdilo:
@cdavernas we already kinda started doing that with the auth stuff you added, having a schema which is not directly linked in our main spec schemas but its there so users can follow it (and is included as part of the spec). 
I think this would be also nice for this -- if you want to support a "custom" type, make it part of the spec by adding a schema which defines how ppl can use it. reason is that even tho it could be specific to runtime impl, people still have to figure out the type value and operation prop format somehow.

ricardozanini:
Makes sense, @tsurdilo. But in this case, don't you think it's a bit weird to keep adding those types to the `enum`? We will indefinitely be adding types as the implementations grow. It should be their responsibility to properly communicate how to format their operations to users for these additional types. We can't afford to maintain all of these.

cdavernas:
@ricardozanini Agreed with your edit. @tsurdilo Your comments do make sense, TY ;)

tsurdilo:
@ricardozanini @cdavernas one more question:

why is it not possible for runtimes that really want to have custom stuff just use "expression" type functions?

you can in runtime check for specific formats like "ends in .drl" or "is java code"
and just handle it? then really custom expression types would not be required imo

ricardozanini:
@tsurdilo definitely can be an option as well. I'd rather have it in `type`, but it's doable. @cdavernas wdyt?

cdavernas:
It's an idea, for sure. Therefore we would only accept a dsl expression or custom file paths, right?

ricardozanini:
> It's an idea, for sure. Therefore we would only accept a dsl expression or custom file paths, right?

I think the type would be either a `string` or a `URI` reference.

ricardozanini:
@tsurdilo @cdavernas, let's settle this discussion in today's meeting. But I think we all agreed to support "external expressions" represented by a URI using function `type` `expression`.

ricardozanini:
@tsurdilo @cdavernas updated based on our conversation today

specification.md
```diff
@@ -1565,6 +1569,56 @@ For more information about functions, reference the [Functions definitions](#Fun
...
+
+Some runtime implementations might support additional function types that extend the ones
+defined in the specification. In those cases you can define a custom function type with for example:
+
+```json
+{
+"functions": [
+  {
+    "name": "sendOrderConfirmation",
+    "operation": "/path/to/my/script/order.ts#myOperation",
```
tsurdilo, line 1585: maybe here instead of "myOperation" should use "myFunction" im not typescript expert tho :)

tsurdilo:
@ricardozanini thanks! just left 1 small comment

specification.md
ricardozanini, line 1585:
```suggestion
    "operation": "/path/to/my/script/order.ts#myFunction",
```

PR #545, opened 2021-12-13
Simplified inject state, fixed jq filter for stateDataFilter section

mswiderski:

- [x] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

There are two changes
- `inject state` simplified to remove attributes that do not bring value such as timeouts, errors and state data filter as intended behaviour of `inject state` is to inject static content so there is literally no way it can cause failures for just some instances and by that error handling and timeouts do not apply. Similar for `stateDataFilter` singe the `data` is static there is no need to apply filtering as `data` should be described without filter in the way it is expected to be injected.
- fixed `jq` expression used in state data filter section

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -407,7 +407,7 @@ The first way would be to use both "input", and "output":
 {
   "stateDataFilter": {
     "input": "${ {vegetables: .vegetables} }",
-    "output": "${ {vegetables: .vegetables[] | select(.veggieLike == true)} }"
+    "output": "${ {vegetables: [.vegetables[] | select(.veggieLike == true)]} }"
   }
 }
 ```
@@ -424,7 +424,7 @@ The second way would be to directly filter only the "veggie like" vegetables wit
 ```json
 {
   "stateDataFilter": {
-    "input": "${ {vegetables: .vegetables[] | select(.veggieLike == true)} }"
+    "input": "${ {vegetables: [.vegetables[] | select(.veggieLike == true)]} }"
   }
 }
 ```
--- a/specification.md
+++ b/specification.md
@@ -2666,8 +2666,6 @@ reference the [Workflow Timeouts](#Workflow-Timeouts) section.
 | data | JSON object which can be set as state's data input and can be manipulated via filter | object | yes |
 | [stateDataFilter](#state-data-filters) | State data filter | object | no |
 | [transition](#Transitions) | Next transition of the workflow after injection has completed | object | yes (if end is set to false) |
-| [timeouts](#Workflow-Timeouts) | State specific timeout settings | object | no |
-| [onErrors](#Error-Definition) | States error handling and retries definitions | array | no |
 | [compensatedBy](#Workflow-Compensation) | Unique name of a workflow state which is responsible for compensation of this state | String | no |
 | [usedForCompensation](#Workflow-Compensation) | If true, this state is used to compensate another state. Default is "false" | boolean | no |
 | [metadata](#Workflow-Metadata) | Metadata information| object | no |
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -1471,16 +1471,6 @@
           "type": "object",
           "description": "JSON object which can be set as states data input and can be manipulated via filters"
         },
-        "timeouts": {
-          "type": "object",
-          "description": "State specific timeouts",
-          "properties": {
-            "stateExecTimeout": {
-              "$ref": "timeouts.json#/definitions/stateExecTimeout"
-            }
-          },
-          "required": []
-        },
         "stateDataFilter": {
           "description": "State data filter",
           "$ref": "#/definitions/statedatafilter"


cdavernas (approved):
LGTM! Many thanks!

specification.md
```diff
@@ -407,7 +407,7 @@ The first way would be to use both "input", and "output":
 {
   "stateDataFilter": {
     "input": "${ {vegetables: .vegetables} }",
-    "output": "${ {vegetables: .vegetables[] | select(.veggieLike == true)} }"
+    "output": "${ {vegetables: [.vegetables[] | select(.veggieLike == true)}] }"
```
tsurdilo, line 410: typo here, should be 
         {vegetables: [ .vegetables[] | select(.veggieLike == true)]}

specification.md
```diff
@@ -424,7 +424,7 @@ The second way would be to directly filter only the "veggie like" vegetables wit
 ```json
 {
   "stateDataFilter": {
-    "input": "${ {vegetables: .vegetables[] | select(.veggieLike == true)} }"
+    "input": "${ {vegetables: [.vegetables[] | select(.veggieLike == true)}] }"
```
tsurdilo, line 427: same typo here

tsurdilo:
+1 on removing timeouts. 
not a big fan of removing its state data filter

mswiderski:
@tsurdilo so what would be the use case for stateDataFilter? Since data is static, applying filter to it means not needed complexity- you should keep things simple and define data as it would be already after filter being applied.

specification.md
mswiderski: fixed

specification.md
mswiderski: fixed

cdavernas:
@tsurdilo I had the same perspective than you, but @mswiderski does make a great point. I'd tend to agree that it's useless in that state, isn't it?

tsurdilo:
Yeah i think complexity mostly it could help:

For example if we have:

            {
            "vegetables": [
              {
                "name": "potato",
                "amount": 2
              },
              {
                "name": "broccoli",
                "amount": 0
              },
              {
                "name": "celery",
                "amount": 5
              }
            ]
          }

and our inject state adds a "needForMeal" prop to each vegetable item:
        .{vegetables: .vegetables | map(. + {"needForMeal": true}) }

would make it simpler than

       {vegetables:  [ (.vegetables | map(. + {"needForMeal": true})  | .[] | select(.amount > 0) ) ] }

but also if you want to do some filtering regardless of the injected static data

mswiderski:
Then why not do this:

````
 {
        "vegetables": [
          {
            "name": "potato",
            "amount": 2,
            "needForMeal": true
          },
          {
            "name": "broccoli",
            "amount": 0 ,
            "needForMeal": true
          },
          {
            "name": "celery",
            "amount": 5 ,
            "needForMeal": true
          }
        ]
      }
````
I mean that data is static so I don’t see a point in splitting that into two things. 

Sorry but I really don’t see the use case that filter brings on static data that is embedded in workflow definition meaning won’t change between instances of that definition. Or am I missing something here?

tsurdilo:
oh ok, i forgot we removed that inject state "data" can no longer be an expression :( 

in that case ok the only reason i could argue to keep state data filter is in case user wants to filter the workflow data output on the last state in the workflow definition.

If user wants to filter the workflow data to make whats returned is in a format and contains only data that they want as result, they would need to apply it in state data filter before the inject state (if inject is last wf state)

tsurdilo:
@cdavernas i think it was a mistake to make "data" in inject state a json object and no longer an expression. makes inject state kinda useless imo

ricardozanini (approved):

mswiderski:
> @cdavernas i think it was a mistake to make "data" in inject state a json object and no longer an expression. makes inject state kinda useless imo

I wouldn't say that it makes it useless, rather specific in its responsibility. Executing expression to produce results that as well allow to do filtering - this is exactly what operations with expressions are for. Having inject doing the same wold be confusing in my opinion.

What I find useful about inject state is that is a great fit for fallback results. Let's say you invoke a rest call that return 404 so you can use inject to put in the data you would like to have for further processing. Or if you wait for an event with timeout, again better to have a data that clearly identifies the meaning rather than null etc.

tsurdilo:
If it's not too much of a problem, let's keep the state data filter for now and keep everything else as is in PR. 
Removing the useless timeout for inject i think is awesome but think we can/should address the state data filter later if that's ok please.

mswiderski:
Sure, will update the pr tomorrow morning

mswiderski:
@tsurdilo updated to bring back `stateDataFilter`

tsurdilo (approved):

cdavernas:
@tsurdilo In Synapse, I use the inject state's data as an expression-based object, meaning I'll check all its properties recursively to determine whether or not I should evaluate them.

PR #561, opened 2021-12-22
Sleep state changes

mswiderski:

- [x] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

Removes not relevant parts of the sleep state
- stateDataFilter - it's never used by the sleep state so no need to define it in the spec
- errors - not sure there are any errors that can be thrown from sleep state?
- timeouts - sounds really weird to set timeout on a sleep
- compensations - neither of `compensatedBy` or `usedForCompensation` makes sense on sleep state

This is a proposal based on review of the spec during runtime implementation and would like to take the opportunity to clean up anything that might impact users

### DIFF
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -700,33 +700,24 @@
         }
       },
       "additionalProperties": false,
-      "then": {
-        "required": [
-          "name",
-          "type",
-          "duration"
-        ]
-      },
-      "else": {
-        "oneOf": [
-          {
-            "required": [
-              "name",
-              "type",
-              "duration",
-              "end"
-            ]
-          },
-          {
-            "required": [
-              "name",
-              "type",
-              "duration",
-              "transition"
-            ]
-          }
-        ]
-      }
+      "oneOf": [
+        {
+          "required": [
+            "name",
+            "type",
+            "duration",
+            "end"
+          ]
+        },
+        {
+          "required": [
+            "name",
+            "type",
+            "duration",
+            "transition"
+          ]
+        }
+      ]
     },
     "eventstate": {
       "type": "object",
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -687,62 +687,19 @@
           "$ref": "#/definitions/end",
           "description": "State end definition"
         },
-        "stateDataFilter": {
-          "description": "State data filter",
-          "$ref": "#/definitions/statedatafilter"
-        },
         "duration": {
           "type": "string",
           "description": "Duration (ISO 8601 duration format) to sleep"
         },
-        "timeouts": {
-          "type": "object",
-          "description": "State specific timeouts",
-          "properties": {
-            "stateExecTimeout": {
-              "$ref": "timeouts.json#/definitions/stateExecTimeout"
-            }
-          },
-          "required": []
-        },
-        "onErrors": {
-          "type": "array",
-          "description": "States error handling definitions",
-          "items": {
-            "type": "object",
-            "$ref": "#/definitions/error"
-          },
-          "additionalItems": false
-        },
         "transition": {
           "description": "Next transition of the workflow after the workflow sleep",
           "$ref": "#/definitions/transition"
-        },
-        "compensatedBy": {
-          "type": "string",
-          "minLength": 1,
-          "description": "Unique Name of a workflow state which is responsible for compensation of this state"
-        },
-        "usedForCompensation": {
-          "type": "boolean",
-          "default": false,
-          "description": "If true, this state is used to compensate another state. Default is false"
         },
         "metadata": {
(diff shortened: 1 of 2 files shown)


tsurdilo:
> stateDataFilter - it's never used by the sleep state so no need to define it in the spec

+1 

> errors - not sure there are any errors that can be thrown from sleep state?
timeouts - sounds really weird to set timeout on a sleep

I think this in a way might imply that "sleep" is jus like a thread.sleep which isnt always imo the case.
application that runs the workflow might ask a "server/cluster" to create a timer for this sleep, and if that creation of a "durable" timer fails, there can be errors. So even tho I agree that it's not needed in most situations, for some runtimes it may be, so would rather keep it if possible.

> compensations - neither of compensatedBy or usedForCompensation makes sense on sleep state

not sure why? a "sleep 2 min"->"withdraw from account" sequence should have the ability to be compensated by "sleep 2 min"->"deposit to account"?

ricardozanini (approved):

mswiderski:
I’d say that creation of the state and its execution are two different things to be honest. Not sure it should be part of workflow definition error handling.  As what would you do if you can’t create sleep state in error handling? I’d say it should put workflow instance into error date as a whole than try to handle that as sort of recoverable error. 

When it comes to compensation you want to compensate withdraw and not the sleep itself so compensation should be on the other state. Sleep state does not define anything than a delay. So to do withdraw on account needs to be via another (usually operation) state.

tsurdilo:
ok but compensation is still part of orchestration logic. what if rule is to deposit only during business hours and have to sleep for some time

tsurdilo:
if we disable sleep state for compensation then we should also disable the action sleep (before/after) imo. otherwise think we can leave it in

mswiderski:
> ok but compensation is still part of orchestration logic. what if rule is to deposit only during business hours and have to sleep for some time before can do it

then this should be part of the compensation logic meaning it would be several states that would compose the composition

mswiderski:
> if we disable sleep state for compensation then we should also disable the action sleep (before/after) imo. otherwise think we can leave it in

what do you mean by action sleep?

tsurdilo:
https://github.com/serverlessworkflow/specification/blob/main/schema/workflow.json#L194

it does same thing as sleep state, but can be defined for each action

mswiderski:
thanks for the link, now I see these on action and there it is exactly what I would expect. It delays execution of certain action but does not have any error handling, timeouts or compensation attached to it. So I'd expect same to be for sleep state. Plus the case with durable timer creation failure sleep on action has same consequence.

> imo sleep state is useful when you want to perform something at exact time (fixes some shortcomings of cron for example) and having it part of compensation would not be bad.

the problem I see with compensating sleep state is that it would be like traveling back in time. as compensation is to undo/reverse the actions done by given state. And that cannot really happen, right?

>  for example how would you do something like "when triggering compensation for state X, wait Y hours before running sequence of steps A, B, C..."

I'd would do this with operation state that has action with sleep before, it is a perfect fit for that scenario imo

tsurdilo:
yeah it should work . only somewhat awkward as "wait time Y before" would be somewhat hidden in action of state A and could not be expressed  independently. might not be big deal tho idk

mswiderski:
I actually think it's event better as you have the logic for compensation in single place. I believe you could also have one state to transition to another state during compensation and then you could make it via sleep state and then operation state.

tsurdilo (approved):

schema/workflow.json
```diff
@@ -687,62 +687,19 @@
...
-          "type": "boolean",
-          "default": false,
-          "description": "If true, this state is used to compensate another state. Default is false"
-        },
         "metadata": {
           "$ref": "common.json#/definitions/metadata"
         }
       },
       "additionalProperties": false,
-      "if": {
```
tsurdilo, line 736: removed the "if" but not the "then" etc.
the section should just be:

        "additionalProperties": false,
        "oneOf": [
          {
            "required": [
              "name",
              "type",
              "duration",
              "end"
            ]
          },
          {
            "required": [
              "name",
              "type",
              "duration",
              "transition"
            ]
          }
        ]

tsurdilo:
@mswiderski found schema issue, see comment.

tsurdilo (changes requested):
issue with json schema def

schema/workflow.json
mswiderski: Updated, could you please double check if it is ok now?

mswiderski:
@tsurdilo found one small issue in the spec - duration attribute was set to integer why it should be string so made this update as well.

tsurdilo (approved):

PR #562, opened 2021-12-23
add name and namespace properties to extensions

tsurdilo:

- [x ] Specification
- [x ] Schema
- [ ] Examples
- [x ] Extensions
- [x ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

updates extensions definition to add "name" and "namespace" parameters. also adds default namespace value 
for extensions provided by this project. users can define different namespaces for their custom extensions.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -6171,7 +6171,7 @@ Some other examples of information that could be recorded in metadata are:
 The workflow extension mechanism allows you to enhance your model definitions with additional information useful for
 things like analytics, rate limiting, logging, simulation, debugging, tracing, etc.
 
-Model extensions do no influence control flow logic (workflow execution semantics).
+Extensions do no influence control flow logic (workflow execution semantics).
 They enhance it with extra information that can be consumed by runtime systems or tooling and
 evaluated with the end goal being overall workflow improvements in terms of time, cost, efficiency, etc.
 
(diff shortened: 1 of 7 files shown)


specification.md
```diff
@@ -6186,9 +6186,32 @@ You can define extensions in your workflow definition using its top-level `exten
...
+      "path": "file://myextensions/mykpi.yml"
+    }
+  ]
+}
+```
+
+Extension lookup can be performed by either specifying a URI to the extensions definition (JSON/YAML) by
+specifying it as the value of the `path` parameter.
+If the path parameter is not specified, the lookup should be done by `namespace` and `extensionId` properties.
+In this case, runtimes must provide means to map `namespace` and `extensionId` to an extension resource.
```
mswiderski, line 6214: Just wondering if we should not stick to just `extensionId` based lookup. That would make it way easier to figure out what is used and why. 

That would allow as well simple scheme for ids of extensions that could make it more portable and maybe even easier to share across runtimes. What do you think?

tsurdilo: Fine with me
@globalflea can you please add your input here as well? Your issue explicitly mentions namespace. Would the simplified lookup @mswiderski is proposing here work for you as well?

ricardozanini: If we stick with `extensionId` instead what's the purpose of this PR then? Because one can use the id as the namespace like `io.serverlessworkflow.extensions.kpi`

tsurdilo: I think that's what we are trying to figure out and need @globalflea input to see what his requirements are for this and how to best incorporate them. Please provide more info.

globalflea: Happy Holidays,
It depends on the intent of ExtensionId.  AFAIK, it functions like a pointer to the location of the extension element - i.e. it points to a specific extension within the file referred by Path parameter.  If this is the case, then it should not be used as for namespaces. 

On the other hand, if it not intended as a 'pointer', then by the same argument, i would say even either the Name or the ExtensionId is redundant - e.g Name can be of the  form "X.Y.ExtensionId.Name" or some sort of agreed convention like that

tsurdilo: thanks for the update. will think about it some more and update pr soon

tsurdilo: still looking into this...didnt forget :)

PR #578, opened 2022-01-24
Fix workflow id mention on parallel state branch def

ricardozanini:
There's no workflow ID in the branch definition any more :)

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -4064,7 +4064,8 @@ version: '2.0'
 
 </details>
 
-The `workflowId` property defined the unique ID of the sub-workflow to be invoked.
+The `workflowId` property define the unique ID of the sub-workflow to be invoked.
+The workflow id cannot be the same id of the workflow where the action is defined.
 
 The `version` property defined the unique version of the sub-workflow to be invoked.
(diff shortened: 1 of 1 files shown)


tsurdilo:
@ricardozanini thanks for PR!!!

prob need a discussion on this separately of PR, just wondering what you think about "the workflow id can't be the same id of the workflow where the branch is defined."

this really implies that a workflow definition cannot be both used as a "main" and a "subflow"....i wonder if we really need to have this restriction in place. do you mind starting a discussion on that and let's see where it goes :)

tsurdilo (approved):

ricardozanini:
> this really implies that a workflow definition cannot be both used as a "main" and a "subflow"

Hmm, I don't understand this way. As I understand, the branch I'm calling can't be the caller workflow. Maybe we need to clarify the text a little bit more. wdyt?

tsurdilo:
@ricardozanini 
yeah i think 

       The workflow id cannot be the same id of the workflow where the action is defined.

(and the other place where we mention this)
should probably lift this hard restriction and just say something like
"should in most cases not be the same id as the parent workflow id" 
wdyt?

tsurdilo (approved):

Issue #584, opened 2022-01-31
Clearly define the use for top-level "key" property


tsurdilo:
We added the top workflow-level "key" property for a specific use case where the workflow id might get "generated" by some CMS or something.
Because of this we kinda misuse it as something that identifies uniqueness of the workflow definition. In reality it's use is more on the workflow instance uniqueness level. 

Request here to either remove it or clearly state what its use is/should be.

mswiderski:
I agree that this is currently confusing in the spec but I'd like to keep it but repurpose it instead.

So `key` should not reference definition but reference instance of the workflow. So the `key` attribute would be an expression that could be evaluated upon start of the workflow based on the input of the workflow instance to assign alternative id e.g. order number, employee id etc.

That way we have `id` as the main identifier of the workflow definition and each instance of the definition could use `key` expression to assign domain specific identifier of the instance instead of or next to the one that is generated by runtime.

tsurdilo:
hows should that work when data for key comes from one or more events handled by starting event state? whats proposed  rule when you have both data input and are event-teiggered? is key immutable?

mswiderski:
key will be set only once upon instance creation and it is immutable, same as id is. It has exactly the same behavior as id - is set only once when new instance is created.

tsurdilo:
you mean the run/exec id, not the wf def id right? 
if so its pretty hard imo to define when its set without going into runtime details as in when wf instance is created. wdyt

tsurdilo:
without looking at runtime specific instance creation could it be set on starting wf state (with special instructions for event state)? if so can state handle expression error?

mswiderski:
I don't think there is a need to go into specifics of runtime and instance creation. I think it is enough to describe that it will be set when instance is created and that part can be specific to runtime but anyway each instance needs to be started at some point. Either by incoming events, manual trigger and so on.

mswiderski:
`key` expression failures should be ignored and simply key is not set when expression fails.

cdavernas:
Completely agree with @mswiderski ! Implemented it already and it just rules! Great, great idea, especially for user Quality Of Life

PR #598, opened 2022-02-14
Turn the `key` top-level property into a workflow instance identifier expression

cdavernas:

- [x] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [x] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

Following @mswiderski AWESOME suggestion in https://github.com/serverlessworkflow/specification/issues/584, change the `key` property into an expression used by runtimes to determine instances unique identifiers, based on input data. 

Beforehand, the property had, IMO, no reason to exist, and it's only added value was to confuse workflow authors by granting them two ways of identifying a workflow definition.

The `key` property has been made optional, thus making the the `id` property always required


ricardozanini (approved):

specification.md
```diff
@@ -1903,10 +1903,30 @@ Following figure describes the main workflow definition blocks.
...
+
+```yaml
+payment:
+  total: 53.21
+  currency: EUR
+  type: card
+  orderId: 1234
+```
+
+... the resulting workflow instance would get the '1234' unique identifier.
```
ricardozanini, line 1929: @cdavernas can you also please incorporate the other comments?

- If the expression evaluated fails, the workflow should ignore it and continue the execution (how this is passed to the end user is an implementation detail) (https://github.com/serverlessworkflow/specification/issues/584#issuecomment-1026777370)
- The key is immutable once the workflow is created (https://github.com/serverlessworkflow/specification/issues/584#issuecomment-1026752612) and (https://github.com/serverlessworkflow/specification/issues/584#issuecomment-1026776835)

cdavernas: Done! @ricardozanini Is it good enough for you?

cdavernas:
@mswiderski WDYT?

I personally think something could be done with start event payload, too. However, the problem is that we would probably have to wait for the start event state to complete and select/merge payload back into state data, at that point usable by the expression. 

Whatever the outcome - and I'm ok to put it in metadata, too - I IMHO think that not answering all possible use cases is not a reason big enough to exclude a somewhat cool, quality of life improvement (for customer facing APIs that can expose human-readable, human-friendly paths).

WDYT?

tsurdilo:
@cdavernas 

> Whatever the outcome - and I'm ok to put it in metadata, too - I IMHO think that not answering all possible use cases is not a reason big enough to exclude a somewhat cool, quality of life improvement (for customer facing APIs that can expose human-readable, human-friendly paths).

i agree with this fully but the problem from the spec side is that:
1. we need to clearly state that it's only useful if you have workflow data input (can be added to this pr)
2. we will get asked to provide same functionality for event-triggered instances

the rules on when "key" needs to be evaluated on starting event state are just too custom imo to justify it. wdyt?

mswiderski:
> so key is only usable if you have wf data input?
it seems never usable for event-driven workflow triggers

not really, it's an expression so it can reference input data but can also generate ids without input data too, like sequence numbers, timestamp based keys etc.

mswiderski:
> @mswiderski WDYT?
> 
> I personally think something could be done with start event payload, too. However, the problem is that we would probably have to wait for the start event state to complete and select/merge payload back into state data, at that point usable by the expression.
> 
> Whatever the outcome - and I'm ok to put it in metadata, too - I IMHO think that not answering all possible use cases is not a reason big enough to exclude a somewhat cool, quality of life improvement (for customer facing APIs that can expose human-readable, human-friendly paths).
> 
> WDYT?

I agree that it has very big value and in my option this is really dependant on when the instance is started. If the first event that matches criteria is received it triggers new workflow instance to be created and as such the event payload is considered the input data of it. By that `key` can be evaluated against that data and still provide valuable outcome.

tsurdilo:
needs a rebase + adding that text we talked about in [this comment](https://github.com/serverlessworkflow/specification/pull/598#issuecomment-1039337291)

Issue #612, opened 2022-03-31
Authentication for REST requests


EnriqueL8:
**What would you like to be added**:
At the moment there is no notion of authentication within the spec when using a `rest` state. Let's assume we have a function such as 
```
functions:
- name: secureAPI
  operation: "openapi.yaml#createPets"
```

The OpenAPI file might contain a [security scheme](https://swagger.io/docs/specification/authentication/) such as Bearer Token. So how do I describe in my SW that I need to provide a bearer token for this request? One of the solutions might be to pass it in the argument for the `functionRef` but this isn't ideal since you would have to add it every time you want to use that function.

The best solution would be to have a new construct within the spec where you can define the different types of authentication for the functions being used. We can use the existing [auth]( https://github.com/serverlessworkflow/specification/blob/main/specification.md#Auth-Definition)  used to access the resource containing the function invocation information but in this case, it would be authentication information for the function invocation. Not saying we should use the same contract, but a have similar one for function invocation.

**Why is this needed**:
In order to describe the authentication needed a function invocation.

cdavernas:
@EnriqueL8 you can use the top level `auth` property. It currently supports basic, oauth2 and bearer. I'd recommend using a secret-based OAuth2 auth definition so that you do not expose any sensitive info in your flow.

EnriqueL8:
As far as I understood, the `authRef` is used to access the resource containing the function invocation information but not  for authentication information at function invocation

cdavernas:
The `authRef` is used to reference an authentication mechanism declared at top level. The referenced scheme and its properties should be implemented and handled by the runtime when calling the function. 
For example, if you referenced a basic auth definition, the runtime should, when calling an OpenApi function, add the appropriate Authorization header with the specified username and password. 
Does that answer your question or am I missing smtg or failling to understand your use case?

cdavernas:
Well, but if you don't have an intermediate human task or a mechanism of the like, what would be the difference between pre-defining the username/password in an `authDef` or directly within the function?
You don't necessarily have to use a secret btw, you very well could do something like the following:

```
...
auth:
  - name: basic
    scheme: basic
    properties:
    - username: serverlessworkflow
      password: rules
...
```

I would, however, advise against using such a scheme (basic or bearer), and would rather advise using oauth2. As a matter of fact, using the oauth2 scheme, the runtime would have to "produce" the token by contacting the configured `authority`, using the defined `client_id`, `client_secret`, etc, which is obviously way safer.

[...]

PR #613, opened 2022-03-31
Clarification on usage of authDef on function definitions

ricardozanini:
Signed-off-by: Ricardo Zanini <zanini@redhat.com>

- [x] Specification
- [ ] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

There's a small lack of clarity in the usage of the `authDef` in the function defintions.

Please see #612 for more context.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -2045,7 +2045,7 @@ If we have the following function definition:
 ```
 
 The `authRef` property is used to reference an authentication definition in
-the `auth` property and should be applied to access the `https://secure.resources.com/myapi.json`
+the workflow `auth` property. It can also be used to access the `https://secure.resources.com/myapi.json`
 OpenApi definition file.
 
 The `functions` property can be either an in-line [function](#Function-Definition) definition array, or an URI reference to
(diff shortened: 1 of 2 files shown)


specification.md
```diff
@@ -3539,10 +3539,10 @@ If `false`, both Event payload and context attributes should be accessible.
 
 ##### Auth Definition
 
-Auth definitions can be used to define authentication information that should be applied
+Auth definitions can be used to define authentication information that is applied
 to resources defined in the operation property of [function definitions](#Function-Definition).
-It is not used as authentication information for the function invocation, but just to access
-the resource containing the function invocation information.
+
+If defined, it will replace the [security scheme](https://spec.openapis.org/oas/v3.1.0#security-scheme-object) described in the OpenAPI file.
```
EnriqueL8, line 3545: What do we mean by replacing the security scheme? The OpenAPI file describes the contract of a particular service, how can we replace this? Do you mean that instead of using the security scheme defined by the OpenAPI to call the service it would use the one defined in this section?

ricardozanini: That's the original desired behavior, yes. But based on [this conversation](https://github.com/serverlessworkflow/specification/issues/612#issuecomment-1085051096), we are likely to restructure the way about auth definitions, so you could define auth info for fetching the spec resource and another to replace the original security scheme in the spec if desired.

cdavernas (approved):
LGTM! Many thanks!

specification.md
EnriqueL8: Okay, but what do you mean by the `original security scheme`? The one in the OpenAPI? Or the one defined to fetch the spec? I think there is some confusion here over what is overriding what if someone reads the spec without the context of that issue.

ricardozanini: The on in the Open API. This text is linked to the Open API spec doc. This will be enlightened as soon as we add the property to separate the resource vs invocation auth to the spec.

cdavernas:
@tsurdilo @antmendoza  Can you review so that we can proceed with merging this PR?

antmendoza (approved):
👍

cdavernas:
@ricardozanini Could you resolve conflicts? I'll merge it ASAP thereafter!

ricardozanini:
@cdavernas I'm gonna close this PR since now we have it settled in yours about this usage. After rebase, there's nothing left :)

PR #617, opened 2022-04-13
Removed optional fields from the list of required fields in schema files.

paulovmr:

- [ ] Specification
- [X] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

It removes some fields from the list of required fields in the schema files as per specification basis.

### DIFF
--- a/schema/auth.json
+++ b/schema/auth.json
@@ -22,9 +22,6 @@
       }
     ]
   },
-  "required": [
-    "auth"
-  ],
   "definitions": {
     "authdef": {
       "type": "object",
--- a/schema/errors.json
+++ b/schema/errors.json
@@ -22,9 +22,6 @@
       }
     ]
   },
-  "required": [
-    "errors"
-  ],
   "definitions": {
     "errordef": {
       "type": "object",
--- a/schema/events.json
+++ b/schema/events.json
@@ -22,9 +22,6 @@
       }
     ]
   },
-  "required": [
-    "events"
-  ],
   "definitions": {
     "eventdef": {
       "type": "object",
--- a/schema/functions.json
+++ b/schema/functions.json
@@ -22,9 +22,6 @@
       }
     ]
   },
-  "required": [
-    "functions"
-  ],
   "definitions": {
     "function": {
       "type": "object",
--- a/schema/retries.json
+++ b/schema/retries.json
@@ -22,9 +22,6 @@
       }
     ]
   },
-  "required": [
-    "retries"
-  ],
   "definitions": {
     "retrydef": {
       "type": "object",
--- a/schema/secrets.json
+++ b/schema/secrets.json
@@ -19,8 +19,5 @@
         "minItems": 1
       }
     ]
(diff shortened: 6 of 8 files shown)


ricardozanini (approved):

tsurdilo:
Not sure what this is fixing, could you explain?
These required fields should be local to the schema. For main workflow schema the required fields are for example set here: https://github.com/serverlessworkflow/specification/blob/main/schema/workflow.json#L185

paulovmr:
@tsurdilo In my specific case, we are trying to use all those schemas to validate workflow files. If those fields are required, then the files will have several "Missing property X" errors/warnings relating to those optional properties.

tsurdilo:
@paulovmr 
that's weird, which tool(s) are you using? This is not happening in VSCode when validating against the main workflow scheme or on our editor https://serverlessworkflow.io/editor that has those files as well.

cdavernas:
@tsurdilo Yeah I'm not sure I understand either, I'm using the same files too without any problem/warning.

paulovmr:
@tsurdilo We are implementing a validation that checks against all schemas (not only the main one), not using one existing tool. 

If we use, let's say, the `schema/events.json` besides the main workflow schema to validate a workflow file that doesn't have a `events` property, which is valid according to the specification, we will have a error/warning, as you would expect since the events schema contains the property `events` as required. Is there a reason why they should be required even in their specific schemas?

cdavernas:
@paulovmr I think you have some kind of validation problem with your tool then. The SW validation tool uses external schema files, too, just check requests that are made when using it. 
I'm using Newtonsoft and NJsonSchema to do the validation using external files, and I'm not having any warning poping up.

tsurdilo:
@paulovmr @cdavernas 

> We are implementing a validation that checks against all schemas (not only the main one),

For your workflow definition validation you should only validate against the main workflow schema.

The individual schemas you are requesting to change should be used in cases where you have an external json definition for them, for example in your workflow definition:

       "timeouts": "mytimeouts.json"

you would validate mytimeouts.json against the timeouts schema. 
At least this is how we tried to design these schemas.

paulovmr:
@tsurdilo @cdavernas You are right, I checked now the main workflow schema and there are references to the others in the specific properties, so they only will be applied when the property is present, makes sense. The main one should be indeed enough. I will close this PR, thanks for the clarification!

tsurdilo:
Hey @paulovmr did you need this stuff to work on "your" vscode extension, or "your" diagram generation: https://github.com/kiegroup/kie-tools/tree/main/packages/serverless-workflow-editor/src/diagram?

We do depend on our community support and encourage it, but we encourage the real open source way and not what seems to be going on here.

tsurdilo:
@paulovmr did you think of paying our ts sdk maintainer at least a little respect when you copied the diagram code? i guess not.....shady.

PR #629, opened 2022-05-12
Changed the `$CONST` keyword into `$CONSTANTS` for ubiquity purpose

cdavernas:

- [x] Specification
- [ ] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

Changes the `$CONST` keyword into `$CONSTANTS` for ubiquity purpose

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -4649,12 +4649,12 @@ decide which activity to perform based on the transaction value.
     {
      "name": "Process Larger Transaction",
      "functionRef": "Banking Service - Larger Tx",
-     "condition": "${ .tx >= $CONST.largetxamount }"
+     "condition": "${ .tx >= $CONSTANTS.largetxamount }"
     },
     {
      "name": "Process Smaller Transaction",
      "functionRef": "Banking Service - Smaller Tx",
-     "condition": "${ .tx < $CONST.largetxamount }"
+     "condition": "${ .tx < $CONSTANTS.largetxamount }"
     }
    ],
    "end": true
@@ -4694,10 +4694,10 @@ states:
    actions:
     - name: Process Larger Transaction
       functionRef: Banking Service - Larger Tx
-      condition: "${ .tx >= $CONST.largetxamount }"
+      condition: "${ .tx >= $CONSTANTS.largetxamount }"
     - name: Process Smaller Transaction
       functionRef: Banking Service - Smaller Tx
-      condition: "${ .tx < $CONST.largetxamount }"
+      condition: "${ .tx < $CONSTANTS.largetxamount }"
    end: true
 functions:
  - name: Banking Service - Larger Tx
--- a/specification.md
+++ b/specification.md
@@ -5963,8 +5963,8 @@ for example:
 }
 ```
 
-Constants can only be accessed inside Workflow expressions via the `$CONST` variable.
-Runtimes must make `$CONST` available to expressions as a predefined variable.
+Constants can only be accessed inside Workflow expressions via the `$CONSTANTS` variable.
+Runtimes must make `$CONSTANTS` available to expressions as a predefined variable.
 
 Here is an example of using constants in Workflow expressions:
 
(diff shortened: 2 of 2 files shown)


ricardozanini (approved):

ricardozanini:
That's fine, but it will break compatibility with the previous versions.

fjtirado (commented):
Why are we introducing a breaking compatibility change just to introduce a longer reserved word?
This will force implementors to change all their files using constants
Const is shorter and as easy to read as constants. 
So, whats the reason for that change?

cdavernas:
@ricardozanini indeed, that's why I wanted to squeeze it in before 0.9 which, as far as I'm aware of, not gave any of thoses

cdavernas:
@fjtirado Replace all and that's done. Not such a huge compatibility problem imho. Plus, I don't see why you couldn't support both keywords before 1.0 if you don't want to go down that path.

The need for that change imo is ubiquity. The DSL is to be used by non-technical profiles too, and using abbreviations is not recommended, not even in dev, in your code. Ubiquity guarantees intuitive understanding of what you are looking at. Otherwise why not rename WORKFLOW to WF or SECRETS to SEC too then, if shortness is the goal. CONSTANTS is just one letter longer than WORKFLOW. Hell, we could even use CONSTANT if you prefer.

ricardozanini:
To be fair, `const` is a very well-known abbreviation for `constants`.

cdavernas:
@ricardozanini I agree totally, not saying it's obscure terminology or whatever, but is less reader friendly imo, and less ubiquitous: it could stand for constraints too, for instance

cdavernas:
The other terms we have been using are on point, perfectly explicit. Not that one, imo.

fjtirado:
@cdavernas 
Implementing $CONST in jsonpath properly was not straighforward. Although that is a proper language construct for a variable in jq, the proper language construct for jsonpath will be @.CONSTANT (the same applies for secret and workflow)
If you really want to have a valid expression without doing hacky string replamencent before evaluating it, I would argue that the spec should just define the magic word (WORKFLOW, CONST/CONSTANT, SECRET) and states that this should be variables of the expression. 
And still not able to see whats the benefit of changing that, const is intuitive shortcut for constants, as C++ language demostrated
Also, another complaint I have is why there was not issue to discuss this change before actually opening the PR

cdavernas:
> Also, another complaint is of procedure: why there was no issue to discuss this breaking compatibility change before actually opening the PR?

@fjtirado We have been doing PR, then discussing on them for a while now, but ok. I don't think we have established any "procedure" as you say, especially if said PR concerns a 1mn "Replace All" kind of work. 
I guess, however, that you are right: we should probably start with a GitHub conversation rather than with a PR.

> Implementing $CONST in jsonpath properly was not straighforward

What do you mean by JsonPath? Are you using it for expression evaluation?
With  `jq` - which is the default expression language-, you can pass json arguments like, say: `--argjson CONST {CONSTJSON}` so in my use case it's an easy fix, really.

> And still not able to see what's the benefit of changing that, const is intuitive shortcut for constants, as C++ language demonstrated

C++, C#, whatever, I agree with you, as I said to @ricardozanini, I'm not contradicting the fact that in our world (tech one, it is), const is a perfectly ubiquitous terminology; I'm just saying it might not be the case for a person working in, say business, and wants to use SW to orchestrate high level business tools. Furthermore, I believe that the IMO reasonable idea that a concept such as "constraints" might come into play at some point in time (spec or runtime side), thus becoming confusing.

Please understand I've no other interest than ubiquity here, it implies as well a (minor) change on Synapse, for example. So if you think we should not do that, and that's the consensus, then I'll close the PR 😉 

But I'd like to hear what @antmendoza and @tsurdilo think about it before taking it one way or the other (or another 😜)

fjtirado:
> Oh ok! Yeah the jq vs jsonpath vs ... is yet another discussion, but yeah, of course, any and all implementations are/should be allowed. I really never intended to force the use of the $ prefix or anything like that.

Implementation wise to support constant rather than const just implies changing the value of a constant (no pump intended ;))
but to support both, implies changing the expression evaluator to add a new constant that is resolved the same way than the previous one. This implies at least three classes (the two languages expression implementations plus the one holding the constant) rather than one. Anyway, not a big deal, so if this goes forward, Ill probably do that to avoid changing all examples that currently use const. 
But I think we need to define a procedure that, for every breaking compatibility change, albeit minimum one, we should open an issue to discuss first. In this particular case, it is just a question of taste, I prefer the existing word than the proposed one (const/constant refers to the same concept, there is not ambiguity when using the shorted one). And I also  feel that when changing something that is already there, it needs to be a good reason to do that (the existing one is either wrong, ambigous, error promt or any other reason I cannot foresee in this case).

cdavernas:
@fjtirado @tsurdilo @ricardozanini Should we proceed with that PR, or just drop it alltogether?

cdavernas:
@tsurdilo Not urgent at all, no. But I'd be happier tackling this, if ever, sooner than later, as it is a breaking change

ricardozanini:
I think we are missing a governance model to handle this kind of situation, like voting in a (sort of) controversial feature.

cdavernas:
Agreed, let's then vote on that one on next meeting. 

Created #644 to address future situations

PR #631, opened 2022-05-13
Changed "operation" to be required in Function Definitions

hbelmiro:
- [x] Specification
- [ ] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

`operation` not being required in Function definitions seems to be a bug. The web editor complains if we don't define an operation in a function definition.

Screenshot from 2022-05-13 12-01-00

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -3192,7 +3192,7 @@ Note that `transition` and `end` properties are mutually exclusive, meaning that
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
 | name | Unique function name | string | yes |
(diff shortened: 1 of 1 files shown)


ricardozanini:
If the editor is complaining, then I understand the scheme is correct, the only missing issue is in the description?

cdavernas (approved):

hbelmiro:
> If the editor is complaining, then I understand the scheme is correct, the only missing issue is in the description?

Yes. The scheme is correct.
https://github.com/serverlessworkflow/specification/blob/da216bf69700d5ef0bcfd441bb6bd46285226b02/schema/functions.json#L40

PR #633, opened 2022-05-24
Fix various issues in the specification and schema

MarianMacik:

- [x] Specification
- [x] Schema (just description of properties)
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

While getting familiar with the spec I found a couple of discrepancies, old wordings and typos.

* Fixed data filters example
* Fixed links to Examples section
* ~Made wording abour produced/consumed events more understandable~
* requestedSubject description fix
* maxAttempts must be a positive integer (spec allowed 0 while schema requires a positive integer), ~fixed also examples that number of 1 means retry once, not zero times~
* schedule can be also a string, not just an object
* ~spec doesn't allow `2020-03-20T09:00:00Z/2020-03-20T15:00:00Z` ISO8601 format - fixed examples with `2020-03-20T09:00:00Z/PT2H`~ - will be resolved in version 0.9 of the spec
* compensatedBy description updated
* Fixed compensation example steps to better reflect the example image definition

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -674,7 +674,7 @@ a workflow with a single event state and show how data filters can be combined.
             }],
             "stateDataFilter": {
                 "input": "${ .greetings } ",
-                "output": "${ .finalCustomerGreeting }"
+                "output": "${ { finalCustomerGreeting } }"
             },
             "end": true
         }
@@ -1785,7 +1785,7 @@ definition "id" must be a constant value.
 | autoRetries | If set to true, [actions](#Action-Definition) should automatically be retried on unchecked errors. Default is false | boolean| no |
 | [retries](#Retry-Definition) | Workflow retries definitions. Can be either inline retries definitions (if array) or URI pointing to a resource containing json/yaml retry definitions (if string) | array or string| no |
 | [states](#Workflow-States) | Workflow states | array | yes |
-| [extensions](#Workflow-Extensions) | Workflow extensions definitions | array or string | no |
+| [extensions](#Extensions) | Workflow extensions definitions | array or string | no |
 | [metadata](#Workflow-Metadata) | Metadata information | object | no |
 
 <details><summary><strong>Click to view example definition</strong></summary>
@@ -3342,7 +3342,7 @@ that triggers workflow instance creation, or continuation of workflow instance e
 that the workflow instance creates during its execution (produced).
 The default value (if not specified) of the `kind` property is `consumed`.
 Note that for `produced` event definitions, implementations must provide the value of the CloudEvent source attribute.
-In this case (i.e., when the `kind` property is set to `produced`), the `source` property of the event definition is not required.
+In this case (i.e., when the `kind` property is set to `produced`), the `source` property of the event is not required in the workflow definition.
 Otherwise, (i.e., when the `kind` property is set to `consumed`), the `source` property must be defined in the event definition.
 
 Event correlation plays a big role in large event-driven applications. Correlating one or more events with a particular workflow instance
@@ -3539,7 +3539,7 @@ See [here](https://oauth.net/2/) for more information about OAuth2 Authenticatio
 | password | String or a workflow expression. Contains the user password. Used only if grantType is 'resourceOwner' | string | no |
 | audiences | Array containing strings or workflow expressions. Contains the OAuth2 audiences | array | no |
 | subjectToken | String or a workflow expression. Contains the subject token | string | no |
-| requestedSubject | String or a workflow expression. Contains the client identifier | string | no |
+| requestedSubject | String or a workflow expression. Contains the requested subject | string | no |
 | requestedIssuer | String or a workflow expression. Contains the requested issuer | string | no |
 | [metadata](#Workflow-Metadata) | Metadata information| object | no |
 
@@ -3594,7 +3594,7 @@ correlation:
 Used to define event correlation rules. Only usable for `consumed` event definitions.
 
 The `contextAttributeName` property defines the name of the CloudEvent [extension context attribute](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#extension-context-attributes).
-The `contextAttributeValue` property defines the value of the defined the CloudEvent [extension context attribute](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#extension-context-attributes).
+The `contextAttributeValue` property defines the value of the defined CloudEvent [extension context attribute](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#extension-context-attributes).
 
 ##### OnEvents Definition
 
@@ -3704,7 +3704,7 @@ This is visualized in the diagram below:
 | --- | --- | --- | --- |
 | name | Unique Action name | string | no |
 | [functionRef](#FunctionRef-Definition) | References a reusable function definition | object | yes if `eventRef` & `subFlowRef` are not defined |
-| [eventRef](#EventRef-Definition) | References a `trigger` and `result` reusable event definitions | object | yes if `functionRef` & `subFlowRef` are not defined |
+| [eventRef](#EventRef-Definition) | References a `produce` and `consume` reusable event definitions | object | yes if `functionRef` & `subFlowRef` are not defined |
 | [subFlowRef](#SubFlowRef-Definition) | References a workflow to be invoked | object or string | yes if `eventRef` & `functionRef` are not defined |
 | [retryRef](#retry-definition) | References a defined workflow retry definition. If not defined uses the default runtime retry definition | string | no |
 | nonRetryableErrors | List of references to defined [workflow errors](#Defining Errors) for which the action should not be retried. Used only when `autoRetries` is set to `true` | array | no |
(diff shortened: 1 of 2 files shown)


MarianMacik:
@ricardozanini FYI

ricardozanini:
@MarianMacik can you please sign the DCO and your commits?

specification.md
```diff
@@ -4111,7 +4110,7 @@ For more information, see the [Workflow Error Handling](#Workflow-Error-Handling
 | --- | --- | --- | --- |
 | name | Unique retry strategy name | string | yes |
 | delay | Time delay between retry attempts (ISO 8601 duration format) | string | no |
-| maxAttempts | Maximum number of retry attempts. Value of 0 means no retries are performed | string or number | no |
+| maxAttempts | Maximum number of retry attempts. A positive integer value | string or number | no |
```
tsurdilo, line 4113: The issue here is "0", should be "1". 
Setting maxAttempts to 1 means no retries.

specification.md
```diff
@@ -5249,7 +5248,7 @@ To start, let's define a workflow top-level `retries` definition:
     "maxAttempts": 10
   },
   {
-    "name": "DoNotRetryStrategy",
+    "name": "JustOnceRetryStrategy",
```
tsurdilo, line 5251: not needed. please revert

specification.md
```diff
@@ -5267,7 +5266,7 @@ retries:
   - name: SecondRetryStrategy
     delay: PT10M
     maxAttempts: 10
-  - name: DoNotRetryStrategy
+  - name: JustOnceRetryStrategy
```
tsurdilo, line 5269: not needed please revert

specification.md
```diff
@@ -5304,7 +5303,7 @@ Our retry definitions can be referenced by state actions. For example:
     },
     {
       "functionRef": "MyFourthFunction",
-      "retryRef": "DoNotRetryStrategy"
+      "retryRef": "JustOnceRetryStrategy"
```
tsurdilo, line 5306: not needed please revert

specification.md
```diff
@@ -5327,7 +5326,7 @@ actions:
       - SomeErrorThree
   - functionRef: MyThirdFunction
   - functionRef: MyFourthFunction
-    retryRef: DoNotRetryStrategy
+    retryRef: JustOnceRetryStrategy
```
tsurdilo, line 5329: not needed please revert

specification.md
MarianMacik: @tsurdilo I understand the number of retry attempts as a number of times we actually retry. So 1 means we will retry once if the invocation originally failed. Or am I missing something?

specification.md
```diff
@@ -674,7 +674,7 @@ a workflow with a single event state and show how data filters can be combined.
             }],
             "stateDataFilter": {
                 "input": "${ .greetings } ",
-                "output": "${ .finalCustomerGreeting }"
+                "output": "${ finalCustomerGreeting: .finalCustomerGreeting }"
```
tsurdilo, line 677: I don't think this is correct. 
The action data filter already pushes to "finalCustomerGreeting"
and spec says "The "toStateData" expression says that we should add/merge this selected event data to the state data ... If this property exists it should be merged, if it does not exist, one should be created."

I believe what is there now is correct.

specification.md
```diff
@@ -3539,7 +3538,7 @@ See [here](https://oauth.net/2/) for more information about OAuth2 Authenticatio
 | password | String or a workflow expression. Contains the user password. Used only if grantType is 'resourceOwner' | string | no |
 | audiences | Array containing strings or workflow expressions. Contains the OAuth2 audiences | array | no |
 | subjectToken | String or a workflow expression. Contains the subject token | string | no |
-| requestedSubject | String or a workflow expression. Contains the client identifier | string | no |
+| requestedSubject | String or a workflow expression. Contains the requested subject | string | no |
```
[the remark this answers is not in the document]
cdavernas: @tsurdilo The fix is correct, it should indeed be `requested subject`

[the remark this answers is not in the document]
tsurdilo: Thanks, let's then also change it [here](https://github.com/serverlessworkflow/specification/blob/main/schema/auth.json#L179) @MarianMacik  please

specification.md
```diff
@@ -4537,7 +4536,7 @@ If the start definition is of type `object`, it has the following structure:
 ```json
 {
   "stateName": "MyStartingstate",
-  "schedule": "2020-03-20T09:00:00Z/2020-03-20T15:00:00Z"
+  "schedule": "2020-03-20T09:00:00Z/PT2H"
```
tsurdilo, line 4539: Can you explain why this is needed? This is  mean to be ISO_8601
<start>/<end> time duration that tells runtimes the start and end times when instances should be created.

Maybe if you wanted we could do a separate sample that does <start>/<duration>, that would be ok too if you wanted.

specification.md
MarianMacik: @tsurdilo then I suggest to name it `maxInvocations` or similar before v1.0 of spec as otherwise it is quite misleading.

specification.md
```diff
@@ -4546,7 +4545,7 @@ If the start definition is of type `object`, it has the following structure:
 
 ```yaml
 stateName: MyStartingstate
-schedule: 2020-03-20T09:00:00Z/2020-03-20T15:00:00Z
+schedule: 2020-03-20T09:00:00Z/PT2H
```
tsurdilo, line 4548: same here, it's just showing start/end time interval for this workflow sample

tsurdilo:
@MarianMacik thanks for the PR, left some small comments.

specification.md
tsurdilo: From the retries perspective, first invocation of an action is an attempt.

specification.md
tsurdilo: Ok nice catch,  if its not there then let's add it (and maybe say that its primary use is within schedule property).

specification.md
MarianMacik: From the spec:

> The `maxAttempts` property determines the maximum number of retry attempts allowed and is a positive integer value.

But I got you. I will change it to "maximum number of attempts allowed..." so it is more clear and will handle 1 correctly, wdyt?

specification.md
```diff
@@ -3341,8 +3341,7 @@ The `kind` property defines this event as either `consumed` or `produced`. In te
 that triggers workflow instance creation, or continuation of workflow instance execution (consumed), or an event
 that the workflow instance creates during its execution (produced).
 The default value (if not specified) of the `kind` property is `consumed`.
-Note that for `produced` event definitions, implementations must provide the value of the CloudEvent source attribute.
-In this case (i.e., when the `kind` property is set to `produced`), the `source` property of the event definition is not required.
+Note that for `produced` event definitions (i.e., when the `kind` property is set to `produced`), the `source` property of the event definition is not required.
```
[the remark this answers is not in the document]
MarianMacik: It seems I missed the difference here:
Note that for `produced` event definitions, implementations must provide the value of the CloudEvent source attribute. - This is the source attribute of the CE message.
In this case (i.e., when the `kind` property is set to `produced`), the `source` property of the event definition is not required. - This is the source property in the event definition inside the SWF definition file.

Originally I thought it was telling the same twice. Now it makes sense.

specification.md
tsurdilo: I think all we need to say is "setting this value to 1 disables retries." Maybe that would be the clearest?

specification.md
tsurdilo: start/end is the time interval 
Rn / start / end would be repeating interval

I would change all "R/...." with "Rn/...." to show that repeating is optional

specification.md
tsurdilo: Yes! ^^ do that

ricardozanini: This is related to #632

specification.md
ricardozanini, line 3344:
```suggestion
Note that for `produced` event definitions (i.e., when the `kind` property is set to `produced`), the `source` property of the event is not required in the workflow definition. However, implementations must provide this value in runtime.
```

What about this?

ricardozanini (commented):
Thank you! One small comment only. LGTM as soon as we handle all the suggestions.

specification.md
MarianMacik: But again, the spec already says that

> Unlike full ISO 8601, this derivative format does not allow expression of an explicit number of recurrences

So basically Rn, e.g. R5/...., is not supported - so we cannot specify bounded interval.
Do you want me to rewrite it also for Rn?

specification.md
MarianMacik: @ricardozanini It is, but Tristan found another issue, which is complementary to mine.

specification.md
MarianMacik: Fixed.

specification.md
MarianMacik: Fixed.

specification.md
MarianMacik: Fixed.

specification.md
MarianMacik: Fixed.

specification.md
MarianMacik: Fixed.

specification.md
MarianMacik: Fixed.

specification.md
MarianMacik: Added start/end use case. Do you also want me to add R[n] to the spec? Then the ISO 8601 spec says:
"Leaving out the value for [n] or specifying a value of -1, means an unbounded number of repetitions. A value of 0 for [n] means the interval is not repeated. "
Then it is unclear to me how we should interpret it. If R0 means interval is not repeated, so we have just one interval - then we should start one workflow at its beginning? With R1 we will repeat interval once, so there will be 2 intervals specified - and at each interval's start we will execute a workflow, so there will be 2 workflows triggered?

Then for `R[n]/<Duration>/<End>` we will execute it n+1 times, last one executed "duration" before "end"?

MarianMacik: @tsurdilo I think we can at least leave start/end use case as it is straight-forward, just let me know what you think about it.

MarianMacik:
I think I addressed everything, let me know what else should be done. Thanks!

tsurdilo:
@MarianMacik thanks. will review again today and get back to you.

specification.md
MarianMacik: ISO 8601 has been reverted. Just examples were updated to show a correct possibility of interval syntax (no start/end).

ricardozanini (approved):

PR #636, opened 2022-06-03
Minor fixes on spec

hbelmiro:
- [x] Specification
- [ ] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

Fixes broken links and fixes fields that should be described as enum

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -3900,14 +3900,14 @@ onErrors definition. Note that errors raised during functions that are invoked a
 
 Allows defining invocation of a function via event. 
 
-| Parameter | Description | Type | Required |
-| --- | --- | --- | --- |
-| [produceEventRef](#Event-Definition) | Reference to the unique name of a `produced` event definition | string | yes |
-| [consumeEventRef](#Event-Definition) | Reference to the unique name of a `consumed` event definition | string | no |
-| consumeEventTimeout | Maximum amount of time (ISO 8601 format) to wait for the consume event. If not defined it be set to the [actionExecutionTimeout](#Workflow-Timeout-Definition) | string | no |
-| data | If string type, an expression which selects parts of the states data output to become the data (payload) of the event referenced by `produceEventRef`. If object type, a custom object to become the data (payload) of the event referenced by `produceEventRef`. | string or object | no |
-| contextAttributes | Add additional event extension context attributes to the trigger/produced event | object | no |
-| invoke | Specifies if the function should be invoked sync or async. Default is sync | string | no |
+| Parameter                            | Description                                                                                                                                                                                                                                                       | Type             | Required |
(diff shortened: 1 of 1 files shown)


specification.md
```diff
@@ -3827,12 +3827,12 @@ Note that if used with `string` type, the invocation of the function is synchron
 If you need to define parameters in your `functionRef` definition, you can define
 it with its `object` type which has the following properties:
 
-| Parameter | Description | Type | Required |
-| --- | --- | --- | --- |
-| refName | Name of the referenced [function](#Function-Definition) | string | yes |
-| arguments | Arguments (inputs) to be passed to the referenced function | object | yes if function type is `graphql`, otherwise no |
+| Parameter    | Description                                                                                                                                  | Type   | Required                                        |
```
tsurdilo, line 3830: I am not markdown expert but isnt this type of space formatting harder to maintain than

        | --- | --- | --- | --- |

hbelmiro: There's nothing in the documentation related to this.
I'm using IntelliJ and it complains if the table is not properly formatted. Then it offers an autofix to format the tables. VSCode formatter also format these tables.
The advantage of well formatted tables is that they are much easier to read. But I understand that on IDEs that don't do that automatically it might be harder to maintain.
Since VSCode is a general language IDE, I think we can assume that IDEs in general can easily format tables.

Anyway, I can format all of them or revert it, whatever you guys think is better.

hbelmiro: IntelliJ has native support to Markdown. https://www.jetbrains.com/help/idea/markdown.html#table
On VSCode I just noticed I have an extension for that: https://marketplace.visualstudio.com/items?itemName=fcrespo82.markdown-table-formatter

ricardozanini: Thank you for the explanation, @hbelmiro! @tsurdilo I'll start working on the CI/build tools to format the docs. Usually, people have a shell script to do the formatting in these repos. I'll look around.

ricardozanini (approved):

specification.md
tsurdilo: i think in end markdown should be human readable so users dont need to see parsed page and still be able to read doc if possible, imo :)

ricardozanini: yeah I agree. My comment was related to keep the formatting uniform.

cdavernas: Can the conversation be marked as resolved, so that we can proceed with merging this PR?

tsurdilo: i think the tables should be human redable. the formatting seems to make this much harder imo. wdyt?

cdavernas: I agree, but with MD they hardly ever are anyways. But yeah, imposing formatting everywhere sounds like an avoidable overhead.

cdavernas: @hbelmiro Cheers!!!

PR #641, opened 2022-06-20
Resource and invocation authentication definition

### LINKED ISSUE #612: Authentication for REST requests

cdavernas:

- [x] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

**Discussion or Issue link**:
Closes #612 

Adds the possibility to configure the `authDefinition` to use for both function's resource access and function invocation.

As discussed in #612, using a string-based `authRef` defines the `authDefinition` to use to invoke the configured function.

Also, as proposed by @ricardozanini, if the `resource` auth of a function has been defined, all other functions using the same resource (as defined by the `operation` property) must implicitly used the configured authDefinition. In other words, a resource auth definition is tightly coupled to its resource, even if it is omitted by other functions using the same resource.


ricardozanini:
hmm there are a lot of commits here, I think you messed with the rebase, @cdavernas? 😆

cdavernas:
@ricardozanini OMG 😭 That's the rebase I had to do for signing off...

PR #647, opened 2022-06-27
Fixes #420 - Add http invocation function definition

### LINKED ISSUE #420: Refactor `rest` Operation

1. Rename the `rest` operation type into `openapi`: the actual REST operation is an `openapi` operation, and could therefore be confusing to some.
2. Create a `rest` operation type, which allows configuring a request from top to bottom, thus enabling out-of-the-box support for legacy systems.

The `rest` operation definition could look like this:

```yaml
- name: MyOperation
  operation: 'http://mysite.com/api/{pathParam}#POST'
  type: rest
```
And it's reference like this:

```yaml
- name: MyAction
  functionRef:
    refName: MyOperation
    arguments:
      headers:
        - name: headerParam
          value: 'headerValue'
      path:
        - name: pathParam   #references the {pathParam} value in the operation url
          value: 'myendpoint' #replaces the {pathParam} value in the operation url
      query:
        - name: queryParam
          value: 'queryvalue'
      cookie:
        - name: cookieParam
          value: 'cookievalue'
      body:
        firstName: Tihomir
        lastName: Surdilovic
```

**Why is this needed**:

1. Calling a cat a cat 😸 
2. Allows out-of-the-box support for legacy systems and broadens SW use cases even further 🔫 

**Notes for implementers**:

The idea is basically to allow users to somehow mimic an OpenAPI doc in place of a function reference's `arguments` property, thus allowing the invocation of "dumb" http services.

ricardozanini:
Signed-off-by: Ricardo Zanini <zanini@redhat.com>

- [x] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

**Discussion or Issue link**: #420 

- Limit the `rest` function to OpenAPI specification descriptors
- Add the new `http` function type for simple use cases that requires an HTTP invocation. REST operations can still be performed if well defined in the function definition. Of course, simple use cases are also supported. 

This is a WIP.

**Additional information:** Based on https://www.rfc-editor.org/rfc/rfc9110.html

### DIFF
--- a/schema/functions.json
+++ b/schema/functions.json
@@ -35,16 +35,15 @@
           "minLength": 1
         },
         "operation": {
-          "type": "string",
-          "description": "If type is `rest`, <path_to_openapi_definition>#<operation_id>. If type is `asyncapi`, <path_to_asyncapi_definition>#<operation_id>. If type is `rpc`, <path_to_grpc_proto_file>#<service_name>#<service_method>. If type is `graphql`, <url_to_graphql_endpoint>#<literal \\\"mutation\\\" or \\\"query\\\">#<query_or_mutation_name>. If type is `odata`, <URI_to_odata_service>#<Entity_Set_Name>. If type is `expression`, defines the workflow expression. If type is `http`, <HTTP_method>#<request_target>",
-          "minLength": 1
+          "type": "object",
+          "$ref": "#/definitions/operation"
         },
         "type": {
           "type": "string",
-          "description": "Defines the function type. Is either `rest`, `http`,`asyncapi, `rpc`, `graphql`, `odata`, `expression`, or `custom`. Default is `rest`",
+          "description": "Defines the function type. Is either `rest`, `openapi`,`asyncapi, `rpc`, `graphql`, `odata`, `expression`, or `custom`. Default is `openapi`.",
           "enum": [
             "rest",
-            "http",
+            "openapi",
             "asyncapi",
             "rpc",
             "graphql",
@@ -52,7 +51,7 @@
             "expression",
             "custom"
           ],
-          "default": "rest"
+          "default": "openapi"
         },
         "authRef": {
           "oneOf": [
@@ -64,13 +63,13 @@
             {
               "type": "object",
               "description": "Configures both the auth definition used to retrieve the operation's resource and the auth definition used to invoke said operation",
-              "properties":{
-                "resource":{
+              "properties": {
+                "resource": {
                   "type": "string",
                   "description": "References an auth definition to be used to access the resource defined in the operation parameter",
                   "minLength": 1
                 },
-                "invocation":{
+                "invocation": {
                   "type": "string",
                   "description": "References an auth definition to be used to invoke the operation"
                 }
@@ -90,6 +89,25 @@
       "required": [
         "name",
         "operation"
+      ]
+    },
+    "operation": {
+      "oneOf": [
+        {
+          "type": "string",
+          "description": "If type is `openapi`, <path_to_openapi_definition>#<operation_id>. If type is `asyncapi`, <path_to_asyncapi_definition>#<operation_id>. If type is `rpc`, <path_to_grpc_proto_file>#<service_name>#<service_method>. If type is `graphql`, <url_to_graphql_endpoint>#<literal \\\"mutation\\\" or \\\"query\\\">#<query_or_mutation_name>. If type is `odata`, <URI_to_odata_service>#<Entity_Set_Name>. If type is `expression`, defines the workflow expression.",
+          "minLength": 1
+        },
+        {
+          "type": "object",
+          "description": "OpenAPI Path Object definition",
+          "$comment": "https://spec.openapis.org/oas/v3.1.0#paths-object",
+          "patternProperties": {
+            "^/": {
+              "type": "object"
+            }
+          }
+        }
       ]
     }
   }
--- a/schema/functions.json
+++ b/schema/functions.json
@@ -54,9 +54,32 @@
           "default": "rest"
         },
         "authRef": {
-          "type": "string",
-          "description": "References an auth definition name to be used to access to resource defined in the operation parameter",
-          "minLength": 1
+          "oneOf": [
+            {
+              "type": "string",
+              "description": "References the auth definition to be used to invoke the operation",
+              "minLength": 1
+            },
+            {
+              "type": "object",
+              "description": "Configures both the auth definition used to retrieve the operation's resource and the auth definition used to invoke said operation",
+              "properties":{
+                "resource":{
+                  "type": "string",
+                  "description": "References an auth definition to be used to access the resource defined in the operation parameter",
+                  "minLength": 1
+                },
+                "invocation":{
+                  "type": "string",
+                  "description": "References an auth definition to be used to invoke the operation"
+                }
+              },
+              "additionalProperties": false,
+              "required": [
+                "resource"
+              ]
+            }
+          ]
         },
         "metadata": {
           "$ref": "common.json#/definitions/metadata"
--- a/schema/functions.json
+++ b/schema/functions.json
@@ -36,14 +36,15 @@
         },
         "operation": {
           "type": "string",
-          "description": "If type is `rest`, <path_to_openapi_definition>#<operation_id>. If type is `asyncapi`, <path_to_asyncapi_definition>#<operation_id>. If type is `rpc`, <path_to_grpc_proto_file>#<service_name>#<service_method>. If type is `graphql`, <url_to_graphql_endpoint>#<literal \\\"mutation\\\" or \\\"query\\\">#<query_or_mutation_name>. If type is `odata`, <URI_to_odata_service>#<Entity_Set_Name>. If type is `expression`, defines the workflow expression.",
+          "description": "If type is `rest`, <path_to_openapi_definition>#<operation_id>. If type is `asyncapi`, <path_to_asyncapi_definition>#<operation_id>. If type is `rpc`, <path_to_grpc_proto_file>#<service_name>#<service_method>. If type is `graphql`, <url_to_graphql_endpoint>#<literal \\\"mutation\\\" or \\\"query\\\">#<query_or_mutation_name>. If type is `odata`, <URI_to_odata_service>#<Entity_Set_Name>. If type is `expression`, defines the workflow expression. If type is `http`, <HTTP_method>#<request_target>",
           "minLength": 1
         },
         "type": {
           "type": "string",
-          "description": "Defines the function type. Is either `rest`, `asyncapi, `rpc`, `graphql`, `odata`, `expression`, or `custom`. Default is `rest`",
+          "description": "Defines the function type. Is either `rest`, `http`,`asyncapi, `rpc`, `graphql`, `odata`, `expression`, or `custom`. Default is `rest`",
           "enum": [
             "rest",
+            "http",
             "asyncapi",
             "rpc",
             "graphql",
--- a/specification.md
+++ b/specification.md
@@ -23,9 +23,10 @@
     + [Using multiple data filters](#using-multiple-data-filters)
     + [Data Merging](#data-merging)
   * [Workflow Functions](#workflow-functions)
-    + [Using Functions for RESTful Service Invocations](#using-functions-for-restful-service-invocations)
+    + [Using Functions for OpenAPI RESTFul Service Invocations](#using-functions-for-openapi-restful-service-invocations)
+    + [Using Functions for HTTP Service Invocations](#using-functions-for-http-service-invocations)
     + [Using Functions for Async API Service Invocations](#using-functions-for-async-api-service-invocations)
-    + [Using Functions for RPC Service Invocations](#using-functions-for-rpc-service-invocations)
+    + [Using Functions for gRPC Service Invocations](#using-functions-for-grpc-service-invocations)
     + [Using Functions for GraphQL Service Invocations](#using-functions-for-graphql-service-invocations)
       - [Invoking a GraphQL `Query`](#invoking-a-graphql-query)
       - [Invoking a GraphQL `Mutation`](#invoking-a-graphql-mutation)
@@ -991,8 +992,9 @@ They can be referenced by their domain-specific names inside workflow [states](#
 
 Reference the following sections to learn more about workflow functions:
 
-* [Using functions for RESTful service invocations](#Using-Functions-for-RESTful-Service-Invocations)
-* [Using Functions for Async API Service Invocations](#Using-Functions-for-Async-API-Service-Invocations)
+* [Using functions for OpenAPI RESTful service invocations](#using-functions-for-openapi-restful-service-invocations)
++ [Using functions for HTTP Service Invocations](#using-functions-for-http-service-invocations)
+* [Using functions for Async API Service Invocations](#Using-Functions-for-Async-API-Service-Invocations)
 * [Using functions for gRPC service invocation](#Using-Functions-For-RPC-Service-Invocations)
 * [Using functions for GraphQL service invocation](#Using-Functions-For-GraphQL-Service-Invocations)
 * [Using Functions for OData Service Invocations](#Using-Functions-for-OData-Service-Invocations)
@@ -1002,7 +1004,7 @@ Reference the following sections to learn more about workflow functions:
 We can define if functions are invoked sync or async. Reference
 the [functionRef](#FunctionRef-Definition) to learn more on how to do this.
 
-#### Using Functions for RESTful Service Invocations
+#### Using Functions for OpenAPI RESTful Service Invocations
 
 [Functions](#Function-Definition) can be used to describe services and their operations that need to be invoked during
 workflow execution. They can be referenced by states [action definitions](#Action-Definition) to clearly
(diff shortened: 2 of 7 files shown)


specification.md
```diff
@@ -23,9 +23,10 @@
     + [Using multiple data filters](#using-multiple-data-filters)
     + [Data Merging](#data-merging)
   * [Workflow Functions](#workflow-functions)
-    + [Using Functions for RESTful Service Invocations](#using-functions-for-restful-service-invocations)
+    + [Using Functions for OpenAPI-based RESTFul Service Invocations](#using-functions-for-openapi-based-restful-service-invocations)
+    + [Using Functions for HTTP Service Invocations](#using-functions-for-http-service-invocations)
     + [Using Functions for Async API Service Invocations](#using-functions-for-async-api-service-invocations)
-    + [Using Functions for RPC Service Invocations](#using-functions-for-rpc-service-invocations)
+    + [Using Functions for gRPC Service Invocations](#using-functions-for-rpc-service-invocations)
```
tsurdilo, line 29: could also change the title of the section from rpc to grpc please

specification.md
```diff
@@ -1059,6 +1061,65 @@ Note that the referenced function definition type in this case must be `rest` (d
...
+The specification supports making HTTP invocations via a simplified interface in the [functions definition](#Function-Definition).
+
+The HTTP invocation interface can be described in the `operation` attribute of the function definition, as shown in the example:
+
+```json
+{
+"functions": [
+  {
+    "name": "queryUserById",
+    "operation": "get:/users/{id}"
```
tsurdilo, line 1081: i think best to keep operation to use same delimiter "#" as others would do just

"METHOD#EXPRESSION"

feel free to ping me if you want and we can talk

specification.md
```diff
@@ -1059,6 +1061,65 @@ Note that the referenced function definition type in this case must be `rest` (d
...
+
+Additional HTTP Headers can be passed via the `metadata` function definition. For example:
+
+```json
+{
+"functions": [
+  {
+    "name": "queryUserById",
+    "operation": "get:/users/{id}",
+    "metadata": {
```
tsurdilo, line 1107: wouldnt put metadata in spec doc, maybe if you want to showcase this is better to add it to a new sample

specification.md
```diff
@@ -1059,6 +1061,65 @@ Note that the referenced function definition type in this case must be `rest` (d
...
+    "name": "queryUserById",
+    "operation": "get:/users/{id}"
+  }
+]
+}
+```
+
+In this case, the `operation` attribute is composed by:
+
+* The HTTP method used by the HTTP invocation. Can be any [valid HTTP method](https://www.rfc-editor.org/rfc/rfc9110.html#name-method-definitions) such as `get`, `post`, `put`, etc.
```
tsurdilo, line 1089: should they be upper case letters? i think maybe if thats the standard we should enforce it as well

tsurdilo: POST, GET, PUT, PATCH, and DELETE

specification.md
ricardozanini: In this case, HTTP Headers are metadata of the request in a sense. What we can do is:

1. Not to describe headers in the operation or function definition and encourage implementations to add a way to do it via configurations. We can argue that's an infra use case.
2. Let the metadata solve this problem like we are doing
3. Introduce an extension for this case, so implementations can optionally use them, so users could define their custom headers there.

I don't have a strong opinion, but I like the idea to describe headers params using the `metadata` field: https://www.rfc-editor.org/rfc/rfc9110.html#section-6

> a headers lookup table of name/value pairs for extending that control data and conveying additional information about the sender, message, content, or context

specification.md
ricardozanini: I was struggling with this because to me the `operation` field could be a valid URI. But yeah, I guess we can go for `UPDATE#/users/{id}`.

ricardozanini:
@cdavernas mind taking a look? I think we can try instantiating a few use cases to verify if this architecture is suitable IRL.

cdavernas:
@ricardozanini Will do tomorrow morning!

specification.md
tsurdilo, line 29: please rename the header where this links to: #using-functions-for-rpc-service-invocations

specification.md
```diff
@@ -1059,6 +1061,140 @@ Note that the referenced function definition type in this case must be `rest` (d
 
 For more information about functions, reference the [Functions definitions](#Function-Definition) section.
 
+#### Using functions for HTTP Service Invocations
+
+For a few reasons, workflow designers might see necessary making straight HTTP invocations to other systems, such as:
```
tsurdilo, line 1066: I would not mention this just get to point what it does :)

specification.md
```diff
@@ -1059,6 +1061,140 @@ Note that the referenced function definition type in this case must be `rest` (d
...
+The specification supports describing HTTP invocations via a simplified interface in the [functions definition](#Function-Definition).
+
+The HTTP invocation interface can be described in the `operation` attribute of the function definition, as shown in the example:
+
+```json
+{
+"functions": [
+  {
+    "name": "queryUserById",
+    "operation": "GET#/users/{id}",
```
tsurdilo, line 1081: would it be ok to not use relative url
i assume since type is http that this does not support https, so we could say that "http:// is pre-pended always 
and operation could be like "GET#myapp.com/users{id}"

specification.md
tsurdilo: ok but see this from user perspective, spec has now a built in feature for http 
but your core example shows the use of metadata to get it working

im not saying not to like use metadata, but for core spec definition http invocation 
metadata should not be necessary.

tsurdilo: imo all we need to do here is add the most basic way to do a get/post
nothing more nothing else, make samples as simple as possible please

specification.md
```diff
@@ -1059,6 +1061,140 @@ Note that the referenced function definition type in this case must be `rest` (d
...
+      "x-custom-header": "the-value"
+    }
+  }
+]
+}
+```
+
+> Runtimes are encouraged to support at least `Content-type: application/json` header when making HTTP invocations.
+
+Avoid passing sensible information in headers such as authentication tokens. See the [auth definition](#auth-definition) for more information.
```
tsurdilo, line 1119: spec has secrets so not sure why user cannot set them? 
so wait, auth def applies here to? we do not support https

specification.md
```diff
@@ -1059,6 +1061,140 @@ Note that the referenced function definition type in this case must be `rest` (d
...
+    "operation": "POST#/users",
+    "type": "http"
+  }
+]
+}
+```
+
+Then you can reference the `createUser` function and filter the data to invoke it.
+
+Data example:
```
tsurdilo, line 1159: this is workflow data input right? maybe worth making it clear.

tsurdilo:
> @ricardozanini Will do tomorrow morning!

also don't forget that this is spec and the dsl needs to be user friendly first ;)

specification.md
```diff
@@ -1059,6 +1061,140 @@ Note that the referenced function definition type in this case must be `rest` (d
...
 
+#### Using functions for HTTP Service Invocations
+
+For a few reasons, workflow designers might see necessary making straight HTTP invocations to other systems, such as:
+
+1. Target service might not expose an OpenAPI specification descriptor;
+2. Target service might not implement RESTful interfaces;
+3. The use case requires a simple HTTP request invocation.
+
+The specification supports describing HTTP invocations via a simplified interface in the [functions definition](#Function-Definition).
```
gibson042, line 1072:
```suggestion
The specification supports describing HTTP requests via a simplified interface in the [functions definition](#Function-Definition).
```
(cf. https://www.rfc-editor.org/rfc/rfc9110.html#name-messages )

specification.md
```diff
@@ -1059,6 +1061,140 @@ Note that the referenced function definition type in this case must be `rest` (d
...
+
+The HTTP invocation interface can be described in the `operation` attribute of the function definition, as shown in the example:
```
gibson042, line 1074:
```suggestion
Here is an example function definition for HTTP requests with method `GET` and request target corresponding with [URI Template](https://www.rfc-editor.org/rfc/rfc6570.html) `/users/{id}` and host ???.
```

specification.md
gibson042: Agreed. How is the scheme specified (most relevantly "http:" vs. "https:")? How is the host specified?

specification.md
```diff
@@ -1059,6 +1061,140 @@ Note that the referenced function definition type in this case must be `rest` (d
...
+  {
+    "name": "queryUserById",
+    "operation": "GET#/users/{id}",
+    "type": "http"
+  }
+]
+}
+```
+
+The `operation` attribute is composed by a string separated by the hashtag symbol (`#`):
```
gibson042, line 1088:
`````suggestion
Note that the [Function Definition](#Function-Definition)'s `operation` property must have the following format:

```text
<HTTP_method>#<request_target>
```
`````
(cf. https://www.rfc-editor.org/rfc/rfc9110.html#name-messages )

File: specification.md [hunk 12]
```diff
@@ -1059,6 +1061,140 @@ Note that the referenced function definition type in this case must be `rest` (d
...
+
+* The HTTP method used by the HTTP invocation. Can be any [valid HTTP method](https://www.rfc-editor.org/rfc/rfc9110.html#name-method-definitions) such as `GET`, `POST`, `PUT`, etc.
```
gibson042, line 1090: [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-methods) defines only "standardized methods that are commonly used", but HTTP supports many more than that: https://www.iana.org/assignments/http-methods/http-methods.xhtml#methods

specification.md
```diff
@@ -1059,6 +1061,140 @@ Note that the referenced function definition type in this case must be `rest` (d
...
+* The endpoint **location path**, not the full endpoint URL. Runtimes implementations are **discouraged** to expect the full URL in this field since this is a matter of infrastructure configuration.
```
gibson042, line 1091: What does this mean? Where are non-HTTP terms like "location path" defined?

specification.md
```diff
@@ -1059,6 +1061,140 @@ Note that the referenced function definition type in this case must be `rest` (d
...
+
+The table below lists how to describe the HTTP parameters in the `operation` attribute:
+
+| Parameter | Description | Example |
+| --- | --- | --- |
+| Path | Path parameters are part of the location and must be enclosed with brackets (`{}`). The parameter name can be later referenced by the workflow states. | `/user/{id}` |
+| Query | [Query parameters](https://www.rfc-editor.org/rfc/rfc9110.html#section-4.2.2) must be enclosed with brackets (`{}`) after the query key separator (`=`). The parameter name can be later referenced by the workflow states. | `/user?status={status}` |
```
gibson042, line 1098:
```suggestion
| Path | Path parameters must be enclosed with braces (`{}`) as defined by [URI Templates](https://www.rfc-editor.org/rfc/rfc6570.html). The parameter name can be later referenced by the workflow states. | `/user/{id}` |
| Query | [Query parameters](https://www.rfc-editor.org/rfc/rfc9110.html#section-4.2.2) must be enclosed with braces (`{}`) as defined by [URI Templates](https://www.rfc-editor.org/rfc/rfc6570.html). The parameter name can be later referenced by the workflow states. | `/user?status={status}` |
```

Also, note that URI Templates support a more straightforward `/user{?status}` syntax for query parameters: https://www.rfc-editor.org/rfc/rfc6570.html#section-3.2.8

specification.md
```diff
@@ -1059,6 +1061,140 @@ Note that the referenced function definition type in this case must be `rest` (d
...
+
+Additional HTTP Headers can be passed via the `metadata` function definition. For example:
```
gibson042, line 1100:
```suggestion
Additional HTTP header fields can be passed via the `metadata` function definition. For example:
```
(cf. https://www.rfc-editor.org/rfc/rfc9110.html#name-header-fields )

specification.md
```diff
@@ -1059,6 +1061,140 @@ Note that the referenced function definition type in this case must be `rest` (d
...
+    "type": "http",
+    "metadata": {
+      "x-custom-header": "the-value"
+    }
+  }
+]
+}
+```
+
+> Runtimes are encouraged to support at least `Content-type: application/json` header when making HTTP invocations.
```
gibson042, line 1117:
```suggestion
> Runtimes are encouraged to support at least `Content-Type: application/json` header field for HTTP requests.
```
Also, how much freedom is allowed to runtimes here? What _MUST_ they support, what _MAY_ they support, and what _MUST NOT_ they support?

specification.md
gibson042:
```suggestion
Avoid passing sensitive information in headers such as authentication tokens. See the [auth definition](#auth-definition) for more information.
```

specification.md
```diff
@@ -1059,6 +1061,140 @@ Note that the referenced function definition type in this case must be `rest` (d
...
+
+Avoid passing sensible information in headers such as authentication tokens. See the [auth definition](#auth-definition) for more information.
+
+HTTP request body doesn't need to be described in the function definition. **All the data within the state** should be passed to the HTTP request. See the [action data filter](#action-data-filters) if you need to limit the amount of data passed to a given action.
```
gibson042, line 1121:
```suggestion
HTTP request content doesn't need to be described in the function definition. **All the data within the state** should be passed to the HTTP request. See the [action data filter](#action-data-filters) if you need to limit the amount of data passed to a given action.
```
(cf. https://www.rfc-editor.org/rfc/rfc9110.html#name-content )

> **All the data within the state** should be passed to the HTTP request.

In what format?

specification.md
```diff
@@ -1059,6 +1061,140 @@ Note that the referenced function definition type in this case must be `rest` (d
...
+      "name":"CreateNewUser",
+      "type":"operation",
+      "actions":[
+       {
+       "functionRef": "createUser",
+       "actionDataFilter": {
+          "fromStateData": "${ .user }",
+          "toStateData": "${ .user.id }"
+       }
+      }],
```
gibson042, line 1183: Is there any reason for this JSON to not be aligned?

specification.md
```diff
@@ -1059,6 +1061,140 @@ Note that the referenced function definition type in this case must be `rest` (d
...
+      "end": true
+  }]
+}
+```
+
+In this case, only the contents of the `user` attribute will be passed to the function. The user ID returned by the HTTP request body will then be added to the state data:
```
gibson042, line 1189: > only the contents of the `user` attribute will be passed to the function

What controls the format?

specification.md
```diff
@@ -1148,7 +1284,7 @@ Our defined function definition can then we referenced in a workflow [action](#A
 }
 ```
 
-#### Using Functions for RPC Service Invocations
+#### Using Functions for gRPC Service Invocations
```
gibson042, line 1287: This breaks the two preceding `[…](#using-functions-for-rpc-service-invocations)` links.

schema/functions.json
```diff
@@ -36,14 +36,15 @@
         },
         "operation": {
           "type": "string",
-          "description": "If type is `rest`, <path_to_openapi_definition>#<operation_id>. If type is `asyncapi`, <path_to_asyncapi_definition>#<operation_id>. If type is `rpc`, <path_to_grpc_proto_file>#<service_name>#<service_method>. If type is `graphql`, <url_to_graphql_endpoint>#<literal \\\"mutation\\\" or \\\"query\\\">#<query_or_mutation_name>. If type is `odata`, <URI_to_odata_service>#<Entity_Set_Name>. If type is `expression`, defines the workflow expression.",
+          "description": "If type is `rest`, <path_to_openapi_definition>#<operation_id>. If type is `asyncapi`, <path_to_asyncapi_definition>#<operation_id>. If type is `rpc`, <path_to_grpc_proto_file>#<service_name>#<service_method>. If type is `graphql`, <url_to_graphql_endpoint>#<literal \\\"mutation\\\" or \\\"query\\\">#<query_or_mutation_name>. If type is `odata`, <URI_to_odata_service>#<Entity_Set_Name>. If type is `expression`, defines the workflow expression. If type is `http`, <http_method>#<location_path>",
```
gibson042, line 39:
```suggestion
          "description": "If type is `rest`, <path_to_openapi_definition>#<operation_id>. If type is `asyncapi`, <path_to_asyncapi_definition>#<operation_id>. If type is `rpc`, <path_to_grpc_proto_file>#<service_name>#<service_method>. If type is `graphql`, <url_to_graphql_endpoint>#<literal \\\"mutation\\\" or \\\"query\\\">#<query_or_mutation_name>. If type is `odata`, <URI_to_odata_service>#<Entity_Set_Name>. If type is `expression`, defines the workflow expression. If type is `http`, <HTTP_method>#<request_target>",
```

specification.md
```diff
@@ -3189,11 +3325,24 @@ Note that `transition` and `end` properties are mutually exclusive, meaning that
...
+| type | Defines the function type. Can be either `rest`, `http`, `asyncapi`, `rpc`, `graphql`, `odata`, `expression`, or [`custom`](#defining-custom-function-types). Default is `rest` | enum | no |
 | authRef | References an [auth definition](#Auth-Definition) name to be used to access to resource defined in the operation parameter | string | no |
 | [metadata](#Workflow-Metadata) | Metadata information. Can be used to define custom function information | object | no |
 
+Function Operation description by type:
+
+| Type | Operation Description |
+| ---- | --------- |
+| `rest` | <path_to_openapi_definition>#<operation_id> |
+| `http` | <http_method>#<location_path> |
```
gibson042, line 3338:
```suggestion
| `http` | <HTTP_method>#<request_target> |
```

specification.md
tsurdilo: +100! 
i think we are opening a can of worms with this honestly unless we do add all these methods.
with this and all other comments from @gibson042 just confirms that maybe we really just need to allow runtimes to define this via extensions and provide a simple example only (showing that http is doable if wanted) and not get this "monster" into spec....

cdavernas: @tsurdilo I disagree. Those methods are slmost never used, if ever. We could just say that they are all supported but that for the sake of brevity only common ones are documented.

@gibson042 I would personally fo the following:

`{method}#{url including scheme and host}`

schema/functions.json
cdavernas: Here I would do the following, as suggested in last weekly call:

- replace `rest` by `openapi`
- replace `rpc` by `grpc`

Falling to do so is admitting that we prefer calling a dog a cat because that's what we did until now

specification.md
tsurdilo: my problem is not i dont like the functionality its that we can already do this two ways via custom function type and expression function types. so if this does not from user perspective give some ability even if only limited like u said with grpc then its just adding more confusion than it solves

tsurdilo: and iirc we added custom function type before for just this :) 
and to add with this we are sctually adding a custom custom type cause not depending on a specification

specification.md
```diff
@@ -1002,7 +1004,7 @@ Reference the following sections to learn more about workflow functions:
 We can define if functions are invoked sync or async. Reference
 the [functionRef](#FunctionRef-Definition) to learn more on how to do this.
 
-#### Using Functions for RESTful Service Invocations
+#### Using Functions for OpenAPI-based RESTful Service Invocations
```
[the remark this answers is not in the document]
ricardozanini: I agree, brain fart. I'll change it.

specification.md
ricardozanini: Sorry, I disagree. 

> Agreed. How is the scheme specified (most relevantly "http:" vs. "https:")? How is the host specified?

Implementations would instruct users on how to properly configure their endpoints per function. In a real use case scenario, these endpoints would change when deploying the workflow on different envs. 

I wouldn't expect users to hardcode the URLs here. The path is a description of the operation, not the full endpoint 

So, for instance:

```json
"functions": [
  {
    "name": "queryUserById",
    "operation": "GET#/users/{id}",
    "type": "http"
  } 
]
```

Describes that this function, in the **path** `/users/{id}` via the GET method would perform a read operation in the users' domain.

Implementations can find a way to provide an external configuration for users to reliably set up their deployments across environments without having to change the operation definition.

For example: `my.configuration.workflowid.funcid.host=http://myservice.com`.

specification.md
```diff
@@ -1059,6 +1061,140 @@ Note that the referenced function definition type in this case must be `rest` (d
...
+The `operation` attribute is composed by a string separated by the hashtag symbol (`#`):
+
+* The HTTP method used by the HTTP invocation. Can be any [valid HTTP method](https://www.rfc-editor.org/rfc/rfc9110.html#name-method-definitions) such as `GET`, `POST`, `PUT`, etc.
+* The endpoint **location path**, not the full endpoint URL. Runtimes implementations are **discouraged** to expect the full URL in this field since this is a matter of infrastructure configuration.
+
+The table below lists how to describe the HTTP parameters in the `operation` attribute:
+
+| Parameter | Description | Example |
+| --- | --- | --- |
+| Path | Path parameters are part of the location and must be enclosed with brackets (`{}`). The parameter name can be later referenced by the workflow states. | `/user/{id}` |
```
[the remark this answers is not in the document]
ricardozanini: I'm sorry, I don't understand. What are incorrect? This is a description of the path parameter in the `operation` attribute. I can make it more clear if needed.

schema/functions.json
ricardozanini: I would do in a different PR, if you don't mind?

specification.md
ricardozanini: See my other comment about this.

specification.md
ricardozanini: JSON, as implied by the spec. But I agree to add it here as well. Thanks!

specification.md
ricardozanini: Yes, I made a mistake.

specification.md
ricardozanini: Do you mean the content format? I don't follow. Everything is JSON from the spec perspective.

specification.md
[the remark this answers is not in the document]
ricardozanini: I see. I used the table because it was easier to explain, I'll change to a list.

specification.md
gibson042: Then why not point inside a _file_ expressing such configuration, as is done for
* [OpenAPI](https://github.com/serverlessworkflow/specification/blob/main/specification.md#using-functions-for-restful-service-invocations) (`"operation": "file://confirmationapi.json#sendOrderConfirmation"`), and
* [Async API](https://github.com/serverlessworkflow/specification/blob/main/specification.md#using-functions-for-async-api-service-invocations) (`"operation": "file://streetlightsapi.yaml#onLightMeasured"`), and
* [gRPC](https://github.com/serverlessworkflow/specification/blob/main/specification.md#using-functions-for-rpc-service-invocations) (`"operation": "file://myuserservice.proto#UserService#ListUsers"`), and
* [GraphQL](https://github.com/serverlessworkflow/specification/blob/main/specification.md#using-functions-for-graphql-service-invocations) (`"operation": "https://example.com/pets/graphql#query#pet"`), and
* [OData](https://github.com/serverlessworkflow/specification/blob/main/specification.md#using-functions-for-odata-service-invocations) (`"operation": "https://services.odata.org/V3/OData/OData.svc#Persons"`)

?

specification.md
gibson042: What other comment?

specification.md
gibson042: That is non-obvious when talking about generic HTTP requests, and might well need to change in the future.

specification.md
tsurdilo: ok lets discuss on zoom so u can explain me pls

specification.md
tsurdilo: yes, making it similar would to what we already have would be preferred pls

tsurdilo: still tho like been saying imo users should be able to define a single function def for all http calls which then is rly same thing as what we have already with custom function type. ill show example what i mean in meet

tsurdilo: i think we can handle all non-spec-compliant invocations by nust updating custom function type a little

PR #652, opened 2022-08-08
Fixes #420 - Add REST invocation function definition based on OpenAPI Path Object

### LINKED ISSUE #420: Refactor `rest` Operation

ricardozanini:

- [x] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

**Discussion or Issue link**:

See #647 

In our latest Community Meeting, we discussed the possibility of adding support for RESTful operation descriptors in the Function definitions based on OpenAPI Path Item object. Here's the PR. The former PR has been closed.

The "Security Object" will be ignored by the specification since we already have the Auth Definition to fit this role. Two places to define the same information can be troublesome and error prune.

**Additional information:**

See reference: https://spec.openapis.org/oas/v3.1.0#paths-object

### DIFF
--- a/schema/functions.json
+++ b/schema/functions.json
@@ -36,15 +36,15 @@
           "pattern": "^[a-z0-9](-?[a-z0-9])*$"
         },
         "operation": {
-          "type": "string",
-          "description": "If type is `rest`, <path_to_openapi_definition>#<operation_id>. If type is `asyncapi`, <path_to_asyncapi_definition>#<operation_id>. If type is `rpc`, <path_to_grpc_proto_file>#<service_name>#<service_method>. If type is `graphql`, <url_to_graphql_endpoint>#<literal \\\"mutation\\\" or \\\"query\\\">#<query_or_mutation_name>. If type is `odata`, <URI_to_odata_service>#<Entity_Set_Name>. If type is `expression`, defines the workflow expression.",
-          "minLength": 1
+          "type": "object",
+          "$ref": "#/definitions/operation"
         },
         "type": {
           "type": "string",
-          "description": "Defines the function type. Is either `rest`, `asyncapi, `rpc`, `graphql`, `odata`, `expression`, or `custom`. Default is `rest`",
+          "description": "Defines the function type. Is either `rest`, `openapi`,`asyncapi, `rpc`, `graphql`, `odata`, `expression`, or `custom`. Default is `openapi`.",
           "enum": [
             "rest",
+            "openapi",
             "asyncapi",
             "rpc",
             "graphql",
@@ -52,7 +52,7 @@
             "expression",
             "custom"
           ],
-          "default": "rest"
+          "default": "openapi"
         },
         "authRef": {
           "oneOf": [
@@ -65,14 +65,14 @@
             {
               "type": "object",
               "description": "Configures both the auth definition used to retrieve the operation's resource and the auth definition used to invoke said operation",
-              "properties":{
-                "resource":{
+              "properties": {
+                "resource": {
                   "type": "string",
                   "description": "References an auth definition to be used to access the resource defined in the operation parameter",
                   "minLength": 1,
                   "pattern": "^[a-z0-9](-?[a-z0-9])*$"
                 },
-                "invocation":{
+                "invocation": {
                   "type": "string",
                   "description": "References an auth definition to be used to invoke the operation"
                 }
@@ -92,6 +92,25 @@
       "required": [
         "name",
         "operation"
+      ]
+    },
+    "operation": {
+      "oneOf": [
+        {
+          "type": "string",
+          "description": "If type is `openapi`, <path_to_openapi_definition>#<operation_id>. If type is `asyncapi`, <path_to_asyncapi_definition>#<operation_id>. If type is `rpc`, <path_to_grpc_proto_file>#<service_name>#<service_method>. If type is `graphql`, <url_to_graphql_endpoint>#<literal \\\"mutation\\\" or \\\"query\\\">#<query_or_mutation_name>. If type is `odata`, <URI_to_odata_service>#<Entity_Set_Name>. If type is `expression`, defines the workflow expression.",
+          "minLength": 1
+        },
+        {
+          "type": "object",
+          "description": "OpenAPI Path Object definition",
+          "$comment": "https://spec.openapis.org/oas/v3.1.0#paths-object",
+          "patternProperties": {
+            "^/": {
+              "type": "object"
+            }
+          }
+        }
       ]
     }
   }
--- a/specification.md
+++ b/specification.md
@@ -1092,7 +1092,7 @@ Here is an example function definition for REST requests with method `GET` and r
 }
 ```
 
-Note that the [Function Definition](#Function-Definition)'s `operation` property must follow the OpenAPI Paths Object specification definition.
+Note that the [Function Definition](#Function-Definition)'s `operation` property must follow the [OpenAPI Paths Object](https://spec.openapis.org/oas/v3.1.0#paths-object) specification definition.
 
 The function can be referenced during workflow execution when the invocation of the REST service is desired. For example:
 
@@ -1123,6 +1123,7 @@ Example of the `POST` request sending the state data as part of the body:
   "functions":[
     {
       "name": "createUser",
+      "type": "rest",
       "operation": {
         "/users": {
           "post": {
@@ -1149,16 +1150,15 @@ Example of the `POST` request sending the state data as part of the body:
             }
           }
         }
-      },
-      "type": "rest"
+      }
     }
   ]
 }
 ```
 
-Note that the `requestBody` content schema is described inline rather than a reference to an external document.
+Note that the `requestBody` [`content` attribute](https://spec.openapis.org/oas/v3.1.0#fixed-fields-10) is described inline rather than a reference to an external document.
 
-You can reference the `createUser` function and filter the input data to invoke it, as shown in the example below:
+You can reference the `createUser` function and filter the input data to invoke it. Given the workflow input data:
 
 ```json
 {
@@ -1220,7 +1220,7 @@ In this case, only the contents of the `user` attribute will be passed to the fu
 }
 ```
 
-The specification does not support the [Security Requirement Object](https://spec.openapis.org/oas/v3.1.0#security-requirement-object). To define authentication for the REST operation, use the [Auth Definition](#Auth-Definition). If provided, this field is ignored.
+The specification does not support the [Security Requirement Object](https://spec.openapis.org/oas/v3.1.0#security-requirement-object) since its redundat to function [Auth Definition](#Auth-Definition). If provided, this field is ignored.
 
 #### Using Functions for Async API Service Invocations
 
@@ -3416,7 +3416,7 @@ operation: https://hellworldservice.api.com/api.json#helloWorld
 
 The `name` property defines an unique name of the function definition.
 
-The `type` property defines the function type. Its value can be either `rest`, `openapi` or `expression`. Default value is `openapi`.
+The `type` enum property defines the function type. Its value can be either `rest`, `openapi` or `expression`. Default value is `openapi`.
 
 Depending on the function `type`, the `operation` property can be:
 
--- a/specification.md
+++ b/specification.md
@@ -23,6 +23,7 @@
     + [Using multiple data filters](#using-multiple-data-filters)
     + [Data Merging](#data-merging)
   * [Workflow Functions](#workflow-functions)
+    + [Using Functions for OpenAPI Service Invocations](#using-functions-for-openapi-service-invocations)
     + [Using Functions for RESTful Service Invocations](#using-functions-for-restful-service-invocations)
     + [Using Functions for Async API Service Invocations](#using-functions-for-async-api-service-invocations)
     + [Using Functions for RPC Service Invocations](#using-functions-for-rpc-service-invocations)
@@ -994,8 +995,9 @@ They can be referenced by their domain-specific names inside workflow [states](#
 
 Reference the following sections to learn more about workflow functions:
 
-* [Using functions for RESTful service invocations](#Using-Functions-for-RESTful-Service-Invocations)
-* [Using Functions for Async API Service Invocations](#Using-Functions-for-Async-API-Service-Invocations)
+* [Using functions for OpenAPI Service invocations](#using-functions-for-openapi-service-invocations)
++ [Using functions for RESTful Service Invocations](#using-functions-for-rest-service-invocations)
+* [Using functions for Async API Service Invocations](#Using-Functions-for-Async-API-Service-Invocations)
 * [Using functions for gRPC service invocation](#Using-Functions-For-RPC-Service-Invocations)
 * [Using functions for GraphQL service invocation](#Using-Functions-For-GraphQL-Service-Invocations)
 * [Using Functions for OData Service Invocations](#Using-Functions-for-OData-Service-Invocations)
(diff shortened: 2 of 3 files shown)


schema/functions.json
```diff
@@ -90,6 +90,25 @@
         "name",
         "operation"
       ]
+    },
```
tsurdilo, line 93: i dont think this is right. we should find a way to restrict object type on the value of type property please.

tsurdilo: also description if object type should add something like "added only for rest type"

schema/functions.json
```diff
@@ -90,6 +90,25 @@
...
+      "oneOf": [
+        {
+          "type": "string",
+          "description": "If type is `openapi`, <path_to_openapi_definition>#<operation_id>. If type is `asyncapi`, <path_to_asyncapi_definition>#<operation_id>. If type is `rpc`, <path_to_grpc_proto_file>#<service_name>#<service_method>. If type is `graphql`, <url_to_graphql_endpoint>#<literal \\\"mutation\\\" or \\\"query\\\">#<query_or_mutation_name>. If type is `odata`, <URI_to_odata_service>#<Entity_Set_Name>. If type is `expression`, defines the workflow expression.",
+          "minLength": 1
+        },
+        {
+          "type": "object",
+          "description": "OpenAPI Path Object definition",
+          "$comment": "https://spec.openapis.org/oas/v3.1.0#paths-object",
```
tsurdilo, line 104: whats the "$" on comment prop. also why hardcode 3.1.0 version here. i think we should find a way to be version independent

specification.md
```diff
@@ -1055,10 +1057,168 @@ For example:
...
 ```
 
-Note that the referenced function definition type in this case must be `rest` (default type).
+Note that the referenced function definition type in this case must be `openapi` (default type).
 
+#### Using functions for RESTful Service Invocations
+
+The specification also supports describing REST invocations in the [functions definition](#Function-Definition) using [OpenAPI Paths Object](https://spec.openapis.org/oas/v3.1.0#paths-object).
```
tsurdilo, line 1066: I'd like to see here more description on **why** we use openapi for rest here. 
Something to extend that we do not want to create anything proprietary and want to use 
standards like everywhere else. Put some emphasis to users explain them whats the decision behind this.

tsurdilo: Look at some of the comparisons especially the one we have with google cloud workflow. i think there we mention that they have custom properties for rest / http that we could use to show more of the "why"

specification.md
```diff
@@ -1055,10 +1057,168 @@ For example:
...
+
+Here is an example function definition for REST requests with method `GET` and request target corresponding with [URI Template](https://www.rfc-editor.org/rfc/rfc6570.html) `/users/{id}`:
+
+```json
```
tsurdilo, line 1070: is it not possible for operation here to also be a string? link to file that contains the object? also i am not sure if "operation" should be top-level object here but need to think about this more.

specification.md
```diff
@@ -1055,10 +1057,168 @@ For example:
...
+          }
+        }
+      },
+      "type":"rest"
+    }
+  ]
+}
+```
+
+Note that the [Function Definition](#Function-Definition)'s `operation` property must follow the OpenAPI Paths Object specification definition.
```
tsurdilo, line 1092: link would be nice to have for OpenApi PAths object otherwise user has to google it themselves

specification.md
```diff
@@ -1055,10 +1057,168 @@ For example:
...
+      "type":"operation",
+      "actions":[
+        {
+          "functionRef":"queryUserById",
+          "arguments":{
+            "id":"${ .user.id }"
+          }
+        }
+      ],
+      "end":true
```
tsurdilo, line 1110: this seems not necessary maybe just "..." or use yaml and put a comment with "..."

specification.md
```diff
@@ -1055,10 +1057,168 @@ For example:
...
+    }
+  ]
+}
+```
+
+Example of the `POST` request sending the state data as part of the body:
+
+```json
+{
+  "functions":[
```
tsurdilo, line 1120: this is 30 lines of code and looks super ugly inside a workflow definition itself...and this is a small example. i really think we should not show this to users and use spring operation to a file that contains the object needed

tsurdilo: locala file is ok ..or even expression. i honestly do not like this when looking at it at all :(

specification.md
```diff
@@ -1055,10 +1057,168 @@ For example:
...
+                    },
+                    "required": ["name", "email"]
+                  }
+                }
+              }
+            }
+          }
+        }
+      },
+      "type": "rest"
```
tsurdilo, line 1150: i think type should be above operation for readability

specification.md
```diff
@@ -1055,10 +1057,168 @@ For example:
...
+  },
+  "user":{
+    "id":"5678U",
+    "name":"John Doe",
+    "email":"john@doe.com"
+  }
+}
+```
+
+The specification does not support the [Security Requirement Object](https://spec.openapis.org/oas/v3.1.0#security-requirement-object). To define authentication for the REST operation, use the [Auth Definition](#Auth-Definition). If provided, this field is ignored.
```
tsurdilo, line 1220: Hmmm this is weird. Why dont we use Security Requirement Object. How we answer user question on why?

tsurdilo: I don't see any info here on validating the inline openapi definitions. Is there anything we can add to tell users that yes we picked to do this cause xyz (like previous comment) and include that its possible to validate (if it really is, idk) something like that would be nice to have

specification.md
```diff
@@ -3189,11 +3349,24 @@ Note that `transition` and `end` properties are mutually exclusive, meaning that
 | Parameter | Description | Type | Required |
 | --- | --- | --- | --- |
 | name | Unique function name | string | yes |
-| operation | If type is `rest`, <path_to_openapi_definition>#<operation_id>. If type is `asyncapi`, <path_to_asyncapi_definition>#<operation_id>. If type is `rpc`, <path_to_grpc_proto_file>#<service_name>#<service_method>. If type is `graphql`, <url_to_graphql_endpoint>#<literal \"mutation\" or \"query\">#<query_or_mutation_name>. If type is `odata`, <URI_to_odata_service>#<Entity_Set_Name>. If type is `expression`, defines the workflow expression. | string | yes |
-| type | Defines the function type. Can be either `rest`, `asyncapi`, `rpc`, `graphql`, `odata`, `expression`, or [`custom`](#defining-custom-function-types). Default is `rest` | enum | no |
+| operation | See the table "Function Operation description by type" below. | string or object | yes |
```
tsurdilo, line 3352: could it be linked?

specification.md
```diff
@@ -3228,12 +3401,14 @@ operation: https://hellworldservice.api.com/api.json#helloWorld
 
 The `name` property defines an unique name of the function definition.
 
-The `type` property defines the function type. Its value can be either `rest` or `expression`. Default value is `rest`.
+The `type` property defines the function type. Its value can be either `rest`, `openapi` or `expression`. Default value is `openapi`.
 
 Depending on the function `type`, the `operation` property can be:
 
-* If `type` is `rest`, a combination of the function/service OpenAPI definition document URI and the particular service operation that needs to be invoked, separated by a '#'.
+* If `type` is `openapi`, a combination of the function/service OpenAPI definition document URI and the particular service operation that needs to be invoked, separated by a '#'.
```
tsurdilo, line 3408: is this duplicate content that we previously already tell user or something else?

cdavernas (approved):
Looks awesome, many thanks  ❤️! Exactly what I had in mind, too: you rule!

ricardozanini:
@cdavernas thanks! I'll rebase and deal with the comments today.

schema/functions.json
ricardozanini: I don't think so. The `AuthRef` object can be applied to any remote invocation such as gRPC, GraphQL, openapi, rest, odata, etc.

If we would make this change, I'd say to do it in another PR. There's no change here apart from the formatting and the default to `openapi` type.

schema/functions.json
ricardozanini: We have been using schema 7.x since the beginning, and I don't think it will break anything. `$comment` is pretty useful since SDKs/Tooling can use it to parse this data against the OpenAPI spec. So we don't need to replicate their structure here.
 
Additionally, schema 7.x is from [March 2018](https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-01#appendix-B).

Honestly, I don't see why we should not use `$comment` based on this tradeoff.

specification.md
ricardozanini: No, it's not possible to use string, since it's a complex object. We would have to "invent" a new way of expression this complex object in a string. The schema is already forcing it. This is what we agreed upon to have the OpenAPI spec path incorporated.

specification.md
ricardozanini: Yeah.. it's pretty big, but it's a possibility for a more complex operation, and I believe we should keep it to illustrating. Since we do support:

```json
{
    "functions": "file://myfuncs.json"
}
```

It is possible to externalize these definitions. Especially now with #691. wdyt?

specification.md
```diff
@@ -1055,10 +1057,168 @@ For example:
...
+          }
+        }
+      },
+      "type": "rest"
+    }
+  ]
+}
+```
+
+Note that the `requestBody` content schema is described inline rather than a reference to an external document.
```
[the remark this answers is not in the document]
ricardozanini: `content` is an attribute part of the OpenAPI path, I'll rephrase it.

specification.md
ricardozanini: > Hmmm this is weird. Why dont we use Security Requirement Object. How we answer user question on why?

Because it's redundant to our `Auth Definition`. I'll add this to the text.

> I don't see any info here on validating the inline openapi definitions. Is there anything we can add to tell users that yes we picked to do this cause xyz (like previous comment) and include that its possible to validate (if it really is, idk) something like that would be nice to have

We can validate based on the import schema we did in the functions schema JSON document (the`$comment` attribute). This way, SDKs and tools can use the OpenAPI spec to validate the operation.

specification.md
ricardozanini: We don't have a particular session for it; the mentioned table is two lines below.

ricardozanini:
@tsurdilo, I believe I've incorporated most of your comments. There's one left that I'll handle later. Can you take another look, please?

tsurdilo:
looking

tsurdilo:
Will do review today (one last time) if thats ok.

cdavernas:
@tsurdilo You had a chance to confirm this one?

cdavernas:
@antmendoza Could you review this PR so that we can proceed with merging it?

schema/functions.json
cdavernas: Resolved?

schema/functions.json
```diff
@@ -90,6 +90,25 @@
...
+    "operation": {
+      "oneOf": [
+        {
+          "type": "string",
+          "description": "If type is `openapi`, <path_to_openapi_definition>#<operation_id>. If type is `asyncapi`, <path_to_asyncapi_definition>#<operation_id>. If type is `rpc`, <path_to_grpc_proto_file>#<service_name>#<service_method>. If type is `graphql`, <url_to_graphql_endpoint>#<literal \\\"mutation\\\" or \\\"query\\\">#<query_or_mutation_name>. If type is `odata`, <URI_to_odata_service>#<Entity_Set_Name>. If type is `expression`, defines the workflow expression.",
+          "minLength": 1
+        },
+        {
+          "type": "object",
+          "description": "OpenAPI Path Object definition",
```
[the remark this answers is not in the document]
cdavernas: @ricardozanini @shaisachs makes a good (second) point. WDYT?

schema/functions.json
cdavernas: @tsurdilo Any further toughts? If not, please resolve ;)

specification.md
cdavernas: @tsurdilo ?

specification.md
cdavernas: > this is 30 lines of code and looks super ugly inside a workflow definition itself...and this is a small example. i really think we should not show this to users and use spring operation to a file that contains the object needed

@tsurdilo I totally agree with you, man! However, let's not forget this function type (and maybe this should be emphazied somewhere) is but for compatibility with "legacy" APIs, or undocumented ones over which the consumer has no power. In other words, one should (and probably would) always prefer using the `openapi` function type.

> It is possible to externalize these definitions. Especially now with #691. wdyt?

It is possible, indeed, but I would STRONGLY advise against it, as doing so would be equivalent to using a proper OpenAPI documentation file, thus using the `openapi` function type instead.

specification.md
cdavernas: Ok with you @tsurdilo ?

schema/functions.json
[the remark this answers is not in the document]
shaisachs: We discussed the first point offline and I get it now so no worries on that point. The second point is much more tactical and I think still something to consider.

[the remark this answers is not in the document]
cdavernas: @ricardozanini Any thoughts/feedback?

specification.md
ricardozanini: To me we keep the OpenAPI spec.

specification.md
ricardozanini: @shaisachs The reason behind this feature is to have the ability to define a rest operation without relying on an OpenAPI spec file. Thus, we incorporated part of the spec in the function definition since it's a well-known spec for REST operations definitions.

specification.md
ricardozanini: I agree. We can either add a note there adding this point or not mention it at all. I don't think we mention it.

schema/functions.json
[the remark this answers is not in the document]
ricardozanini: Let's move to `Path Item`!!

Many thanks, @shaisachs! @cdavernas I'll review it today so we can move forward.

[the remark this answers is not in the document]
ricardozanini: @shaisachs @cdavernas [I don't see the `path` attribute in the `path item` object](https://spec.openapis.org/oas/v3.1.0#path-item-object-example). We need the path, otherwise, we can't determine which path/function to call. That said, we should follow the Paths Object instead exactly as stated in this PR.

ricardozanini:
@cdavernas I'm gonna rebase this PR. Do you see anything else we need before merging it?

cdavernas:
@ricardozanini nice, cheers mate!
I do not see anything missing atm, no ;)

ricardozanini:
@cdavernas I'm gonna rebase and merge tomorrow.

PR #667, opened 2022-09-06
fix(*): update auth struct type to array or object

lsytj0413:
Signed-off-by: lsytj0413 <511121939@qq.com>

- [x] Other

**Discussion or Issue link**:

- According to the examples, the Auth field't type should be array or object, not array or string

**Additional information:**

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -1777,7 +1777,7 @@ definition "id" must be a constant value.
 | [timeouts](#Workflow-Timeouts) | Defines the workflow default timeout settings | string or object | no |
 | [errors](#Defining-Errors) | Defines checked errors that can be explicitly handled during workflow execution | string or array | no |
 | keepActive | If "true", workflow instances is not terminated when there are no active execution paths. Instance can be terminated with "terminate end definition" or reaching defined "workflowExecTimeout" | boolean | no |
-| [auth](#Auth-Definition) | Workflow authentication definitions | array or string | no |
+| [auth](#Auth-Definition) | Workflow authentication definitions | array or object | no |
 | [events](#Event-Definition) | Workflow event definitions.  | array or string | no |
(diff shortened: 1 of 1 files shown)


ricardozanini (approved):

cdavernas (changes requested):
No, the `auth` should indeed be of `array` or `string` types, I don't know which examples you are speaking about.

In the case `auth` is `string`, it is considered as a reference to an external definition of the `array` (ex: https://test.com/mysharedauth)

ricardozanini:
@lsytj0413, @cdavernas is right. The string, in this case, is an external reference. Can you please point out the examples you saw the error?

lsytj0413:
> @lsytj0413 No, the top level `auth` property should indeed be of `array` or `string` types, I don't know which examples you are speaking about, which would be where the error(s) is/are.
> 
> In the case `auth` is `string`, it is considered as a reference to an external definition of the `array` (ex: https://test.com/mysharedauth)

I have seen [this example](https://github.com/serverlessworkflow/specification/pull/667#pullrequestreview-1097807409) in sdk-go implement and it should validate with no error.

So, this should be an bug in sdk-go?

tsurdilo:
> I have seen [this example](https://github.com/serverlessworkflow/sdk-go/blob/main/parser/testdata/workflows/applicationrequest.json#L8) in sdk-go implement and it should validate with no error.

nice catch, i think the sample in go sdk should be updated

lsytj0413:
Ok，If we confirm it's an bug in sdk-go，i would like to fix it.

ricardozanini:
The SDK is implementing version 0.7 of the spec. Thanks for the PR; I've pushed it to the 0.8 milestone! :)

Issue #668, opened 2022-09-07
Auth definition being able to use $SECRET to avoid hardcoded passwords in file


fjtirado:
**What would you like to be added**:

In Auth Definition section, explicitly mention that users, rather than hardcode the password string, can use $SECRET (afaik, $SECRET usage is restricted to expressions)

**Why is this needed**:
To avoid hardcoding  passwords, either in the workflow file or the json/yaml file, referenced from the workflow file, containing auth definitions.

cdavernas:
@fjtirado You can already do that, actually, by settings the `properties` of an `authDef` to the secret name.

So, instead of doing:

```yaml
auth:
  - name: MyBasicAuth
    scheme: basic
    properties:
      username: myUserName
      password: ********
```

You can do:

```yaml
auth:
  - name: MyBasicAuth
    scheme: basic
    properties: 'mybasicauthsecret.yaml'
```

fjtirado:
@cdavernas  
Just for verifying we have the same understanding
Which will be the content of mybasicauthsecret.yaml?

cdavernas:
@fjtirado It is expected to match 1:1 on the chosen scheme's properties:

```yaml
{
   "username": "myUserName",
   "password": "MyPassword"
}
```

fjtirado:
@fjtirado 
Ok, the problem is that, even with the reference file, the password will still be visible, so the implementation of the spec needs to support $SECRET replacement also there. This is not 100% clear (I mean is currently open to intepretation)
Anyway, we both agree it should be supported ;)

cdavernas:
> Ok, the problem is that, even with the reference file the password will still be visible

I'm not sure I understand. It won't be visible in the definition, as the secret file will be mounted by the runtime. You can review the content of the secret file if you have access to it, but not through the workflow.

fjtirado:
@cdavernas 
The reference yaml file will still have the password visible. I agree it wont be visible anymore in the flow, of course ;)

tsurdilo:
dont think sw mandates it to be a json file its uri that in end can point to a secrets/token provider service of your choice. its as mentioned up to runtime to dowhat it wants.

fjtirado:
ok, focusing on the inline approach, what about explicitly establishing the $SECRET is supported for auth
Right now, we have
[...]

PR #669, opened 2022-09-07
${SECRET} might be used in auth definitions

fjtirado:

- [ ] Specification
- [ ] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

**Discussion or Issue link**:

**Additional information:**


tsurdilo:
Hey, sorry I don't fully understand the change. The first sentences says that $SECRETS can be used inside expressions, so any prop value type string where you can define expression. I think that auth props fall under this as well. Not sure why we remove "only" from first sentence as those two are not mutually exclusive. Can you explain? @cdavernas ?

fjtirado:
Expressions are not going to work for auth definition, which might be processed at build time (to load the URI the auth information is associated to)

tsurdilo:
I think no expressions would work at build time anyways as they are evaluated against the workflow state (data). I am confused sorry :( 
I think would be maybe better approach to define properties such as workflow id (top level) and version etc that we say cannot and should not use expressions, rather than the other way around.

tsurdilo:
Hey, sorry are we talking about https://github.com/serverlessworkflow/specification/blob/main/schema/auth.json#L9 by chance here? It's a type string and who says you cannot use an expression there? I am not sure that we define that uris have to be resolved at compile time anywhere?

fjtirado:
Im trying to support this https://github.com/serverlessworkflow/specification/issues/668#issuecomment-1239120599

tsurdilo:
Thanks, i understand better now, but who says that it cannot be

           properties: '${ $SECRETS.mysecretsobj }'

where mysecretsobj is a json object type? Maybe im missing something

fjtirado:
> Thanks, i understand better now, but who says that it cannot be
> 
> ```
>        properties: '${ $SECRETS.mysecretsobj }'
> ```
> 
> where mysecretsobj is a json object type? Maybe im missing something

see update comment https://github.com/serverlessworkflow/specification/pull/669#issuecomment-1239655437

tsurdilo:
Thanks, i think i kinda understand better now. But don't you then limit yourself as to having to compile your workflow definitions and everything on your prod machines? Secrets can differ between your staging and prod envs. 

I am not sure that your implementation strategies here affect the spec, spec could care less how you resolve that issue :)

fjtirado:
Ok, independently of the implementation strategy, any implementation will have two part (parsing and execution), even if the two are executed at the same time.
When an auth definition is going to be used during parsing (which is the case when downloading an authentication protected open api URI) the auth definition values cannot be resolved using arbitrary expressions (because there wont be any workflow model to validate the expression againts,  since the workflow is not being executed yet), therefore it makes sense to do an exception and state that you can use $SECRET.password rather than using a hardcoded one for auth definition values

tsurdilo:
> it makes sense to do an exception and state that you can use $SECRET.password rather than using a hardcoded one for auth definition values

full agree, and thats fine cause password param is a string type so you can hard code or use expression or a combination of the two (or call an expression function if you wanted). 

maybe this is a chicken and egg type of problem thats why there might be confusion on my end. i feel that all we have from spec side to state which string type params are not allowed to have values defines as expressions, the rest is up to you all

fjtirado:
> maybe this is a chicken and egg type of problem thats why there might be confusion on my end. i feel that all we have from spec side to state which string type params are not allowed to have values defines as expressions, the rest is up to you all

I feel that auth definition is one of those places where strings should not be expressions (at least for the initial defintion of auth definition, which only applies to the downloading of the function definition operation uri), basically becuase of what I already explain (operation uri is part of the flow definition, you cannot start the flow before processing it, so expressions cannot be evaluated). Anyway, I agree with you that the effort to be done is to mention which strings cannot be resolved as expressions

cdavernas:
> Expressions wont work for auth definition (when auth definition applies to the resolution of the URI)

Not always true. I guess that's how you chose to go with Kogito, but in Symapse we download the document at functions execution time, not at build time (because it might have changed in the meanwhile). Therefore, in Synapse you can use expressions on auth for an OpenApi document too

You are trying to enforce a rule (not allowing runtime expressions in auth) that applies solely on your use case. I personally would like to keep it going the way it was, and not start confusing people with partial data injection (just the $secrets) in one specific case

cdavernas:
> any smart implementation will do that rather than downloading it in every execution

Same comment as above: too opiniatef. What makes you think the only smart way is to predownload a document at a point in time that only exist in your use case (I.e. build time). Out of Kogito, what would build time even mean?

cdavernas:
> But this latter one will need secret support to avoid hardcoding the password)

Again I disagree. Imo the safest and cleanest way is to inject the secret object directly, instead of going through the pain of selecting the properties you want to 1:1 match, which would be an exotic (exceptional) use case

fjtirado:
@cdavernas  I think we are deviating from the original discussion, so lets rewind a bit. 
We agree that are fields (like version, name) of the workflow file which definitely are fixed (so should not be resolved through a runtime expression). And what we are discussing is if auth definition should support runtime expressions or not. 
I think the underneath issue is that auth definition definition is ambigous after latter modifications. Initially, auth definition was only used for including authorization information needed to download the operation uri (when this uri is an http one). This was changed to also mention that auth definition can also be used for invoking the operation. While the latter is obviously a runtime operation, the former is not necessarily a runtime one. By saying that auth definition should support runtime expression, you are imposing an architecture implementation decision too. So the more neutral approach will be to say, auth definition, depending on implementation, might support runtime expressions. It is up to the implementor to determine which expressions are supported there.

cdavernas:
> So the more neutral approach will be to say, auth definition, depending on implementation, might support runtime expressions. It is up to the implementor to determine which expressions are supported there.

Yeah, you are right, seems like both a reasonable and rational decision: on this implementers should be able to make their own ADR.

fjtirado:
@cdavernas @tsurdilo I changed the PR aligned what we have discussed.

fjtirado:
Yes, Ill open a new one

PR #670, opened 2022-09-09
Note allowing restrictions on auth definition expressions

fjtirado:
This allows implementors to impose restrictions on expression support for auth definition without affecting their compliance to the standard.

- [ ] Specification
- [ ] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

**Discussion or Issue link**:

**Additional information:**


specification.md
```diff
@@ -3555,12 +3555,14 @@ The `scheme` property defines the auth scheme to be used. Can be "bearer", "basi
 The `properties` property defines the auth scheme properties information.
 Can be one of ["Basic properties definition"](#basic-properties-definition), ["Bearer properties definition"](#bearer-properties-definition), or ["OAuth2 properties definition"](#oauth2-properties-definition)
 
+Implementations might impose restrictions on the expressions that might be used to resolve auth properties.
```
ricardozanini, line 3558:
```suggestion
Implementations might impose restrictions on the expressions used to resolve auth properties depending on the context used. 
```

I think we need to be more specific here. For example, we could argue that an implementation could restrict some expressions for security reasons. I'm on the fence; to me, this is still implementation-specific. Not sure if we need to explicitly have this here.

fjtirado: I think is ok to mention in spec those places where there is implementor discrection.

specification.md
```diff
@@ -3555,12 +3555,14 @@ The `scheme` property defines the auth scheme to be used. Can be "bearer", "basi
...
 
+Implementations might impose restrictions on the expressions supported to resolve auth properties, depending on the context used. 
+
 ###### Basic Properties Definition
 
 See [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#basic_authentication_scheme) for more information about Basic Authentication scheme.
 
 The Basic properties definition can have two types, either `string` or `object`. 
-If `string` type, it defines a workflow expression referencing a [workflow secret](#workflow-secrets) that contains all needed Basic auth information.
+If `string` type, it defines a workflow expression that contains all needed Basic auth information.
```
tsurdilo, line 3565: Please link "workflow expression" to https://github.com/serverlessworkflow/specification/blob/main/specification.md#workflow-expressions

specification.md
```diff
@@ -3555,12 +3555,14 @@ The `scheme` property defines the auth scheme to be used. Can be "bearer", "basi
 The `properties` property defines the auth scheme properties information.
 Can be one of ["Basic properties definition"](#basic-properties-definition), ["Bearer properties definition"](#bearer-properties-definition), or ["OAuth2 properties definition"](#oauth2-properties-definition)
 
+Implementations might impose restrictions on the expressions supported to resolve auth properties, depending on the context used.
```
tsurdilo, line 3558: I think this sentence might be more confusing to reader than help. Think this is best to be added if needed in the runtime impl docs.

fjtirado: This is part of the effort do identify those constructs, like name. version or so one, where expressions are not applicable. 
In that case, for auth definition, expression support might be limited, up to the implementor.

cdavernas: @fjtirado Now that I think of it, I guess it therefore means you'd need to add that comment for `functionDef` `operation` property, which can as well be a runtime expression.

fjtirado: yes, in general, support for expressions in function def should be implementor choice, in my opinion.

fjtirado: so, we can move this sentence from hre to function definition and that will be it, wdty?

tsurdilo: i dont think operation should allow to be expression

tsurdilo: it introduces all kinds of problems

tsurdilo: of you want different uris to end points like in dev env vs prod then use different external files in each env

cdavernas: > i dont think operation should allow to be expression

@tsurdilo We discussed that a long time ago, and you had the opposite take on it. As you said, it's a required feature to have in order to allow different configuration for different environments, resolved at runtime.

Does not change much for me, Synapse supports it anyways, but I think it's a must have feature for real life use cases

fjtirado: as you can guess, I do not think either operation or auth definitions used for downloading the operation definition should be expressions. 
The only reason I open the discussion is to avoid hardcoding password in auth definition (using $SECRET there)

ricardozanini: I agree with @tsurdilo. Operations should be immutable, IMO. This kind of configuration is a matter of infrastructure configuration, not business domain. If we allow this, it will open a huge can of worms.

tsurdilo: >  If we allow this, it will open a huge can of worms.

yes, and if possible (let me know) id like that can of works to be on the runtime impl side :)

fjtirado: To me, conceptually at least, auth definition is part of the operation definition that Ricardo and Tiho think should be "inmutable". Thats why I was reluctant to accept "indiscrimanted" support for expressions in auth def, but, after listening Charles thought, I now think is a grey area. 
We can say nothing and let implementation decide. So if everybody agree Ill remove this sentence and just keep the other modification.

tsurdilo (approved):

cdavernas (approved):
LGTM! Cheers!

tsurdilo (approved):

PR #673, opened 2022-09-12
[WIP] Update to Error Handling

tsurdilo:
Updates to error handling

* need to update examples and comparisons and go over spec doc.*
* need to add sample that shows action error handling.

rendered error handling section https://github.com/tsurdilo/specification/blob/uperrhandling/specification.md#workflow-error-handling

error data schema: in pr and can look and try it out here: https://github.com/tsurdilo/specification/blob/uperrhandling/schema/errordata.json


cdavernas (changes requested):
Awesome work, thank you!
I think there are however some conceptual changes I believe we should work on. I created a gist for my counter proposal (schemas will follow in case we reach a consensus):

https://gist.github.com/cdavernas/f482d8e82819decf2d330e98416d9ff4

ricardozanini:
Let my thoughts be in the gist.

Issue #611, opened 2022-03-31
Support for output data schema


EnriqueL8:
**What would you like to be added**:
Currently when defining a new workflow one can specify `dataInputSchema` which we can use to validate the input. Yet a user starting a workflow instance does not know what the output of this workflow is. What is the expected to get when the workflow finishes? This is primarily useful for HTTP requests.

I propose in the same we have `dataInputSchema` to have a `dataOutputSchema` which accepts a JSON schema.

**Why is this needed**:
This will make it easier to document the outputs of workflow for example in an OpenAPI document as the response from the workflow runtime. This makes it especially useful in a microservice architecture where someone can look at the OpenAPI and know what to expect. 

I'm not sure how this fits in with output CloudEvents? Ideally, this output doesn't have to be a CloudEvent.

cdavernas:
@EnriqueL8 This is a good suggestion IMO. However, we should decide if that's for informational purpose only or not.
As a matter of fact, such a schema could be used to verify that the output of a workflow instance matches what's expected just before a successfull completion.

EnriqueL8:
Good point, I hadn't thought of it as a validation step but now that you say it it does feel natural for it to be.

fjtirado:
I vote for same functionality than dataInputSchema, thats it, by default the output is validated againts the output shema if present

tsurdilo:
> Yet a user starting a workflow instance does not know what the output of this workflow is. What is the expected to get when the workflow finishes?

A JSON result (state data output of last workflow state) or a failure (thats dependent on the runtime).

> This is primarily useful for HTTP requests.

If you are exposing workflows from rest endpoint, think you could describe your workflow using OpenAPI which does include response schema. Similarly for event-triggered workflow you can use AsyncApi. 

Trying to figure out what's the purpose of this request as if we add this it can go pretty deep into what all can be added that other frameworks have already.

fjtirado:
@tsurdilo I agree with your concerns, but for these use cases, you just avoid specifying an output schema. I mean, the same than inputSchema, the new field will be optional.

cdavernas:
I'm gonna have a spec binge next week. Should I work on that PR too guys? @tsurdilo @ricardozanini @fjtirado 

I'm a big fan of the idea, personally speaking!

PR #696, opened 2022-10-10
outputDataSchema addition

fjtirado:
This adds possibility of validating workflow output in a similar way that is already done for workflow input

See issue https://github.com/serverlessworkflow/specification/issues/611

Signed-off-by: Francisco Javier Tirado Sarti <ftirados@redhat.com>

- [x] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

**Discussion or Issue link**:

**Additional information:**

### DIFF
--- a/roadmap/README.md
+++ b/roadmap/README.md
@@ -45,6 +45,7 @@ _Status description:_
 | ✏️️| Add SDKs for more languages (Python, PHP, Rust, etc) |   |
 | ✏️️| Add more samples  |   |
 | ✏️️| Enforce SemVer `version`  |   |
+| ✏️️| Add `dataOutputSchema`  |   |
 
 ## <a name="v08"></a> Released version v0.8
 
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -39,80 +39,10 @@
       "additionalItems": false
     },
     "inputDataSchema": {
-      "oneOf": [
-        {
-          "type": "string",
-          "description": "URI of the JSON Schema used to validate the workflow data input",
-          "minLength": 1
-        },
-        {
-          "type": "object",
-          "description": "Workflow data input schema definition",
-          "properties": {
-            "schema": {
-              "oneof": [
-                {
-                  "type": "string",
-                  "description": "URI of the JSON Schema used to validate the workflow data input",
-                  "minLength": 1
-                },
-                {
-                  "type": "object",
-                  "description": "The JSON Schema object used to validate the workflow data input",
-                  "$schema": "http://json-schema.org/draft-07/schema#"
-                }
-              ]
-            },
-            "failOnValidationErrors": {
-              "type": "boolean",
-              "default": true,
-              "description": "Determines if workflow execution should continue if there are validation errors"
-            }
-          },
-          "additionalProperties": false,
-          "required": [
-            "schema"
-          ]
-        }
-      ]
+      "$ref": "#/definitions/validationSchema"
     },
     "outputDataSchema": {
-      "oneOf": [
-        {
-          "type": "string",
-          "description": "URI of the JSON Schema used to validate the workflow data output",
-          "minLength": 1
-        },
-        {
-          "type": "object",
-          "description": "Workflow data output schema definition",
-          "properties": {
-            "schema": {
-              "oneof":[
-                {
-                  "type": "string",
-                  "description": "URI of the JSON Schema used to validate the workflow data output",
-                  "minLength": 1
-                },
-                {
-                  "type": "object",
-                  "description": "The JSON Schema object used to validate the workflow data output",
-                  "$schema": "http://json-schema.org/draft-07/schema#"
-                }
-              ]
-            },
-            "failOnValidationErrors": {
-              "type": "boolean",
-              "default": true,
-              "description": "Determines if an exception should be thrown when the output model does not match the provided schema"
-            }
-          },
-          "additionalProperties": false,
-          "required": [
-            "schema"
-          ]
-        }
-      ]
+      "$ref": "#/definitions/validationSchema"
     },
     "secrets": {
       "$ref": "secrets.json#/secrets"
@@ -1937,6 +1867,44 @@
       },
       "additionalProperties": false,
       "required": []
+    },
+    "validationSchema" :  {
+      "oneOf": [
+        {
+          "type": "string",
+          "description": "URI of the JSON Schema used to validate the workflow",
+          "minLength": 1
+        },
+        {
+          "type": "object",
+          "description": "Workflow data input schema definition",
+          "properties": {
+            "schema": {
+              "oneof":[
+                {
+                  "type": "string",
+                  "description": "URI of the JSON Schema used to validate the workflow",
+                  "minLength": 1
+                },
+                {
+                  "type": "object",
+                  "description": "The JSON Schema object used to validate the workflow",
+                  "$schema": "http://json-schema.org/draft-07/schema#"
+                }
+              ]
+            },
+            "failOnValidationErrors": {
+              "type": "boolean",
+              "default": true,
+              "description": "Determines if error should be thrown if there are validation errors"
+            }
+          },
+          "additionalProperties": false,
+          "required": [
+            "schema"
+          ]
+        }
+      ]
     }
   }
 }
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -38,7 +38,7 @@
       },
       "additionalItems": false
     },
-    "dataInputSchema": {
+    "inputDataSchema": {
       "oneOf": [
         {
           "type": "string",
@@ -76,7 +76,7 @@
         }
       ]
     },
-    "dataOutputSchema": {
+    "outputDataSchema": {
       "oneOf": [
         {
           "type": "string",
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -76,6 +76,44 @@
         }
       ]
     },
+    "dataOutputSchema": {
+      "oneOf": [
+        {
+          "type": "string",
+          "description": "URI of the JSON Schema used to validate the workflow data output",
+          "minLength": 1
+        },
+        {
+          "type": "object",
+          "description": "Workflow data output schema definition",
+          "properties": {
+            "schema": {
+              "oneof":[
+                {
+                  "type": "string",
+                  "description": "URI of the JSON Schema used to validate the workflow data output",
+                  "minLength": 1
+                },
+                {
+                  "type": "object",
+                  "description": "The JSON Schema object used to validate the workflow data output",
+                  "$schema": "http://json-schema.org/draft-07/schema#"
+                }
+              ]
+            },
+            "failOnValidationErrors": {
+              "type": "boolean",
+              "default": true,
+              "description": "Determines if an exception should be thrown when the output model does not match the provided schema"
+            }
+          },
+          "additionalProperties": false,
+          "required": [
+            "schema"
+          ]
+        }
+      ]
+    },
     "secrets": {
       "$ref": "secrets.json#/secrets"
     },
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -38,10 +38,10 @@
       },
       "additionalItems": false
     },
-    "inputDataSchema": {
+    "dataInputSchema": {
       "$ref": "#/definitions/validationSchema"
     },
-    "outputDataSchema": {
+    "dataOutputSchema": {
       "$ref": "#/definitions/validationSchema"
     },
     "secrets": {
--- a/specification.md
+++ b/specification.md
@@ -1769,6 +1769,7 @@ definition "id" must be a constant value.
 | version | Workflow version. MUST respect the [semantic versioning](https://semver.org/) format | string | no |
 | annotations | List of helpful terms describing the workflows intended purpose, subject areas, or other important qualities | array | no |
 | dataInputSchema | Used to validate the workflow data input against a defined JSON Schema| string or object | no |
+| dataOutputSchema | Used to validate the workflow data output against a defined JSON Schema| string or object | no |
 | [constants](#Workflow-Constants) | Workflow constants | string or object | no |
 | [secrets](#Workflow-Secrets) | Workflow secrets | string or array | no |
 | [start](#Start-Definition) | Workflow start definition | string or object | no |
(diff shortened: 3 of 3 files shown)


tsurdilo:
There is also question on what happens on failure to validate subflow result as this can be defined in the subflow workflow definition, so this is tied to the error handling still WIP pr. Imo they should be committed at the same time if we decided to go w/ this.

tsurdilo:
related to https://github.com/serverlessworkflow/specification/pull/673, please do not merge one without the other

fjtirado:
@tsurdilo I opened this PR because reading  @cdavernas  I believe this feature was accepted. 
In my view, the output validation should be performed only for parent process (not subprocess) and if it fails, it behave as when input validation fails: it throws an exception if the failOnValidation is true. 
Since it only applies to parent process, we can consider it is is not part of error handling feature, because it does not really belong to flow execution. Thats why I just change the dataInputShema bits to add dataOuputSchema.

tsurdilo:
Yeah definitely lets add it, just would be nice to update the little things mentioned. Ty!!

fjtirado:
> 

@tsurdilo Sure, Ill do, thanks

fjtirado:
@tsurdilo I tried to reflect your concerns into the text.

specification.md
```diff
@@ -1859,9 +1860,21 @@ The `version` property can be used to provide a specific workflow version. It mu
...
 
-The `dataInputSchema` property can be used to validate the workflow data input against a defined JSON Schema.
-This check should be done before any states are executed. `dataInputSchema` can have two different types.
-If it is an object type it has the following definition:
+The `dataInputSchema` and `dataOutputSchema` properties can be used to validate model against a defined JSON Schema.
+
+The `dataInputSchema` property validates the [workflow data input](#Workflow-Data-Input). Validation should be performed before any states are executed. In case of
+a starting [Event state](#Event-state), it is not used to validate its event payloads. The `failOnValidationErrors` property  determines if workflow execution should continue in case of validation errors. 
+
+The `dataOutputSchema` property validates the [Workflow data output](#workflow-data-output). Validation should be performed after workflow execution has finished successfully. Successfully means the workflow has completed an end state without errors.
```
cdavernas, line 1868: I would instead write something like:

Successfully means that the workflow has completed an end state without errors

cdavernas (changes requested):
Please consider #705 before merging, as requested change might be included in this PR as well.

fjtirado:
@cdavernas Comments incorporated

cdavernas (approved):
Looks great! Many thanks!

cdavernas (approved):

ricardozanini (approved):

schema/workflow.json
```diff
@@ -76,6 +76,44 @@
...
+                  "type": "object",
+                  "description": "The JSON Schema object used to validate the workflow data output",
+                  "$schema": "http://json-schema.org/draft-07/schema#"
+                }
+              ]
+            },
+            "failOnValidationErrors": {
+              "type": "boolean",
+              "default": true,
+              "description": "Determines if an exception should be thrown when the output model does not match the provided schema"
```
tsurdilo, line 107: align with https://github.com/serverlessworkflow/specification/blob/main/schema/workflow.json#L69

schema/workflow.json
```diff
@@ -76,6 +76,44 @@
         }
       ]
     },
+    "outputDataSchema": {
```
tsurdilo, line 79: This seems a copy of "dataInputSchema"/ "inputDataSchema" as you changed.
Would be better imo to have a single definition called "inputSchema" and then reference it rather than having two separate

specification.md
```diff
@@ -1859,22 +1860,33 @@ The `version` property can be used to provide a specific workflow version. It mu
...
-If it is an object type it has the following definition:
+The `inputDataSchema` and `outputDataSchema` properties can be used to validate input and output data against a defined JSON Schema.
+
+The `inputDataSchema` property validates the [workflow data input](#Workflow-Data-Input). Validation should be performed before any states are executed. In case of
+a start [Event state](#Event-state) the input schema is ignored, if present. The `failOnValidationErrors` property  determines if workflow execution should continue in case of validation errors. 
+
+The `outputDataSchema` property validates the [Workflow data output](#workflow-data-output). Validation should be performed after workflow execution has finished successfully. Successfully means that the workflow has completed an end state without errors. 
+The `failOnValidationErrors` property determines what should be done when the workflow output does not match the provided schema. 
+If `failOnValidationErrors` is true, an error should be thrown. If executed within a subprocess, that error might be handled by the parent workflow.
```
tsurdilo, line 1870: dont use "might", be specific. yes it should be propagated to parent exec and handled as part of the workflow state that invoked the subflow.

specification.md
```diff
@@ -1886,22 +1898,18 @@ If it's a JSON schema object, it has the following definition:
...
-```
+If using string type, then the string value is the external schema URI and `failOnValidationErrors` default value of `true` is assumed.
 
-In this case the `failOnValidationErrors` default value of `true` is assumed.
+Example using string type
 
-The `dataInputSchema` property validates the [workflow data input](#Workflow-Data-Input). In case of
-a starting [Event state](#Event-state), it is not used to validate its event payloads.
+```json
+"inputDataSchema": "URL_to_json_schema"
```
tsurdilo, line 1911: think it should be URI, not URL

specification.md
tsurdilo, line 1870: "thrown" is kinda java-ish. pls use "raised"

tsurdilo (changes requested):
Overall lgtm. Left some comments that should be fixed pls

schema/workflow.json
cdavernas: @tsurdilo I would recommend calling that definition `dataValidation` or whatever, as `inputSchema` would be confusing when used for output 😜

specification.md
cdavernas: Agreed!

specification.md
cdavernas: Totally disagree here! Errors are not raised, they are thrown, it's not java-ish, it's IT-ish and is the proper terminology in that case imo

cdavernas: As we are speaking of error control flow, the terminology `catch` is actually more appropriate. 
As a matter of fact, raising an error is the right terminology when "producing" the error.

Btw, `throw`, `catch` and other keywords of the like are NOT specific to Java or a specific grammar. Those are specific IT terms, very widely and globally used, across languages and techs, to describe, in this case, error control flow.  It is not a matter of opinion, it's a matter of proper vocabulary. Making proper use of that vocabulary ensures ubiquitousness. Using synonyms for the sake of not looking like code is  IMHO very dangerous, misleading to the biggest part of our community (devs mostly) and is even confusing.

tsurdilo: there is no throw and catch in Go :)

schema/workflow.json
tsurdilo: maybe inputOutputValidation then cause we are not validating like event payloads or action results or state data which are all data

specification.md
cdavernas: @tsurdilo indeed, because there is no proper error handling out of the box. 
 https://link.medium.com/nkJbfoMHuvb

Doesn't mean that because they do not provide a core feature that said feature is wrongly named 😜

schema/workflow.json
cdavernas: @tsurdilo Point taken! However, I personally dislike that term too. I'm sure we can, however, find a good one. I'll sleep on it.

specification.md
tsurdilo: | They are basic IT terms

yes exactly, thats why we should if possible stay away from using them inside spec text. DSLs are to be read and understood by non-IT people too

cdavernas: But dude, catch and throw are less misleading that synonyms that do not apply to the concepts.
Again, it's not because something is rightfully used by global consensus in IT that we should not use it and confuse most ppl. 
Those terms apply perfectly, are easily understood by even the dumbest guy around, and avoids guaranteed misinterpretation.

tsurdilo: It would be also easier to use "return error" and "check error" and "panic" and "unwrap" too but i guess this maybe too much. So idk you seem to have a strong opinion about this so not going to argue with you, but don't think try/catch is really that "mainstream" outside devel world maybe

cdavernas: Well, I do have a very strong opinion on that matter, but that's just me, and I'm indeed a bit of a grammar nazy 😜 
Let's see what the others think about it, and let's settle this out in a weekly. 
Wdyt?

cdavernas: Suggested terms are all far fetched imho, and do not really paint the 'action' we want to describe, again imho.

tsurdilo: Sure, lets try that and catch any issues that come up ;)

cdavernas: Jokes appart, I hear you and I do believe you make a very valid point by always thinking of not turning the spec into some kind of IT only language. 
I just think that in this specific case, as the grammar nazy I must admit I am, alternatives sounds a bit far fetched whereas the try/catch/throw one is beautifully simple and ubiquitous. 
Let's see what the others think and settle that out in a weekly, wdyt?

fjtirado: I can use "generated" rather than "thrown", is that ok for you guys?

schema/workflow.json
fjtirado: I think we can use "validationSchema" for the data stucture used to validate schema

specification.md
cdavernas: I don't think that's right either. I like the thrown term. But whathever, it's just a minor semantic detail, so I'll go with whatever you guys settle on 😉

schema/workflow.json
cdavernas: I like that, good idea!

specification.md
cdavernas: I agree with @fjtirado on this one!

fjtirado: I do not have particular interest on discussing a semantic thing, so Tiho let us know which is your chosen word. 
If it helps, thow was not invented by Java, it was chosen for C++. And Java also includes (inherited from other languages) english words like if, else, finally, int, float, double, string, array... ;)

ricardozanini: I'd rather stay with the ubiquitous term, which are try/catch/throw, widely used by the IT community. Raised is yet another good term, thou. But in this context, throw is the best one. And just to try to add a period in this conversation: https://english.stackexchange.com/questions/319126/why-exactly-is-the-verb-to-throw-used-when-an-error-occurs-in-the-program.

fjtirado:
@tsurdilo Comments incorporated, please take a look

cdavernas:
@tsurdilo Can you please review requested changes?

cdavernas:
Linked to https://github.com/serverlessworkflow/specification/issues/705

ricardozanini:
@tsurdilo can you please take a final look so we can merge it? Thank you!

specification.md
```diff
@@ -1859,22 +1860,33 @@ The `version` property can be used to provide a specific workflow version. It mu
...
 
-The `dataInputSchema` property can be used to validate the workflow data input against a defined JSON Schema.
-This check should be done before any states are executed. `dataInputSchema` can have two different types.
-If it is an object type it has the following definition:
+The `inputDataSchema` and `outputDataSchema` properties can be used to validate input and output data against a defined JSON Schema.
+
+The `inputDataSchema` property validates the [workflow data input](#Workflow-Data-Input). Validation should be performed before any states are executed. In case of
+a start [Event state](#Event-state) the input schema is ignored, if present. The `failOnValidationErrors` property  determines if workflow execution should continue in case of validation errors. 
+
+The `outputDataSchema` property validates the [Workflow data output](#workflow-data-output). Validation should be performed after workflow execution has finished successfully. Successfully means that the workflow has completed an end state without errors.
```
tsurdilo, line 1868: I think this sentence is confusing as in "completed an end state without errors" I think this should just say

"It's performed on the output of the workflow execution"..leave it at that imho

specification.md
tsurdilo, line 1870: please change "might" to "can be" or "should be"

specification.md
```diff
@@ -1859,22 +1860,33 @@ The `version` property can be used to provide a specific workflow version. It mu
...
+
+The `inputDataSchema` property validates the [workflow data input](#Workflow-Data-Input). Validation should be performed before any states are executed. In case of
+a start [Event state](#Event-state) the input schema is ignored, if present. The `failOnValidationErrors` property  determines if workflow execution should continue in case of validation errors. 
+
+The `outputDataSchema` property validates the [Workflow data output](#workflow-data-output). Validation should be performed after workflow execution has finished successfully. Successfully means that the workflow has completed an end state without errors. 
+The `failOnValidationErrors` property determines what should be done when the workflow output does not match the provided schema. 
+If `failOnValidationErrors` is true, an error should be thrown. If executed within a subprocess, that error might be handled by the parent workflow. 
+If `failOnValidationErrors` is false, it is up to the implementor to warn the user about that fact using any mean, except throwing an error. For example, printing a log. 
+
```
tsurdilo, line 1872: "setting failiOnvalidationError to false to not throw error even in case of validation errors"

i think would sound better

tsurdilo:
left couple of small change requests for the docs part hope thats ok

specification.md
fjtirado: Done

specification.md
fjtirado: Done

tsurdilo:
I see, i think both says they are the same, they are both schemas one defines data input and the other data output of the workflow exec

ricardozanini:
@tsurdilo @cdavernas @fjtirado I'd rather stay with `dataInputSchema` and `dataOutputschema` if you don't mind. @tsurdilo I think all your comments have been handled. Can you take a final look/approve so we can merge this?

cdavernas:
> @tsurdilo @cdavernas @fjtirado I'd rather stay with `inputDataSchema` and `outputDataSchema` if you don't mind.

Works for me either way! I guess it's just a matter of point of view/personal prefs anyways

ricardozanini:
@cdavernas I agree. Either works for me too if we can move forward!

fjtirado:
@tsurdilo Any further request?

roadmap/README.md
```diff
@@ -45,6 +45,7 @@ _Status description:_
 | ✏️️| Add SDKs for more languages (Python, PHP, Rust, etc) |   |
 | ✏️️| Add more samples  |   |
 | ✏️️| Enforce SemVer `version`  |   |
+| ✏️️| Add `dataOutputSchema`  |   |
```
tsurdilo, line 48: so which one is it? dataOutputSchema as requested or inputDataSchema as in the change you made to workflow.json :)

fjtirado: If I revert to dataInputSchema, I guess thats fine isnt it?

tsurdilo: imho yes, but depends if you and others want to do that change. either way after you guys decide 
please make sure that its reflected in readme.md with the same names you use in workflow schema

fjtirado: @tsurdilo   I revert to the original form, dataInputSchema and dataOutputSchema so it is aligned with readme roapmap.

tsurdilo (approved):

ricardozanini:
@cdavernas any other comments?

cdavernas:
@ricardozanin Nope, looking great to me! Thanks @fjtirado

Issue #691, opened 2022-09-30
Refactor external references


cdavernas:
**What would you like to be added**:

Refactor external references.

What I propose is:

1. To define a `resource` object, such as the following, used to encapsulate external definitions:

```yaml
auth:
  - name: foo_basic
    scheme: basic
    properties:
      username: username
      password: password
functions:
  - name: secured_function
    type: rest
    operation: https://foo.bar#baz
    authRef: foo_basic
```

In other words, an external function definition file would not contain an array of function definitions, but an object such as the above.

In addition, that would allow serverless-workflow compatible apps to provide a self-describing document making it ready to use in Serverless Workflow.

2. Instead of having `functions`, `events`, etc. be of type `oneOf<uri, functionArray>`, they should be only of type `array`, as discussed in the meeting of the 27th of September, and in https://github.com/serverlessworkflow/specification/issues/676.
A new `imports` (or `resources`) top-level properties would replace the need for uri-values, and would allow the import of multiple files instead of just one.

```yaml
name: foobar
version: 0.1.0
specVersion: 0.9
imports:
  - namespace: foo
    uri: https://foobar.baz/sw.res
  - namespace: bar
    uri: https://bazfoo.bar/sw.res
states:
  - name: invoke_external_functions
    type: operation
    actions:
      - name: invoke_foo
        functionRef:
          refName: foo.greet
          arguments:
            greetings: Hello, Serverless Workflow!
     - name: invoke_bar
        functionRef:
          refName: bar.greet
          arguments:
            greetings: Hello, dear readers!
    end: true
```

**Why is this needed**:

Currently, external references are easy and simple to use, but unhappilly do not address all common/reasonnable use cases, as explained in https://github.com/serverlessworkflow/specification/issues/676. In addition, it forbids referencing in a safe manner external functions that define an `authRef`

tsurdilo:
In your example where is the refName: foo.greet function definition? is the function def named "greet" and you append the namespace to it?

cdavernas:
@tsurdilo yes, exactly: `{namespace}.{functionName}`

tsurdilo:
Thanks, yeah i put a comment on that associated issue..do really like it but then why not

     {namespace}.{stateName}
     {namespace}.{actionName}
     {namespace}.{eventDefName}
     ...

I think these type of changes are awesome but in the grand schema of things if added should be thought to be added to entire spec not just one piece at a time imo

tsurdilo:
[...]

PR #701, opened 2022-10-24
Adds support for external resources

cdavernas:

- [x] Specification
- [x] Schema
- [x] Examples
- [ ] Extensions
- [x] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

**Discussion or Issue link**:

https://github.com/serverlessworkflow/specification/issues/691

Adds support for external resources, and removes the ability to set an uri for `auth`, `constants`, `retries`, `secrets`, `functions`, `events`, ...


ricardozanini (commented):
Seems great! Just two minor comments.

specification.md
```diff
@@ -1769,19 +1769,20 @@ definition "id" must be a constant value.
...
-| [functions](#Function-Definition) | Workflow function definitions. Can be either inline function definitions (if array) or URI pointing to a resource containing json/yaml function definitions (if string) | array or string| no |
+| [auth](#Auth-Definition) | Workflow authentication definitions | array | no |
+| [events](#Event-Definition) | Workflow event definitions.  | array | no |
+| [functions](#Function-Definition) | Workflow function definitions. Can be either inline function definitions (if array) or URI pointing to a resource containing json/yaml function definitions (if string) | array | no |
 | autoRetries | If set to true, [actions](#Action-Definition) should automatically be retried on unchecked errors. Default is false | boolean| no |
-| [retries](#Retry-Definition) | Workflow retries definitions. Can be either inline retries definitions (if array) or URI pointing to a resource containing json/yaml retry definitions (if string) | array or string| no |
+| [retries](#Retry-Definition) | Workflow retries definitions. Can be either inline retries definitions (if array) or URI pointing to a resource containing json/yaml retry definitions (if string) | array | no |
```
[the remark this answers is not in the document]
cdavernas: @ricardozanini Should be resolved ;)

tsurdilo:
thanks for the pr. think we need to still update roadmap.md and add a full sample in our examples.md for this please.
would be nice to have more of a description for this feature as its quite a change.

one thing I'm not a big fan of is how much "resources" abstracts from the workflow definition itself. looking at a wf def that contains one ore more resources i get really confused as to what these resources contain and how i can reference these things. I think it would be better (yes more verbose but still) to add resources per type, so for example:


        "functions": {
          "resources": [
            
            ],
          "inline": [
            
            ]
        },
        "timeouts": {
          "resources": [
            
            ],
          "inline": [
            
            ]
        },
        "errors": {
          "resources": [
            
            ],
          "inline": [
            
            ]
        }

(change "inline" to whatever you want to call it). 
Imo this is much easier to manage from user point of view and to understand whats going on by looking at the dsl. wdyt?

cdavernas:
@tsurdilo I personally fond that ugly and counterintuitive and not ubiquitous. Resources are external, and you shouldn't pay attention to them in most cases.

tsurdilo:
if in the future we add a new definitions array like functions, timeouts etc
do we say users have to update their existing reaource files? they might have prod workflows running and cannot update them or re-validate. what would u recommend in that situation? i see the benefit od this just trying to see if we can define it differently maybe

tsurdilo:
for the inline defe i showed you can point to same resources file in each def so could keep the single resource json that contains all

cdavernas:
> if in the future we add a new definitions array like functions, timeouts etc
do we say users have to update their existing reaource files?

Users should be able to do whatever they want in that case. They can append new definitions, or just leave it like it was: there'd be nothing broken anyways. In case of specVersion mismatch though, there's a problem, but that applies to any reference anyways. This should be a runtime concern only.

cdavernas:
> for the inline defe i showed you can point to same resources file in each def so could keep the single resource json that contains all. just this way it would be clear what that resource contains

What you suggest is not clear at all IMHO and forces the users to uselessly copy the same external resource file amongst the multiple resources that might need it. This would make it painfull for everyone.

Again, the PR is the *exact* result of what was discussed and approved, I don't see the need to go over it once again.
The proposal is clear, ubiquitous and clean, and the alternatives you propose are not IMO, and does not fully address requirements.

As for your actual concern which is having the same concepts in two different places in the spec, I'd say that there is no duplicates: on one hand you have definitions of **external resources** (which, again, most users won't even pay attention to), and on the other hand you have the perfectly fine `functions`, `events` and whatnot properties. Those are ubiquitous, clean and straightforward and should IMO not be changed in any ways (aside from forbidding non-inline values)

cdavernas:
> just this way it would be clear what that resource contains

You have no control, spec-side, on what an external resource may or may not contain. At best, you have an idea at a certain point in time, thus the named refs, but that's about it. Why would use ask authors of workflows to duplicate the work that has already been **declaratively** done in the external resource? What's the gain, if any?

fjtirado (approved):
@cdavernas 
Thanks a lot
Thats what we agreed in the last meeting and you reflected it perfectly in this PR. 
I do not think we need to repopen a discussion that was apparently closed @tsurdilo 
This covers a gap that we had in the current spec with minimal changes
Additionally, this removed the difficulties associate with one of. 
And it is easily extensible if a new resource is added

tsurdilo:
> I do not think we need to repopen a discussion that was apparently closed

I dont think its ever too late to question any decisions with this project by anyone. Without questioning whats there now we would not need to make any more changes in that case right, even for this pr, right ? :) 

This spec is what it is because we discussed things over and over and over again to make it the best we can and this should continue imo. 

We do not have a timeline set for 0.9 and there is no rush that i can see in order not to have time or opportunity to discuss this, especially if its project maintainers that would like an open discussion about their opinions.....

fjtirado:
@tsurdilo 
I could agree with you, if we have not invested a complete meeting discussing this and agree on this approach after more than hour of discussion.
In my opinion, once an agreement is reached, it should be respected by all the parts involved on it. 
You comment about splitting the existing properties into two rather than using a new property was discussed during that meeting and it was decided (or that was my understanding) to go on with the separate property approach. . 
Re-discussing this particular aspect again, which of course you are doing with the best of your intentions,  can be be intepreted as breaking such agreement. It would be different if we were discussing small details, but thats not the case.

tsurdilo:
I'm not sure what you are trying to say but one thing is a discussion, another is reviewing a PR. Once you see the actual proposed impl in a pr and try to test it, questions can come up.

I would like to keep the discussions here related to the PR itself, anything else feel free to bring up in weekly meets or offline. I'm not sure how you understand open source works honestly but whatever, ty for your comments.

ricardozanini:
I agree that we need to keep discussing the issues as they appear in a PR that can change the perspective. 

Honestly, I don't see any changes that go against what we've discussed, and I think it's reasonable to raise questions.

Having said that, I believe the discussed plan still holds. As proposed, the `resources` attribute can reasonably solve this problem, whereas adding it to each definition can make it more lenghty and hard to maintain.

fjtirado:
> I'm not sure what you are trying to say but one thing is a discussion, another is reviewing a PR. Once you see the actual proposed impl in a pr and try to test it, questions can come up.

What I'm trying to say is that your proposal of splitting every property that can have a resource into two were already discussed in that meeting. 
Anyway...this proposal has problems that Charles has already pointed out:
- It is too verbose and it is a lot of copy paste (I hope we agree copy paste is an issue) in the spec schema. 
- It will force the user to have an external resource for every different type, making impossible to define one external resource property for every API you want to integrate. The main point of importing extenal resource is to avoid declaring things in several fiels accesing a certain API, but if, in order to invoke that api, you spread all the knowledge in a lot of files, you come back to the original issue (you are not declaring twice, but you have knowledge distributed into several files) 
But my main concern is that I hardly see any advantage over Charles proposal. Charles proposal allows you to put every resource type into a different file (I do not recommend that though) and you can even put the resource type in your file name. So if a user has the issue you mention to suport your proposal, he can still fufill it with Charles approach. On the contrary, if we took your approach, there is no way the user can put all the resources related with an API into the same file. So, Charles proposal is more flexible.

schema/resources.json
```diff
@@ -0,0 +1,100 @@
...
+    ],
+    "definitions":{
+        "resourceRef":{
+            "type": "object",
+            "description": "References an external resource",
+            "properties": {
+              "name": {
+                "type": "string",
+                "description": "The name of the external resource. It must be lowercased and must only contain alphanumeric characters, with the exception of `-` and `_`. The name is used as prefix to reference any definition contained in the referenced resource.",
+                "pattern": "^[a-z0-9-_]+$"
```
tsurdilo, line 27: Can we please remove this pattern for now, and then apply a pattern across all spec props used for lookups.

schema/resources.json
```diff
@@ -0,0 +1,100 @@
...
+                "type": "string",
+                "description": "References the (inline) auth definition used to retrieve the referenced external resource"
+              }
+            },
+            "required": [
+                "name",
+                "uri"
+            ]
+        },
+        "resourceDef":{
```
tsurdilo, line 44: please externalize this into own schema so it can be validated individually

schema/resources.json
```diff
@@ -0,0 +1,100 @@
...
+                  "$ref": "functions.json#/functions"
+                }
+              },
+              "errors": {
+                "type": "array",
+                "items": {
+                  "$ref": "errors.json#/errors"
+                }
+              },
+              "metadata":{
```
tsurdilo, line 93: Please add constants here.

tsurdilo, line 93: add constants and possibly secrets

specification.md
```diff
@@ -5989,6 +5990,60 @@ There are two places in the [workflow definition](#Workflow-Definition-Structure
 
 The `version` property must respect the [semantic versioning](https://semver.org/) guidelines.
 
+### Workflow Resources
```
tsurdilo, line 5993: * Add example 
* Add roadmap 
* Check how we can use this in expressions that call expression functions, for example 
"condition": "${ fn:abc.isMinor }",

ricardozanini (approved):

tsurdilo:
Please add entry to roadmap and if possible an example. Thnks.

specification.md
tsurdilo: Let me know when these things are added so i can review again.

tsurdilo (dismissed):
https://github.com/serverlessworkflow/specification/pull/701#discussion_r1007145246

let me know when to review again

specification.md
cdavernas: Will be done for tomorrow's meeting 😉

schema/resources.json
cdavernas: This PR is now pending merge of #768, so I'll fix resources to match the pattern specified there

schema/resources.json
cdavernas: Constants, I can add it if you think it's worth it. Secrets, however, sounds to me like a good bad idea. IMHO, secrets should only be explicitly granted to a single workflow (as opposed to all referencing workflows): not only would it be extremely challenging to implement (secrets usually need to be "mounted" somehow and made available to consuming apps) but would pose possible security issues.
@ricardozanini WDYT?

examples/README.md
```diff
@@ -4739,3 +4740,141 @@ functions:
 </td>
 </tr>
 </table>
+
+### Import External Resources
+
+#### Description
+
+For the example, let's assume we have developped a number of workflows to orchestrate an hypothetical, event-driven management solution around the Swagger Pet Store API.
```
ElanHasson, line 4748:
```suggestion
For the example, let's assume we have developed a number of workflows to orchestrate an hypothetical, event-driven management solution around the Swagger Pet Store API.
```

cdavernas:
@ricardozanini Bump

cdavernas:
Not surprisingly: stale after sitting for about 2 years

ricardozanini:
@cdavernas can you fix the schemas/example CI? So we can merge?

ricardozanini:
Closed as part of https://github.com/serverlessworkflow/specification/issues/843

PR #703, opened 2022-10-25
fix(spec): remove timeouts property description from inject state 

lsytj0413:
Signed-off-by: lsytj0413 <511121939@qq.com>

- [x] Specification

**Discussion or Issue link**:

- Inject state's timeouts field is missing in table

**Additional information:**

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -2834,10 +2834,6 @@ ${ {people: [.people[] | select(.age >= 40)]} }
 
 This allows you to test if your workflow behaves properly for cases when there are people whose age is greater or equal 40.
 
-The `timeouts` property can be used to define state specific timeout settings. Inject states can define the
-`stateExecTimeout` property. For more information on workflow timeouts reference the
-[Workflow Timeouts](#Workflow-Timeouts) section.
-
 Note that `transition` and `end` properties are mutually exclusive, meaning that you cannot define both of them at the same time.
 
 ##### ForEach State


cdavernas:
@lsytj0413 I believe that's because in all logic `InjectState` cannot define a timeout: either the injection happens, or an evaluation error is thrown.

lsytj0413:
> @lsytj0413 I believe that's because in all logic `InjectState` cannot define a timeout: either the injection happens, or an evaluation error is thrown.

image

But in the specification's description, it has an timeouts field.

cdavernas:
@lsytj0413 I think it's an error. @tsurdilo @ricardozanini can you guys infirm/confirm?

If I'm right, the PR should be about removing the property from the `InjectState`, in both spec and schemas

lsytj0413:
> @lsytj0413 I think it's an error. @tsurdilo @ricardozanini can you guys infirm/confirm?
> 
> If I'm right, the PR should be about removing the property from the `InjectState`, in both spec and schemas

I found this property is removed from https://github.com/serverlessworkflow/specification/pull/545

tsurdilo:
Thanks for pr!

After thinking about this, i would actually like to remove inject state and replace it with an inject action type please.

Reason for this is that an inject operation (just like any expression eval for example) can fail, should be retryable, should have a timeout and it should be caught by error handling like other actions.

would you be willing to work on this please? so remove inject state, create inject action type and then update spec doc and all samples please. wdyt?

lsytj0413:
> Thanks for pr!
> 
> After thinking about this, i would actually like to remove inject state and replace it with an inject action type please.
> 
> Reason for this is that an inject operation (just like any expression eval for example) can fail, should be retryable, should have a timeout and it should be caught by error handling like other actions.
> 
> would you be willing to work on this please? so remove inject state, create inject action type and then update spec doc and all samples please. wdyt?

Ok，I will take this.

cdavernas:
@tsurdilo I agree that inject state should become an action. However, why do you bring retry in?. How does it make any sense for you to retry an injection? If your injection failed, retrying it is not gonna work, even after 1 billion retries: the expression needs to be fixed. If it worked, retrying it a billion times will have the exact same result than first time: you'll replace what you injected with what you injected.

tsurdilo:
good point if we say that inject errors are non retryable type of errors (if we move to actions we should defin that i think) . imo niot sure we can say that all runtimes that impl spec should guarantee injections never to fail (oom issues, runtime issues, expression issues, intermittent failures for whatever reason..) in thise cases users should be able to define at which point to stop retrying and be able to handle error and then do compensation if needed. wdyt?

lsytj0413:
@tsurdilo  done, PTAL

I have change InjectState to an action's injectRef property, you can review it first, and after the proposal is accepted, i will change all examples.

cdavernas (dismissed):

lsytj0413:
> @lsytj0413 thanks for making these updates!
> 
> We had discussion and would like for not to just to what you originally intended. Due to my comments I derailed it into something else, sorry. For now please add a new pr to remove the timeout for inject state. Thanks.

Ok, i have revert the changes to just remove timeout property for inject state, PTAL

cdavernas (approved):

ricardozanini (approved):

lsytj0413:
@tsurdilo  can this merged？

Issue #684, opened 2022-09-16
Group `End` and `Transition` properties


cdavernas:
**What would you like to be added**:
Group `End` and `Transition` properties, in such way that we can address it as an (activity | state | workflow) outcome.

In other words, what I'd like to achieve is:

```yaml
...
states:
  - name: MyState
    outcome:
      type: end
      produceEvents: ...
    ...
...
```

**Why is this needed**:

Because `End` and `Transition` properties are mutually exclusive, and implicitly defines the outcome of an (activity | state | workflow), it makes sense to call is as such. In addition, it simplifies processing of the whole definition, by checking that an outcome has been specified instead of having to do checks on multiple properties. @tsurdilo WDYT?

tsurdilo:
Yes, i wish we did this to begin with. Will update

cdavernas:
Closed as resolved by 0.10, and therefore as part of #843

Issue #677, opened 2022-09-13
Atomic produce and consume actions


cdavernas:
**What would you like to be added**:
Atomic produce and consume actions, meaning splitting the actual event action into two new actions: produceEvent and consumeEvent.

**Why is this needed**:
- Allows for consuming without producing
- Allows for producing without consuming
- Easier to understand
- More ubiquitous

ricardozanini:
@cdavernas, can you evolve a little bit your description (maybe including an example)? Not sure if I understood. We already have the `produceEventRef` and `consumeEventRef` in the `EventAction`: https://github.com/serverlessworkflow/specification/blob/main/specification.md#EventRef-Definition

cdavernas:
@ricardozanini We do, but the [produceEventRef](https://github.com/serverlessworkflow/specification/blob/main/specification.md#Event-Definition) property is required.

We could just mark that property as non-required, but then the actual objet gets confusing with different properties that do the same thing: reference an event.

Example of what I'm proposing:

```yaml
...
actions:
- name: Notify User about completion
  eventRef:
    refName: MyProducedEvent
    kind: produce # *
    data: {
       ...
    }
- name: Do something else
  functionRef: a-test-function
- name: Wait for something to happen
  eventRef:
    refName: MyConsumedEvent
    dataFilter: {
      ...
    }
...
```
* Here we could:

a. Determine the action of the eventRef by consulting the eventDefinition's `kind` property => bad for schema, as we loose that discriminator for produceEvent and consumeEvent properties, which differ

b. Remove the (IMO useless) `kind` property on the `eventDef`, and move it to the `eventRef` object => easy for schema, as it is our discriminator property. Plus, why should I have to repeat the definition of the same event if I want to both consume and produce it?

cdavernas:
I would as well propose to remove the non-ubiquitous, obsolete `invoke` property on the eventRef. There is no sync/async paradigm here, it's an abuse of language applied to a non-applicable "async function" concept.

If need be for that concept, you can achieve it yourself by assembling one produce event and one consume event actions.

tsurdilo:
Think the reason for this is the event definition "source" property. If its consumed event then this has to be set to match the source where to consume from. if it's produced then runtime has to set the source. 

so kind is "kinda" hehe, needed imo unless we say its not used when producing this event which is imo worse user experience maybe. i am not sure in how much % of cases you would consume and produce the exact same event definition, can you show example?

cdavernas:
[...]

PR #706, opened 2022-10-28
Group `End` and `Transition` properties into an `Outcome` object

cdavernas:

- [x] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap

**Discussion or Issue link**:
https://github.com/serverlessworkflow/specification/issues/684

*What?*

Group End and Transition properties, in such way that we can address it as an (activity | state | workflow) outcome.

*Why?*

Because End and Transition properties are mutually exclusive, and implicitly defines the outcome of an (activity | state | workflow), it makes sense to call is as such. In addition, it simplifies processing of the whole definition, by checking that an outcome has been specified instead of having to do checks on multiple properties.

- Easier to validate
- Simplifies workflow schema (a lot), by getting rid of OneOfs transition/end requirements
- More ubiquitous

Work In Progress. Will require thorough reviewing as this requires a ton of little changes in the specification.md, for instance.

### DIFF
--- a/schema/retries.json
+++ b/schema/retries.json
@@ -37,15 +37,15 @@
         },
         "delay": {
           "type": "string",
-          "description": "Time delay between retry attempts (ISO 8601 duration format)"
+          "description": "Time delay between retry attempts (literal ISO 8601 duration format or expression which evaluation results in an ISO 8601 duration)"
         },
         "maxDelay": {
           "type": "string",
-          "description": "Maximum time delay between retry attempts (ISO 8601 duration format)"
+          "description": "Maximum time delay between retry attempts (literal ISO 8601 duration format or expression which evaluation results in an ISO 8601 duration)"
         },
         "increment": {
           "type": "string",
-          "description": "Static value by which the delay increases during each attempt (ISO 8601 time format)"
+          "description": "Time delay increment during each attempt (literal ISO 8601 duration format or expression which evaluation results in an ISO 8601 duration)"
         },
         "multiplier": {
           "type": [
@@ -73,7 +73,7 @@
           ],
           "minimum": 0,
           "maximum": 1,
-          "description": "If float type, maximum amount of random time added or subtracted from the delay between each retry relative to total delay (between 0 and 1). If string type, absolute maximum amount of random time added or subtracted from the delay between each retry (ISO 8601 duration format)"
+          "description": "If float type, maximum amount of random time added or subtracted from the delay between each retry relative to total delay (between 0 and 1). If string type, absolute maximum amount of random time added or subtracted from the delay between each retry (literal ISO 8601 duration format or expression which evaluation results in an ISO 8601 duration)"
         }
       },
       "additionalProperties": false,
--- a/schema/timeouts.json
+++ b/schema/timeouts.json
@@ -43,7 +43,7 @@
       "oneOf": [
         {
           "type": "string",
-          "description": "Workflow execution timeout duration (ISO 8601 duration format). If not specified should be 'unlimited'",
+          "description": "Workflow execution timeout duration (literal ISO 8601 duration format or expression which evaluation results in an ISO 8601 duration). If not specified should be 'unlimited'",
           "minLength": 1
         },
         {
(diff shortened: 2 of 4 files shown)


cdavernas:
Outcome schema is incomplete as far as I can see.
In addition, two approaches are possible here:

1. Simply moving `transition` and `end` properties into `outcome`, just like I did here
2. Merging both properties, like I actually suggested in related issue

Not sure, but I'm now thinking the second way makes more sense (unhappilly 😭). @ricardozanini @tsurdilo WDYT?

cdavernas:
@ricardozanini @JBBianchi @fjtirado 

The PR's stale, I'll reword it. In the meanwhile, what do you guys prefer?

1. Move `end` and `transition` into a new object, like so:

```yaml
outcome:
  end:
     terminate: true | false
     produceEvents: {}
     compensate: true | false
  transition:
     nextState: some-state-name
     produceEvents: {}
     compensate: true | false
```

2. Merge `transition` and `end` into a `outcome` object, because they both share `produceEvents` and `compensate`. This will also allow adding features more easily:

```
outcome:
  type: end | transition
  produceEvents: {}
  compensate: true | false
  terminate: true | false #only if type `end`
  nextState: some-state-name #only if type `transition`
```

cdavernas:
Furthermore, I'd like to take the opportunity to rename the `transition`'s `nextState` property into `to`, which feels more ubiquitous and call to action like:

```yaml
transition:
  to: my-next-state
```` 

vs


```yaml
transition:
  nextState: my-next-state
```

@ricardozanini @JBBianchi @fjtirado WDYT?

fjtirado:
@cdavernas @JBBianchi 
I think we can skip` type` and rename `outcome` to `next`
Therefore, we have 

```
next: 
  to: <state name>
```
or
```
next: 
  terminate: true | false
```
And you can add `publishedEvents` and `compesate` to both of them

cdavernas:
@fjtirado I love fluent/call to action terminology too, but we should then use it accross the whole state(s) concept (instead of `actions`, use the term `do`, for instance - why I'm ok with btw). In the time being, I'd rather stick with a noun.

What about?

```yaml
outcome:
  transitionTo: my-state-name
  compensate: true
```

or

```yaml
outcome:
  end: {} #we cannot use terminate here, because end also defines continueAs
  compensate: true
```

fjtirado:
@cdavernas ok, lets use substantives ;)
why not use "transition" rather than "outcome", so it looks like
```
transition:
   state: my-state-name
```
or
```
transition:
   end: {}
```
At the end of the day, ending is transitioning to the end ;)

cdavernas:
@fjtirado `transition` or `next` sounds imo like its related to transitioning, and therefore does not encompass both concerns.
And you are right, `outcome` is definitly not the right term after all.

What about `flow` or `flowControl`?

```yaml
flow:
  to: my-state
```

or 

```yaml
flow:
  end: {}
```

I think we still need to address the proper definition of what is `state`, an `action` or even a `function`. This is related to #677 too as well as to your legit concern of denormalizing `custom` function types etc. 

The reason I speak of this is because I believe that what states are is a foggy mix of `stateData`,`flowControlActions`(conditions, switch statements, etc.) and even, in some cases, `actions` (inject, sleep, etc.).

I think this simple problem clearly shows some crack in the consistency of core "components" of the spec which we will need to address at some point anyways. Why not now?

fjtirado:
Yes, I think we should separate states from flow control. 
Right now, if you want to reuse a state, you have to defint it twice if the next state is different. 
Lets illustrate with an example, we have states A, B and C
And the flow is start->A->B->C-B->end, since we are including transition information in the state, right now we need to define B twice with two different names, when it will be better to define the states in one place and then the "connections" between them

cdavernas:
Yes, exactly! Something like:

- States: Represent the state of a workflow at a specific point in time, and is made out of actions with shared data. Should be named using present participles (ex: retrieving-info, registering-user, ...).
- Flow controls: Used to transition from one state to another or to end workflow execution.
- Actions/activities:  Individual units of work that perform specific tasks or operations within the workflow.

Right?

Where should we even start 😭

ricardozanini:
That should be done for a 0.10 version since it  is a quite huge change that will break many things :D

Issue #685, opened 2022-09-19
Enforce action name


JBBianchi:
**What would you like to be added**:
The name of actions to be `required`.

**Why is this needed**:
The name is a human, comprehensible way of working. When reviewing the activities of a workflow, knowing that "Action[2]" was a success doesn't really speak to the user. It requires to have the definition under the eyes, count the actions to find the one we are speaking about. Having it named will improve readability and maintainability of workflows and their activities.

ricardozanini:
+1

tsurdilo:
This is done i believe as part of the error handling updates pr.

JBBianchi:
This also applies to other item types like `branch`.

A name should be unique within the scope of its parent but not the whole workflow.

ricardozanini:
As agreed in the meeting call, within the same state, we should guarantee action uniqueness

tsurdilo:
Yes branch names are also going to be enforced as part of error handling updates.

jaliyaudagedara:
Seems this suggested hasn't moved to the specification yet. Should we update the spec?

Issue #671, opened 2022-09-09
Enforce naming conventions


cdavernas:
**What would you like to be added**:

Enforce naming conventions of most (if not all) components.

I propose that we enforce using snake and/or kebab cased names.

**Why is this needed**:

In the case of `functionDefs` of type `expression`, for example, not enforcing a naming convention makes expression parsing a huge mess.

Consider the following example:

```yaml
functions:
  - name: Say Hello
    type: expression
    operation: 'Hello!'
```

Now to call the function in an expression as defined by the spec, you'd need to write something like:

`. | . * { "greetings": "fn:Say Hello" }`

... which is impossible to parse both properly and/or safely. We could of course mandate to put the function's name between parenthesis (ex: `$fn:(Say Hello)`), for example, but I think the easier and cleanier approach is to go with a naming convention

tsurdilo:
Agree, will do pr for it

fjtirado:
> Let's use Kubernetes convention for naming: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/
> 
> They are based on RFC1123, so it's safe for DNS too.

Hmmm, I think we should support _ for naming, without the need of escaping it

fjtirado:
Being practical, I would use this https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#path-segment-names
Since the only name in the flow that might appear in k8s is the id, just avoiding ., .., / and % looks safe
That for forbidden special character. 
For the name convention I think we can support both: snake and kebab (depending on the taste of the consumer ;))

PR #768, opened 2023-05-26
Updates schema to enforce names as required RFC1123 values

cdavernas:

- [x] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

**Discussion or Issue link**:
#671 #685

- Enforce names of states, events, functions, actions, branches and conditions to be required and to follow the [RFC1123](https://datatracker.ietf.org/doc/html/rfc1123) DNS label format.

### DIFF
--- a/schema/auth.json
+++ b/schema/auth.json
@@ -32,7 +32,8 @@
         "name": {
           "type": "string",
           "description": "Unique auth definition name",
-          "minLength": 1
+          "minLength": 1,
+          "pattern": "^[a-z0-9]([-a-z0-9]*[a-z0-9])$"
         },
         "scheme": {
           "type": "string",
--- a/schema/auth.json
+++ b/schema/auth.json
@@ -33,7 +33,7 @@
           "type": "string",
           "description": "Unique auth definition name",
           "minLength": 1,
-          "pattern": "^[a-z0-9]([-a-z0-9]*[a-z0-9])$"
+          "pattern": "^[a-z0-9](-?[a-z0-9])*$"
         },
         "scheme": {
           "type": "string",
--- a/specification.md
+++ b/specification.md
@@ -97,6 +97,7 @@
   * [Workflow Secrets](#workflow-secrets)
   * [Workflow Metadata](#workflow-metadata)
   * [Workflow Context](#workflow-context)
+  * [Naming Convention](#naming-convention)
 - [Extensions](#extensions)
 - [Use Cases](#use-cases)
 - [Examples](#examples)
@@ -6196,6 +6197,12 @@ Implementations may use this keyword to give access to any relevant information 
 In this use case, a third-party service may require information from the caller for traceability purposes.
 
 The specification doesn't define any specific variable within the `WORKFLOW` bucket, but it's considered a reserved keyword.
+
+### Naming Convention
+
+Identifiable components of a workflow definition, such as states, actions, branches, events and functions define a required non-null `name` property which follows the DNS label names as defined by [RFC 1123](https://datatracker.ietf.org/doc/html/rfc1123).
+
+In other words, `names` must be lowercased, start and end with an alphanumeric character, and contain only alphanumeric characters or '-' (dash) character.
 
 ## Extensions
 
--- a/specification.md
+++ b/specification.md
@@ -6200,9 +6200,9 @@ The specification doesn't define any specific variable within the `WORKFLOW` buc
 
 ### Naming Convention
 
-Identifiable components of a workflow definition, such as states, actions, branches, events and functions define a required non-null `name` property which follows the DNS label names as defined by [RFC 1123](https://datatracker.ietf.org/doc/html/rfc1123).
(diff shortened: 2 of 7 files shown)


specification.md
```diff
@@ -6193,6 +6198,12 @@ In this use case, a third-party service may require information from the caller
 
 The specification doesn't define any specific variable within the `WORKFLOW` bucket, but it's considered a reserved keyword.
 
+### Naming Convention
+
+Identifiable components of a workflow definition, such as states, actions, branches, events and functions define a required non-null `name` property which follows the DNS label names as defined by [RFC 1123](https://datatracker.ietf.org/doc/html/rfc1123).
```
ricardozanini, line 6203:
```suggestion
Identifiable components of a workflow definition, such as states, actions, branches, events and functions define a required non-null `name` property which follows the DNS label names as defined by [RFC 1123](https://datatracker.ietf.org/doc/html/rfc1123#page-13).
```

ricardozanini (approved):

specification.md
```diff
@@ -6193,6 +6198,12 @@ In this use case, a third-party service may require information from the caller
...
+
+In other words, `names` must be lowercased, start and end with an alphanumeric character, and contain only alphanumeric characters or '-' (dash) character.
```
gibson042, line 6205: RFC 1123 does not include any lowercasing requirement; did you intend to make this a further restriction?

I think some text is also warranted regarding [RFC 5890 R-LDH labels](https://www.rfc-editor.org/rfc/rfc5890#section-2.3.1) (whose third and fourth characters are both "-") such as `xn--bcher-kva` (which is further an XN-label and valid A-label representing "bücher") and `we--like--it` (which is neither but is still subject to the "_MUST NOT be processed as ordinary LDH labels by IDNA-conforming programs_" restriction).

I don't think it matters whether they are allowed or rejected, but some software _does_ treat them specially so it's worth being explicit.

cdavernas: @gibson042 I thought it did, or at least does it's k8s flavor: 

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#:~:text=RFC%201123%20Label%20Names&text=in%20RFC%201123.-,This%20means%20the%20name%20must%3A,start%20with%20an%20alphanumeric%20character

Anyways, I do think adding the lowercase restriction, as i believe it could help ensure invariance, especially on exotic systems. 

It's not however a strong preference, and I'm obviously open to any and all opinion for and against ;)

As for the double dashes, I would venture to say it's safe to forbid its use altogether, like with the k8s regex I'm using in the PR. WDYT?

gibson042: :+1:

schema/auth.json
```diff
@@ -32,7 +32,8 @@
         "name": {
           "type": "string",
           "description": "Unique auth definition name",
-          "minLength": 1
+          "minLength": 1,
+          "pattern": "^[a-z0-9]([-a-z0-9]*[a-z0-9])$"
```
gibson042, line 36: Per https://github.com/serverlessworkflow/specification/pull/768#discussion_r1208070568 , this should either prohibit "--" at index 2
```suggestion
          "pattern": "^[a-z0-9](([-a-z0-9]){0,2}[a-z0-9])?$|^[a-z0-9][-a-z0-9]([a-z0-9]{2}|-[a-z0-9]|[a-z0-9]-)[-a-z0-9]*[a-z0-9]$"
```
or prohibit any consecutive dashes
```suggestion
          "pattern": "^[a-z0-9](-?[a-z0-9])*$"
```
(here and in all the other analogous regular expressions)

specification.md
gibson042, line 6205:
```suggestion
Identifiable components of a workflow definition, such as states, actions, branches, events and functions define a required non-null `name` property which is based on DNS label names as defined by [RFC 1123](https://datatracker.ietf.org/doc/html/rfc1123) with further restrictions.

Specifically, `name`s must be lowercase, start and end with an alphanumeric character, consist entirely of alphanumeric characters and/or dashes '-', and (for compatibility with [RFC 5890](https://datatracker.ietf.org/doc/html/rfc5890) not contain '--' as the third and fourth characters.
```
or
```suggestion
Identifiable components of a workflow definition, such as states, actions, branches, events and functions define a required non-null `name` property which is based on DNS label names as defined by [RFC 1123](https://datatracker.ietf.org/doc/html/rfc1123) with further restrictions.

Specifically, `name`s must be lowercase, start and end with an alphanumeric character, and consist entirely of alphanumeric characters with optional isolated medial dashes '-' (i.e., dashes must not be adjacent to each other).
```

gibson042 (approved):
@cdavernas Thanks for the updates!

cdavernas:
@gibson042 No, thanks to you for both the comments and the suggested edits ❤️

ricardozanini (approved):

PR #826, opened 2024-02-28
[Fix #677] Atomic produce and consume actions

### LINKED ISSUE #677: Atomic produce and consume actions

fjtirado:

- [X] Specification
- [X] Schema
- [X] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

**Discussion or Issue link**:

Fixes https://github.com/serverlessworkflow/specification/issues/677

Separated produceEventRef from consumeEventRef, removing eventRef, and promoting both to first class actions.
Remove kind property from event definition. Therefore the same event definition can be both used for consumption and publishing. if needed. 
I have couple of doubts about naming notation that I reflected as comments. 
**Additional information:**

### DIFF
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -413,13 +413,13 @@
           "description": "References a function to be invoked",
           "$ref": "#/definitions/functionref"
         },
-        "produceEventRef": {
-          "description": "References a `produce` reusable event definition",
-          "$ref": "#/definitions/produceeventref"
+        "publish": {
+          "description": "Publish an event",
+          "$ref": "#/definitions/publish"
         },
-        "consumeEventRef": {
-          "description": "References a `consume` reusable event definition",
-          "$ref": "#/definitions/consumeeventref"
+        "subscribe": {
+          "description": "Subscribe to an event channel",
+          "$ref": "#/definitions/subscribe"
         },
         "subFlowRef": {
           "description": "References a sub-workflow to invoke",
@@ -472,13 +472,13 @@
         {
           "required": [
             "name",
-            "produceEventRef"
+            "publish"
           ]
         },
         {
           "required": [
             "name",
-            "consumeEventRef"
+            "subscribe"
           ]
         },
         {
@@ -531,13 +531,13 @@
         }
       ]
     },
-    "produceeventref": {
+    "publish": {
       "type": "object",
       "description": "Publish an event",
       "properties": {
-        "name": {
+        "event": {
           "type": "string",
-          "description": "Reference to the unique name of a 'produced' event definition",
+          "description": "Reference to the unique name of a 'published' event definition",
           "pattern": "^[a-z0-9](-?[a-z0-9])*$"
         },
         "data": {
@@ -545,7 +545,7 @@
             "string",
             "object"
           ],
-          "description": "If string type, an expression which selects parts of the states data output to become the data (payload) of the event referenced by 'produceEventRef'. If object type, a custom object to become the data (payload) of the event referenced by 'produceEventRef'."
+          "description": "If string type, an expression which selects parts of the states data output to become the data (payload) of the event referenced by 'publish'. If object type, a custom object to become the data (payload) of the event referenced by 'publish'."
         },
         "contextAttributes": {
           "type": "object",
@@ -557,26 +557,26 @@
       },
       "additionalProperties": false,
       "required": [
-        "name", "data"
+        "event", "data"
       ]
     },
-    "consumeeventref": {
+    "subscribe": {
       "type": "object",
-      "description": "Waits for an event",
+      "description": "Subscribe to an event channel",
       "properties": {
-        "name": {
+        "event": {
           "type": "string",
-          "description": "Reference to the unique name of a 'consumed' event definition",
+          "description": "Reference to the unique name of a 'subscribed' event definition",
           "pattern": "^[a-z0-9](-?[a-z0-9])*$"
         },
-        "consumeEventTimeout": {
+        "timeout": {
           "type": "string",
           "description": "Maximum amount of time (ISO 8601 format) to wait for the result event. If not defined it should default to the actionExecutionTimeout"
         }
        },
       "additionalProperties": false,
       "required": [
-        "name"
+        "event"
       ]
     },
     "subflowref": {
--- a/schema/workflow.json
+++ b/schema/workflow.json
@@ -413,9 +413,13 @@
           "description": "References a function to be invoked",
           "$ref": "#/definitions/functionref"
         },
-        "eventRef": {
-          "description": "References a `produce` and `consume` reusable event definitions",
-          "$ref": "#/definitions/eventref"
+        "produceEventRef": {
+          "description": "References a `produce` reusable event definition",
+          "$ref": "#/definitions/produceeventref"
+        },
+        "consumeEventRef": {
+          "description": "References a `consume` reusable event definition",
+          "$ref": "#/definitions/consumeeventref"
         },
         "subFlowRef": {
           "description": "References a sub-workflow to invoke",
@@ -468,7 +472,13 @@
         {
           "required": [
             "name",
-            "eventRef"
+            "produceEventRef"
+          ]
+        },
+        {
+          "required": [
+            "name",
+            "consumeEventRef"
           ]
         },
         {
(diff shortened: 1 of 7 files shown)


specification.md
```diff
@@ -4175,27 +4155,61 @@ eventRef:
...
-Functions that are invoked via events (sync or async) do not propagate their errors to the associated action definition and the
-workflow state, meaning that any errors that happen during their execution cannot be handled in the workflow states
-onErrors definition. Note that errors raised during functions that are invoked sync or async in this case
-should not fail workflow execution.
+Wait for an event to arrive. 
+
+| Parameter | Description | Type | Required |
+| --- | --- | --- | --- |
+| [name](#Event-Definition) | Reference to the unique name of an event definition. Must follow the [Serverless Workflow Naming Convention](#naming-convention) | string | yes |
+| consumeEventTimeout | Maximum amount of time (ISO 8601 format literal or expression) to wait for the consume event. If not defined it be set to the [actionExecutionTimeout](#Workflow-Timeout-Definition) | string | no |
```
fjtirado, line 4173: Should we use eventTimeout instead?

specification.md
```diff
@@ -4175,27 +4155,61 @@ eventRef:
...
-its retry definition, if defined, should be ignored.
-Functions that are invoked via events (sync or async) do not propagate their errors to the associated action definition and the
-workflow state, meaning that any errors that happen during their execution cannot be handled in the workflow states
-onErrors definition. Note that errors raised during functions that are invoked sync or async in this case
-should not fail workflow execution.
+Wait for an event to arrive. 
+
+| Parameter | Description | Type | Required |
+| --- | --- | --- | --- |
+| [name](#Event-Definition) | Reference to the unique name of an event definition. Must follow the [Serverless Workflow Naming Convention](#naming-convention) | string | yes |
```
fjtirado, line 4172: Should we use eventRef rather than name?

schema/workflow.json
```diff
@@ -556,7 +557,26 @@
...
+    "consumeeventref": {
+      "type": "object",
+      "description": "Waits for an event",
+      "properties": {
+        "name": {
+          "type": "string",
+          "description": "Reference to the unique name of a 'consumed' event definition",
+          "pattern": "^[a-z0-9](-?[a-z0-9])*$"
+        },
+        "consumeEventTimeout": {
```
cdavernas, line 572: Why not take the opportunity to rename that property to `timeout`?

schema/workflow.json
```diff
@@ -521,24 +531,15 @@
         }
       ]
     },
-    "eventref": {
+    "produceeventref": {
```
cdavernas, line 534: I would take this opportunity to rename action top level properties to something like:

```yaml
actions:
  - name: my-action
    call: {} #instead of functionRef, maybe in another PR?
    pub: {} #or publish, or produce, instead of produceEventRef
    sub: {} #or subscribe, or consume, instead of consumeEventRef
```

cdavernas (changes requested):
Many thanks for the work! ❤️

Just a couple of pita comments to discuss😄

schema/workflow.json
fjtirado: I agree, you should have answered my https://github.com/serverlessworkflow/specification/pull/826#discussion_r1505827411

schema/workflow.json
fjtirado: I think call should be done with another PR
I think we can use publish and subscribe, combined what which I commented in https://github.com/serverlessworkflow/specification/pull/826#discussion_r1505827718, so rather than name, we have eventRef property (so a publish that refers to event xxxx, a subscribe that refers to event xxxX)

schema/workflow.json
cdavernas: Sorry, I overlooked it!

specification.md
cdavernas: I would say `timeout` instead.

specification.md
cdavernas: In the case we use the approach suggested by https://github.com/serverlessworkflow/specification/pull/826#discussion_r1517565305, my personal preference goes to `name`. Otherwise, I would go for `to` for for subscribe, and `event` in case of produce.

So:

```yaml
actions:

  - name: my-action-1
    subscribe:
      to: my-consumed-event
      
  - name: my-action-2
    publish:
      event: my-produced-event
```

schema/workflow.json
cdavernas: I think that we should do it in this PR, as we are altering the structure of an action anyways, let's do it right this time then.

`publish` and ` subscribe` sound perfect to me.

I replied to your comment, in its tread 😉

specification.md
fjtirado: hmmm, you subscribe to an event and publish an event, so the "to" looks weird to me. The preposition is not really relevant, which is relevant is that you are refererring to an event that is defined in the header of the workflow.

cdavernas: > hmmm, you subscribe to an event and publish an event, so the "to" looks weird to me

Yeah, you are right. If only we could use a similar terminology than `to` for `publish`, it would trully be awesome.

> you are refererring to an event that is defined in the header of the workflow

Yup, and that's another problem, thus my concern with the proposed terminology.
Ideally, we should not force the user to declare an event at top level. For the sake of readability, one should be able to inline the definition of a non-reused event. In most cases, you are not gonna "reuse" events anyways.

fjtirado: If we adop a fluent syntax, _publish_ and _subsribe_ are the verbs, while _event_ (or _name_, but _event_ has more meaning) is the object, _to_ (and _a_) is just a nexus with no meaning.
Even if we allow defining the event inline, event still makes sense, because it can be the ref to the event definition or the event definition itself.

schema/workflow.json
fjtirado: Regarding call, I really think is better to have smaller PRs, just to keep discussion smaller ;)

ricardozanini: Let's change it here, I think it's reasonable 👍

specification.md
ricardozanini: @cdavernas @fjtirado so:

``` yaml
actions:
  - name: my-action-1
    subscribe: my-consumed-event
     
  - name: my-action-2
    publish: my-produced-event
```

ricardozanini: Seems clear and straightforward to me.

fjtirado: publish and subscribe are object
you need to write
```
publish: 
  event: my-produced-event
  data: .eventContent
subscribe:
  event: my-consumed-evemt
  timeout: 5PTS
```

cdavernas: Then they should be string (ref) or object, imho.

fjtirado: For subscribe maybe. For publish, I put data as required, because an cloud event with no data is pretty strange scenario (can be done by returning empty json, but I think it should be explicit)

fjtirado: I have changed to subscribe and publish, and name to event.

schema/workflow.json
fjtirado: I would say this should be a new issue explaining why we want to change funcionRef to call or functionCall. If we do so, we need to purge the word reference from the spec.

cdavernas (changes requested):
Great work, thank you!

Could I please ask you to update the `roadmap.md` to include the changes brought by this PR?

schema/workflow.json
```diff
@@ -556,7 +557,26 @@
...
       "required": [
-        "produceEventRef"
+        "event", "data"
+      ]
+    },
+    "subscribe": {
+      "type": "object",
+      "description": "Subscribe to an event channel",
+      "properties": {
+        "event": {
```
cdavernas, line 567: Could you move the `event` definition out of both `subscribe` and `publish` and instead reference it?
It removes a duplicate and is therefore easier to maintain.

cdavernas (changes requested):

fjtirado:
> Great work, thank you!
> 
> Could I please ask you to update the `roadmap.md` to include the changes brought by this PR?

This is already there in the roadmap for 0.9

schema/workflow.json
fjtirado: I do not understand, right now event is just an string with the regular expression we are using for names (that is duplicated along the whole spec) , whats the benefit of replacing?
```
"event": {
          "type": "string",
          "description": "Reference to the unique name of a 'published' event definition",
          "pattern": "^[a-z0-9](-?[a-z0-9])*$"
        }
```
by
```
"event": {
    "type": "#/definitions/eventRef" 
}
```
I would see the benefit, but I think is matter of a clean up PR, to put all names under a new definition "nameWithReg"

cdavernas:
> This is already there in the roadmap for 0.9

@fjtirado You are right! Sorry about that!

schema/workflow.json
cdavernas: Just created https://github.com/serverlessworkflow/specification/issues/828 to address it.

cdavernas (approved):

ricardozanini (approved):

cdavernas:
Merging this PR as per contribution rules: it has been out for more than 3 weeks already.

PR #832, opened 2024-03-18
[Fix #712] Describe merge behaviour for non object

### LINKED ISSUE #712: Define specific rules when merging action output by default

Define specific rules when merging action output by default

**Why is this needed**:

So far, the spec mandates that the output of an action without `toState` data filter should be merged back 'as is' into the state data. 
Problem is, if the output of said action is not an object, merging it into state data will effectilvely fault the whole workflow: all evaluations will fail.

Image for instance an action that returns, say, an array or, why not, an integer. Your state was the following beforehand:

```json
{
  "someName": "someValue"
}
```

After the action, the state data would be:

```json
[
 1,
 2,
 3
]
```

Or why not something like:

```json
125
```

Which would result in all further actions to fail. Consider, for example, a next action with a `fromState` data filter => BOOM.

What I propose is that the action output be merged back "as-is" ONLY if/when it's an object. If it's an array or a value type (string, bool, int, ...), it should instead be merged back into a 'actionOuput' property, for example. The name of that property could be based on the name of the action, such as `myActionOutput`.

WDYT?

fjtirado:

- [x] Specification
- [ ] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

**Discussion or Issue link**:

Fixes https://github.com/serverlessworkflow/specification/issues/712

Update specification to make runtimes predicateble when an action produced a non json object and there is not toStateData filter

**Additional information:**

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -439,7 +439,7 @@ The second way would be to directly filter only the "veggie like" vegetables wit
 | fromStateData | Workflow expression that filters state data that can be used by the action | string | no |
 | useResults | If set to `false`, action data results are not added/merged to state data. In this case 'results' and 'toStateData' should be ignored. Default is `true`.  | boolean | no |
 | results | Workflow expression that filters the actions data results | string | no |
-| toStateData | Workflow expression that selects a state data element to which the action results should be added/merged into. If not specified denotes the top-level state data element. In case it is not specified and the result of the action is not an object, that result should be merged as value of an automatically generated key. That key name will be the result of concatenating the action name with `_output` suffix. | string | no |
+| toStateData | Workflow expression that selects a state data element to which the action results should be added/merged. If not specified denotes the top-level state data element. In case it is not specified and the result of the action is not an object, that result should be merged as the value of an automatically generated key. That key name will be the result of concatenating the action name with `-output` suffix. | string | no |
 
 <details><summary><strong>Click to view example definition</strong></summary>
 <p>
@@ -565,25 +565,24 @@ into. With this, after our action executes the state data would be:
 }
 ```
 
-To  illustrate merge of not json both object,  let`s assume that, in previous example, the action definition is at follows
+To illustrate the merge of non-JSON for both objects,  let's assume that, in the previous example, the action definition is the follows
 
 ```json
 "actions":[
     {
-       "name": "fetch_only_pasta",
+       "name": "fetch-only-pasta",
        "functionRef": "breadAndPastaTypesFunction",
        "actionDataFilter": {
           "results": "${ .pasta[1] ]",
        }
     }
  ]
-}
 ```
-Since there is not `toStateData` and the result is not a json object but an string, the model would be
+Since there is no `toStateData` attribute and the result is not a JSON object but a string, the model would be:
 
 ```json
 {
-  "fetch_only_pasta_output": "spaghetti"
+  "fetch-only-pasta-output": "spaghetti"
 }
 ```
 In the case action results should not be added/merged to state data, we can set the `useResults` property to `false`.
--- a/specification.md
+++ b/specification.md
@@ -565,6 +565,27 @@ into. With this, after our action executes the state data would be:
 }
 ```
 
+To  illustrate merge of not json both object,  let`s assume that, in previous example, the action definition is at follows
+
+```json
+"actions":[
+    {
+       "name": "fetch_only_pasta",
+       "functionRef": "breadAndPastaTypesFunction",
+       "actionDataFilter": {
+          "results": "${ .pasta[1] ]",
+       }
+    }
+ ]
+}
+```
+Since there is not `toStateData` and the result is not a json object but an string, the state would be
+
+```json
+{
+  "response": "spaghetti"
+}
+```
 In the case action results should not be added/merged to state data, we can set the `useResults` property to `false`.
 In this case, the `results` and `toStateData` properties should be ignored, and nothing is added/merged to state data.
 If `useResults` is not specified (or it's value set to `true`), action results, if available, should be added/merged to state data.
--- a/specification.md
+++ b/specification.md
@@ -439,7 +439,7 @@ The second way would be to directly filter only the "veggie like" vegetables wit
 | fromStateData | Workflow expression that filters state data that can be used by the action | string | no |
 | useResults | If set to `false`, action data results are not added/merged to state data. In this case 'results' and 'toStateData' should be ignored. Default is `true`.  | boolean | no |
 | results | Workflow expression that filters the actions data results | string | no |
-| toStateData | Workflow expression that selects a state data element to which the action results should be added/merged into. If not specified denotes the top-level state data element. In case it is not specified and the result of the action is not an object, that result should be merged as value of `response` key. If that `response` key already exist in the model, its value will be overwritten. | string | no |
+| toStateData | Workflow expression that selects a state data element to which the action results should be added/merged into. If not specified denotes the top-level state data element. In case it is not specified and the result of the action is not an object, that result should be merged as value of an automatically generated key. That key name will be the result of concatenating the action name with `_output` suffix. | string | no |
 
 <details><summary><strong>Click to view example definition</strong></summary>
 <p>
@@ -579,11 +579,11 @@ To  illustrate merge of not json both object,  let`s assume that, in previous ex
  ]
 }
 ```
-Since there is not `toStateData` and the result is not a json object but an string, the state would be
+Since there is not `toStateData` and the result is not a json object but an string, the model would be
 
 ```json
 {
-  "response": "spaghetti"
+  "fetch_only_pasta_output": "spaghetti"
 }
 ```
 In the case action results should not be added/merged to state data, we can set the `useResults` property to `false`.
--- a/specification.md
+++ b/specification.md
@@ -439,7 +439,7 @@ The second way would be to directly filter only the "veggie like" vegetables wit
 | fromStateData | Workflow expression that filters state data that can be used by the action | string | no |
 | useResults | If set to `false`, action data results are not added/merged to state data. In this case 'results' and 'toStateData' should be ignored. Default is `true`.  | boolean | no |
 | results | Workflow expression that filters the actions data results | string | no |
-| toStateData | Workflow expression that selects a state data element to which the action results should be added/merged into. If not specified denotes the top-level state data element | string | no |
+| toStateData | Workflow expression that selects a state data element to which the action results should be added/merged into. If not specified denotes the top-level state data element. In case it is not specified and the result of the action is not an object, that result should be merged as value of `response` key. If that `response` key already exist in the model, its value will be overwritten. | string | no |
 
 <details><summary><strong>Click to view example definition</strong></summary>
 <p>


specification.md
```diff
@@ -439,7 +439,7 @@ The second way would be to directly filter only the "veggie like" vegetables wit
 | fromStateData | Workflow expression that filters state data that can be used by the action | string | no |
 | useResults | If set to `false`, action data results are not added/merged to state data. In this case 'results' and 'toStateData' should be ignored. Default is `true`.  | boolean | no |
 | results | Workflow expression that filters the actions data results | string | no |
-| toStateData | Workflow expression that selects a state data element to which the action results should be added/merged into. If not specified denotes the top-level state data element | string | no |
+| toStateData | Workflow expression that selects a state data element to which the action results should be added/merged into. If not specified denotes the top-level state data element. In case it is not specified and the result of the action is not an object, that result should be merged as value of `response` key. If that `response` key already exist in the model, its value will be overwritten. | string | no |
```
ricardozanini, line 442: I think we can add an example to this description to make it clearer.

cdavernas, line 442: @fjtirado I had totally forgotten about #712... 
Don't you think it's better to settle on something like: if you dont set `toStateDate`, the output of an action is simply ignored? 
I'm not really confortable with this arbitrary decision to enforce a default behavior for this when not specified.
I personally would just remove: `If not specified denotes the top-level state data element` and instead indicate that if not set, nothing happens, which is IMHO more intuitive.
@JBBianchi @ricardozanini @fjtirado What do you guys think?

cdavernas (changes requested):

specification.md
fjtirado: hmmmm, I think the non object should still be merged (the model has to be always an object so setting it as the flow model is not an option). The discussion, I think, is to decide the default name of the property to which the non object value is assigned- 
I propose `response` because thats the one that is being currently used in our implementation. Might be not best name, as for example `key` for the expression used to generate the workflow instance id, but it is not the worse either. As an enhacement, we can add a property at the header of the workflow definition to specify the name (if response is not the one that is wanted) to enforce implementors to support that.

cdavernas: That's not my point. Object or non-object is irrelevant to the matter I'm speaking about. My concern is the fact that, if the `toStateData` is not set, the output should not be assigned to anything, it should instead be ignored, which I believe is the (intuitive) way to go.

fjtirado: Maybe I am influenced by the current behaviour (merging always) but I found it pretty intuitive. Most actions produce a result and that result should be available to the model. If the result of the action is irrelevant (for an action that actually produce a result, because tipically actios which result is not relevan will not produce any result, so nothing is merged), In many case it is ok to just the result (object or not object) direcly into the model, without the need to perform any key traslation. 
If we force people to write toStateData to perform a merge, the resulting definition will become too verbose, in my opinion. Im actually surprised you are proposing this ;)

fjtirado: Thats said, since we already has customers, maybe we should request feedback about this to them. But I think this will imply them to write toStateData for every rest request they perform, when they are certainly fine with the current merge approach. 
Maybe we can just leave merge as default and globally change that by an optional key in the header, so we have an optional `mergePolicy` which default value is merge, but that can be none.

cdavernas: I understand your point of view and it does make sense IMO.
However, my problem is with the "tricks" that have been put into place to avoid having to automatically merge, meaning the (very ugly) `useResult` property. The latter is not mutually exclusive to `toStateData`, and is therefore leaving the door opened to wierd definitions where both properties are set, even though they address the same concern.
My idea is to remove the `useResult` property, and give the feature to an empty `toStateData`, but I agree that, in that case, the workflow might become a bit more verbose.
I believe that the issue is related to the concept of `stateData`: what is it and how should it be used? If it's a concept that represents contextual data at a point in time (aka state) of a workflow's execution, then I believe it should never be overidden by an action's output. Furthermore, there's a conceptual issue where the output of an action should be implicitly used as the input of next action, save if the `fromStateData` has been used. This always felt wierd to me, and to our clients, too, even though it can be practical in exotic use cases. In my opinion, `stateData` should not be changed by actions if not explicitly requested, and actions should always receive the output, if any, of the previous action, thus making `fromStateData` useless (which it is, providing you could access state data with something similar to a `$STATE` expression parameter).

cdavernas: > Thats said, since we already has customers, maybe we should request feedback about this to them

Yes, you are totally right, I believe this very subtle issue is actually bigger than us, and should be submitted to our community and opposed to their factual use cases.

fjtirado: As as summary of my previous comment, as alternative proposal
1) Remove `actionDataFilter`. Add actionResultFilter which basically modified at will the result of the action execution
2) This result is not merged into the workflow model, but make accessible to the next action (or to the state if it is the last acion) as `$WORKFLOW.lastActionResult`. This is also accesible to `stateDataFilter.output`. Note that lastActionResult might be json object or a non object, it does not really matter
3) Add a new mandatory/optional (point of discussion) property to operation state. `resultPolicy`, posible values are `merge` (lastActionResult is merged into the model)  or `ignore` (do nothing)
4) When merging a non object into the workflow model (for any state), that non object will be associated with `response` key. That `response` key might be change by specifyng another value through workflow level property `nonObjectm̀ergeKey`

fjtirado: I think that `resultPolicy`should be mutually exclusive with `stateDataFilter.ouptut`. I mean, either the user specified  how to merge `lastActionResult` into the model by setting and output filter of the state or he specifies that `lastActionResult` is to be merged as it is or that nothing should be done
Another way to look at it is that resultPolicy has three posible values: filter, merge o none. 
If you specify the filter you can assume its value is filter. Then if you do not say anything, it can be assumed (for continuity with previous version) that you want to merge the result. And only if you do not want to do anything, you specify `ignore` (thats the boolean we currently have, that in any case will be moved to the right place, the state) I think is fine to just an optional boolean "doNotMerge" to operation state. I think 99% of the time you want your operation state to merge the result (that was not true for the action, thats why moving the boolean to state level does the trick)

fjtirado: @cdavernas wdyt?

cdavernas: > Remove actionDataFilter. Add,as replacement actionResultFilter of type string

Good suggestion. However, this removes the user's ability to add potentially critical data to the state's after executing an action: only the last action's output is made available. There are several use cases where that is problematic and too restrictive.

cdavernas: For a short term fix, I think we should indicate that when the `toStateData` property of an `actionDataFilter` is not set, it does not output anything to the state's data.

For the `stateDataFilter`, however, we should proceed as initially planned, meaning that if not set, it should be merged back at top level. If the data is not an object, it should be merged back into a field such as `response`, as you suggested, though I'm not a fan of that specific term. Why not simply calling it `output`, which is more ubiquitous IMHO than response, which implies the concept of a request?

cdavernas: For a longer term solution, I think we should first rename `dataFilter`. The term is not ubiquitous, and might even be confusing to some users.

I would personally rename it to `output`, and modifiy it so that it looks like:

```yaml
actions:

- name: action-with-no-output
  output: false

- name: action-with-filtered-output-1
  output: '${ .myFilteredProperty }'

- name: action-with-filtered-output-2
  output:
    filter: '${ .myFilteredProperty }'

- name: action-that-sets-state-data
  output:
    to: '${  .myStateDataProperty }'
```

The same could be done for `input`. And we could even include the (optional) schema in those new objects.

We could also allow action/state data to be of any kind (object, array, primitive), as I don't see the need to enforce it to be an object, which in addition force us to propose ways around it in specific use cases.

fjtirado: @cdavernas It seems we do not agree on default behaviour over the response of the action. I think that, by default, the result of the last action of an operation state should be merged into the workflow model. This is the current default and is a good one. As a short term fix, all we need to do is to indicate what to do when the result is not an object (my proposal is to use a configurable key, which defaults to "response")

fjtirado: One possibility is to merge automatically the result of the last action and only merge the result of intermediate actions if there is a toStateData or similar. Most typical use case is an operation with just one action that invokes a rest (or other function returning a value, json object or not), for that use case, that will work smoothly. For action chains, I agree that only allowing merging the result of the last action is too restrictive. Therefore, for intermediate actions, your sugesstion of only merging when the user explicitly states it makes perfect sense.

fjtirado: But we are discussing two different things:
1) To merge o nor to merge
2) How to merge a non object json. 
I propose, within the scope of this PR, just discuss number 2)

cdavernas: Agreed!

For 2, I instead propose using `output`, that is IMHO more ubiquitous.

fjtirado: Ok, since action name is now mandatory, we can enforce <actionName>Output, wdyt?

fjtirado: Done

specification.md
```diff
@@ -439,7 +439,7 @@ The second way would be to directly filter only the "veggie like" vegetables wit
 | fromStateData | Workflow expression that filters state data that can be used by the action | string | no |
 | useResults | If set to `false`, action data results are not added/merged to state data. In this case 'results' and 'toStateData' should be ignored. Default is `true`.  | boolean | no |
 | results | Workflow expression that filters the actions data results | string | no |
-| toStateData | Workflow expression that selects a state data element to which the action results should be added/merged into. If not specified denotes the top-level state data element | string | no |
+| toStateData | Workflow expression that selects a state data element to which the action results should be added/merged into. If not specified denotes the top-level state data element. In case it is not specified and the result of the action is not an object, that result should be merged as value of an automatically generated key. That key name will be the result of concatenating the action name with `_output` suffix. | string | no |
```
ricardozanini, line 442:
```suggestion
| toStateData | Workflow expression that selects a state data element to which the action results should be added/merged. If not specified denotes the top-level state data element. In case it is not specified and the result of the action is not an object, that result should be merged as the value of an automatically generated key. That key name will be the result of concatenating the action name with `-output` suffix. | string | no |
```

We should use hyphen to follow the naming conventions.

specification.md
```diff
@@ -565,6 +565,27 @@ into. With this, after our action executes the state data would be:
...
+"actions":[
+    {
+       "name": "fetch_only_pasta",
+       "functionRef": "breadAndPastaTypesFunction",
+       "actionDataFilter": {
+          "results": "${ .pasta[1] ]",
+       }
+    }
+ ]
+}
```
ricardozanini, line 580:
```suggestion
```

specification.md
```diff
@@ -565,6 +565,27 @@ into. With this, after our action executes the state data would be:
 }
 ```
 
+To  illustrate merge of not json both object,  let`s assume that, in previous example, the action definition is at follows
```
ricardozanini, line 568:
```suggestion
To illustrate the merge of non-JSON for both objects,  let's assume that, in the previous example, the action definition is the follows
```

I think we need to rephrase this. Not sure if I understood this paragraph.

specification.md
```diff
@@ -565,6 +565,27 @@ into. With this, after our action executes the state data would be:
...
+       "name": "fetch_only_pasta",
+       "functionRef": "breadAndPastaTypesFunction",
+       "actionDataFilter": {
+          "results": "${ .pasta[1] ]",
+       }
+    }
+ ]
+}
+```
+Since there is not `toStateData` and the result is not a json object but an string, the model would be
```
ricardozanini, line 582:
```suggestion
Since there is no `toStateData` attribute and the result is not a JSON object but a string, the model would be:
```

specification.md
```diff
@@ -565,6 +565,27 @@ into. With this, after our action executes the state data would be:
 }
 ```
 
+To  illustrate merge of not json both object,  let`s assume that, in previous example, the action definition is at follows
+
+```json
+"actions":[
+    {
+       "name": "fetch_only_pasta",
```
ricardozanini, line 573:
```suggestion
       "name": "fetch-only-pasta",
```

specification.md
```diff
@@ -565,6 +565,27 @@ into. With this, after our action executes the state data would be:
...
+       }
+    }
+ ]
+}
+```
+Since there is not `toStateData` and the result is not a json object but an string, the model would be
+
+```json
+{
+  "fetch_only_pasta_output": "spaghetti"
```
ricardozanini, line 586:
```suggestion
  "fetch-only-pasta-output": "spaghetti"
```

ricardozanini (commented):
Great! A few comments regarding style and grammar only.

fjtirado:
@cdavernas please take another look

specification.md
fjtirado: Opened https://github.com/serverlessworkflow/specification/issues/837 to follow up this

cdavernas (approved):

PR #833, opened 2024-03-18
[Fix #686] Remove sleep state

### LINKED ISSUE #686: Get rid of obsolete Sleep state and replace it with an action

Get rid of obsolete Sleep state and replace it with an action. The state has little to no use IMO, especially since the addition of the sleepBefore/sleepAfter properties to the actionDef.

What I propose is to remove the sleepBefore/after properties and the sleep state, and replace them with an action of type sleep or delay.

```yaml
...
actions:
  name: Sleep
  delay: PT30M
...
```

**Why is this needed**:

What the sleep state achieves can be done using the sleepBefore/after properties, which IMO are very wierd in the context of an action. As a matter of fact, it should not be the responsibility or in the domain of the action to delay its own processing or its return. A better way is IMO to achieve that feature with an EXPLICIT action, which will therefore be visible in visualization tools, making the spec more ubiquitous, clean and concise.

fjtirado:

- [x] Specification
- [x] Schema
- [x] Examples
- [ ] Extensions
- [ ] Roadmap
- [ ] Use Cases
- [ ] Community
- [ ] TCK
- [ ] Other

**Discussion or Issue link**:

Fixes https://github.com/serverlessworkflow/specification/issues/686

Removing sleep state, as it is redundant with sleep after and before

**Additional information:**

### DIFF
--- a/examples/book-lending.json
+++ b/examples/book-lending.json
@@ -5,6 +5,7 @@
     "version": "1.0.0",
     "specVersion": "0.8",
     "start": "book-lending-request",
+    "constants" : { "WAIT_BEFORE_POLL" : "PT2W"},
     "states": [
         {
             "name": "book-lending-request",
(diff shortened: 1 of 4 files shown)


ricardozanini (approved):

cdavernas (approved):
LGTM! Cheers!

Issue #843, opened 2024-05-16
Upcoming new DSL for 1.0.0-alpha1 version release


ricardozanini:
**What would you like to be added**:
This umbrella issue is for discussions and links to previous issues, summarizing the upcoming new DSL.

- **Version 0.9.0** will be released as an evolution of 0.8 and will be maintained as long as the community shows interest. It will serve as a transitional version for runtimes already stable on 0.8. Adding new features to it will not be a priority, and it will be treated as a legacy version.

- **Version 1.0.0-alpha1** is a comprehensive refactor of the entire specification, retaining only standards integration. Its core will undergo significant changes, and the interpretation of nodes will be substantially altered. Many open issues will be addressed in this new version, allowing us to close the existing ones, remove them from the 0.9 milestone, and link them here. This will provide stakeholders with a clear understanding of the roadmap.

cdavernas:
Closed by #847

Issue #838, opened 2024-04-29
Add namespace property


fjtirado:
**What would you like to be added**:

It would be nice to add an optional `namespace` property to the specification

**Why is this needed**:
As per latest changes, `name` and `version`uniquely identify a workflow definition. However, it would be nice to group workflow definitions into namespaces, eventually allowing name repetition.

Issue #831, opened 2024-03-14
Use maps instead of arrays for complex objects in the specification


ricardozanini:
**What would you like to be added**:
For the long term, we should use maps (key/value) instead of arrays when listing complex objects. For example:

```yaml
states:
    - name: state-1
        actions:
        - name: action-1
    - name: state-2
    - name: state-3
```

Although it looks fine when using tools such as `JsonPointer` or JSON path to find items in the definition it may pose a challenge: `states/0/actions/0/functionRef`. One must rely on the array to find the objects. If the flow changes and a state or action changes the original order, the queries will be jeopardized.

Instead, we can use maps:

```yaml
states:
    state1:
        action1:
    state2:
    state3:
```

The query then can be pragmatic, order won't matter: `states/state1/action1`.

Additionally, implementation-wise, a few languages may not guarantee a correct order when serializing/deserializing a workflow. Using maps will definitely get rid of such a risk.

**Why is this needed**:
See context: https://github.com/serverlessworkflow/specification/pull/820#discussion_r1523718212

cdavernas:
On a side note, we will I believe need to remove naming restrictions on actions and states, as they will be naturally enforced by property naming rules.

Issue #828, opened 2024-03-11
Rename action `functionRef` property


cdavernas:
**What would you like to be added**:

Now that we split `eventRef` into two new `subscribe` and `publish` properties, we should rename the `functionRef` into something like `call`, so that we aim at consistently using imperative tense.
We should also rename the `functionRef`'s `refName` property into `function`.

```yaml
actions:
- name: my-action
  call:
    function: greet
    arguments:
      greetings: Hello, World!
```

**Why is this needed**:

- Use cleaner, simpler terminology accross the spec, and avoid at all cost using abbreviations.
- Align with https://github.com/serverlessworkflow/specification/pull/826.

PR #847, opened 2024-05-21
1.0.0-alpha1

cdavernas:

- [x] Specification
- [x] Schema
- [x] Examples
- [x] Extensions
- [x] Roadmap
- [x] Use Cases
- [x] Community
- [x] TCK
- [x] Other

**Discussion or Issue link**:
#843

The version `1.0.0-alpha1` is one of the most impacting release of the history of the Serverless Workflow DSL. It aims at both drastically simplify the DSL and at completing it with many new features.

It is intended to be the last release containing breaking changes before the Serverless Workflow DSL `1.0.0` can see the day.

## What's new?

- Total refactor of grammar and semantics. See [motivation](#). 
*Addresses #672.*
- Moved all descriptive properties into the document top level property. Adds a new (required) `namespace` DNS Label property. 
*Addresses #838.*
- Complex objects use maps instead of arrays, thus enabling more intelligible JSON pointers. 
*Addresses #831.*
- All workflow components are now referencable. This allows getting rid of many cumbersome 'oneOf' definitions. 
*Partially addresses #691.*
- Merged the `state` and `action` concepts into `task`, getting rid of state nesting limitations. See [tasks](#).
- Merged `end` and `transition` into [flow directives](#). 
*Addresses #684.*
- Got rid of the concept of compensation. Compensation can easily be designed by branching and/or chaining tasks.
- Support task nesting, thanks to [composite tasks](#). 
*Addresses #413.*
- Allows calling functions over HTTP, GRPC, OpenAPI and AsyncAPI. See [calls](#). 
*Addresses #828.*
- Allows running custom processes, using containers, shell commands, scripts or workflows. See [processes](#).
- Supports both sequential and concurrent task execution modes. See [execution strategy](#).
- Offers fault tolerance features, using [try task](#) to catch and handle [errors](#) raised during execution.
- Enables retrying on specific errors. See [retry policy](#). 
*Addresses #681 and #772.*
- Supports timeouts, used to define a duration after which to raise a new timeout error, possibly faulting the task/workflow. See [timeouts](#).
- Allows raising errors. See [errors](#errors). 
*Addresses #771.*
- Added a serie of Gherkin features used for both conformance testing and Behavior Driven Design (BDD). 
*Addresses #324.*
- Supports concurrent event correlations. 
*Addresses #679 and #680.*
- Enables extending and overriding the DSL, in a runtime agnostic and fully portable fashion, without writting a single line of code. See [extensions](#).


cdavernas:
This is a rebase of previous PR, which had gone critically wrong with signoffs.

ctk/features/for.feature
```diff
@@ -0,0 +1,35 @@
...
+    document:
+      dsl: 1.0.0-alpha1
+      namespace: default
+      name: for
+    do:
+      forEachColor:
+        for:
+          each: color
+          in: '.colors'
+        do:
```
fjtirado, line 21: Can be several "dos" being performed for every iteration of the loop?
If so, how can the second "do" refer to the result of the first "do"?

ctk/features/flow.feature
```diff
@@ -0,0 +1,61 @@
...
+    Given a workflow with definition:
+    """yaml
+    document:
+      dsl: 1.0.0-alpha1
+      namespace: default
+      name: explicit-sequence
+    do:
+      setRed:
+        set:
+          colors: '${ .colors + [ "red" ] }'
```
fjtirado, line 43: As discussed in slack, I believe it will be easier to write `colors: .colors +  [ "red "],` if we assume that by default string values are always expressions.

ctk/features/switch.feature
```diff
@@ -0,0 +1,137 @@
...
+
+  Scenario: Switch task with matching case
+    Given a workflow with definition:
+    """yaml
+    document:
+      dsl: 1.0.0-alpha1
+      namespace: default
+      name: switch-match
+    do:
+      switchColor:
```
fjtirado, line 14: I do not see an if-else  construct, although it can be emulated by
```
switch : 
    when: '.color == "red"'
    then: redAlert
    when: 'color != "red"'
    then: noWorries
```
I feel it would be better to do something like (which also cover the swith case)
condition:
      if: 'color=="red"'
     then: redAlert
     elseif: 'color=="green"'
      then: noWorries
     else: 
      then: willBeGoodToTakeALook

dsl.md
```diff
@@ -0,0 +1,96 @@
...
+
+Additionally, workflows may optionally accept inputs and produce outputs, allowing for data processing and transformation within the workflow execution.
+
+Workflows in the Serverless Workflow DSL can exist in several phases, each indicating the current state of the workflow execution. These phases include:
+
+| Phase |	Description |
+| --- | --- |
+| `Pending` |	The workflow has been initiated and is pending execution. |
+| `Running` |	The workflow is currently in progress. |
+| `Suspended` |	The workflow execution has been paused or halted temporarily. |
```
fjtirado, line 55: I actually prefer the name `waiting` because the workflow is waiting for something to happen (tipically the reception of the event).

ctk/features/try.feature
```diff
@@ -0,0 +1,81 @@
...
+  # Tests error instance path
+  Scenario: Try Handle Caught Error
+    Given a workflow with definition:
+    """yaml
+    document:
+      dsl: 1.0.0-alpha1
+      namespace: default
+      name: try-catch-404
+    do:
+      tryGetPet:
```
fjtirado, line 18: I see that we are using names for some "dos", are those optional or mandatory? 
And which is the purpose? 
I see "tryGetPet" name convenient to track progress of the workflow. With that spirit, probably when waiting for an event, the name should be mandatory (to be able to report where the workflow is actually waiting)

ctk/features/for.feature
cdavernas: No, you can only perform a single do. However it can be a composite task made out of multiple subtasks.

ctk/features/switch.feature
cdavernas: I did that for the sake of (JsonPointers) references, as it's easier to reference a named case than an indexed one.

dsl.md
fjtirado: yes, thats it, I think waiting reflect better the real state of the workflow

cdavernas: In this case, the name is required, as it identifies a task amongst others. For/Do does not require name cause it defines a single task.

ctk/features/for.feature
cdavernas: Does that resolve the conversation?

ctk/features/flow.feature
cdavernas: I thought a lot about it, and I think it's a very oppiniated matter, from both ends. As discussed in slack, I do think your proposal makes a LOT of sense. I also believe, especially because it is opiniated, that the matter is quite subtle, and could be addressed in a countless amount of ways.

As a matter of consequence, I propose we open an issue addressing this specific subject and solve it for `alpha2`, so we don't diverge as of now on what was being done regarding that. WDYT?

ctk/features/try.feature
fjtirado: what the $err object will contain if a do in a for fails?

ctk/features/switch.feature
fjtirado: Im not sure conditions should be referenced (task I agree they should). Anyway I think the most important part of my comment is the altarnative syntax, which is a superset of the current one.  if you prefer to keep the current one, I think adding a `default:` (equivalent to `else`) to the current switch will do

ctk/features/try.feature
cdavernas: You mean, as `instance`? Yeah, I forgot to document that.

So far, in my runtime tests, I'm using the following convention: `/for/{index}/do`.

We could work with anything else, too. It just looked pretty to me 😆

ctk/features/flow.feature
fjtirado: Sure I will open an issue with the different proposals

ctk/features/for.feature
fjtirado: yes, for some reason Im unable to resolve the comment ;)

ctk/features/try.feature
fjtirado: no, its fine. I was actually thinking if we have an alternative to optionally skip the task names overall (not just for foreach), so, when the task name is present we use it for error.instance, but if its not there, we use just a generated, but predictable, name. For example if the second task in the workflow throw an error and the task is unname, instance will be `/do/2 `

ctk/features/switch.feature
cdavernas: I personnally prefer the current state, that I find clearer, but that's just that, personnal preferences.

As for the `default` state, it exists. At most one case can not define a `when`. You can name it however you want, though `default` should be prefered for clarity.

Not defining a default ("whenless") case is the same than saying you want to do the parent switch task's `then`.

This is documented by the scenarii `Switch task with implicit default case` and `Switch task with explicit default case` of the `switch.feature`

ctk/features/try.feature
cdavernas: It's believe a very valid point! 

I'm wondering if the auto generated task name is DSL specific or rather interpreter/UI specific stuff? 

In my opinion, when present in a **collection** of tasks (as in `workflow.do` or `execute.sequentially` or `execute.concurrently`), the name should be required, in whichever way is most convenient to the interpreter/UI, as it's very opinated (some would maybe prefer naming it task-1 rather than task1 or even name it thanks to a zero based index, etc.).

ctk/features/switch.feature
fjtirado: ok, an extra condition with no when

cdavernas: Yes, exactly!

fjtirado: yes resolved

ctk/features/try.feature
cdavernas: the problem I see there is you consider the following:

```yaml
do: 
  execute:
    name: composite # does not work, execute is a collection of tasks
    try: 
       name: tryWithGet
       call: 
    try1:  #we are force to use another name than try, which is already defined. but task1 is not a valid task type.
       name: tryWithGet1
       call: 
```

fjtirado: this drives me to next point, if name is optional, execute should be an array, not an object ;)

fjtirado: but we can discuss on a different issue

ctk/features/flow.feature
fjtirado: https://github.com/serverlessworkflow/specification/issues/848

ctk/features/try.feature
fjtirado: Thats intuitive for me, if user does not write a name, instance is /do/execute/1 (that predictable generated name), if user wants a name there, he can write it

ricardozanini (approved):
I think we should move on and handle the minors in another PR.

@cdavernas do you agree that we need to convert the examples into the new DSL and keep them here?

If so, we can open a new issue and do so.

ctk/features/try.feature
fjtirado: Anyway, lets close this (in current spec user is forcer to write names for anything anyway), Ill open an issue to track it

cdavernas:
@ricardozanini Not necessarilly, as they are given by the features. Those might be difficult to find for newcomers, though, intimidated by ctk, features, ...

But if you feel we should, then I'll get going before merging this one.

ricardozanini:
@cdavernas I can send a PR on top of yours to fix the examples if you want, but tomorrow. Today my day is packed already :D

dsl.md
```diff
@@ -59,37 +81,395 @@ Workflows in the Serverless Workflow DSL can exist in several phases, each indic
...
+#### Components
+
+Serverless Workflow DSL allows for defining reusable components that can be referenced across the workflow. These include:
+
+- [Authentications](dsl-reference.md#authentication)
+- [Errors](dsl-reference.md#error)
+- [Extensions](dsl-reference.md#extension)
+- [Functions](dsl-reference.md#task)
+- [Retries](dsl-reference.md#retry)
+- [Secrets](#)
```
ricardozanini, line 95: Missing anchor

cdavernas:
@ricardozanini I added the first example, to boostrap the PR you spoke about.

In the mean time, should I merge?

ricardozanini:
@cdavernas sure, let's merge if you have the fup issues to handle the pending tasks we have. 👍

cdavernas:
Created #850, #851 and #852 which are the main requirements, aside from all optimizations, corrections, fixes, etc.

cdavernas:
Closes #847

PR #869, opened 2024-05-30
Illustrate do proposal.

fjtirado:
Just opened this one in the context of a slack discussion. Its not intended to be merged or reviewed and will be closed once we end the discussion. 
This will allow writing (in main flow, loops and retries)
Either unnamed single tasks (this assumes singles tasks can be univocally referenced through json pointers, which I believe is a valid one) 
```
do:
   set:
```
on named multiple tasks (multiple tasks needs to be named to be reference univocally) 
```
do: 
  task1: 
    set:
do:
  task2: 
    call: http

```

This will prevent writing
```
do:
  set: 
    set: 
```
which is valid now for main flow (it defines a task named set which invoke set tasks)

### DIFF
--- a/schema/workflow.yaml
+++ b/schema/workflow.yaml
@@ -72,11 +72,7 @@ properties:
         description: The workflow's secrets.
     description: Defines the workflow's reusable components.
   do:
-    type: object
-    minProperties: 1
-    additionalProperties:
-      $ref: '#/$defs/task'
-    description: Defines the tasks the workflow must perform
+    $ref: '#/$defs/do'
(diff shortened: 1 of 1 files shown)


matthias-pichler-warrify:
Is this event still relevant with #875 ?

fjtirado:
@matthias-pichler-warrify Partially , because whatever the do: schema finally is, it should be the same both for main flow, loop and retries. I mean, if we allow sequential execution of task for main flow though do: list of task, in loops and retries we should allow a list of tasks.

PR #879, opened 2024-06-03
Fix #878 - Fix URI format in workflow schema for endpoint attribute

ricardozanini:
**Please specify parts of this PR update:**

- [ ] Specification
- [x] Schema
- [x] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:

Fixes #878

PR #880, opened 2024-06-03
Support URIs with interpolation

### LINKED ISSUE #878: Fix `endpoint` URI format in workflow SCHEMA
**What seems off**:
The Workflow Schema uses the `uri` format in the `endpoint` property. This format is not valid for template URI like `https://petstore.swagger.io/v2/pet/{petId}`

**What you expected to be**:
That either URI and URI template passes the schema tests. See https://json-schema.org/understanding-json-schema/reference/string#resource-identifiers

**Anything else we need to know?**:
See GHA failure from #865 PR.

**Environment**:

- Specification version used:
1.0.0-alpha1

matthias-pichler-warrify:

- [ ] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:

Closes #878 

**What this PR does**:

I added a custom `uri` definition that uses the Regex from the URI RFC to validate usi and therefore allows curly braces `{}`

**Additional information:**

I replaced all occurences of uri ... even those for event listers, externalResources etc ... is that desired?


cdavernas:
> I replaced all occurences of uri ... even those for event listers, externalResources etc ... is that desired?

Yes, you rule! Thank you!

matthias-pichler-warrify:
I saw: https://github.com/serverlessworkflow/specification/pull/869 use `format: uri-template` which I didn't even know existed but seems more correct?

ricardozanini:
@cdavernas I don't think that's desired. JSON Schema already has the `uri-template` format that we can leverage, so only the `endpoint` attribute requires that the URI has a template, not all of them. I went through each URI field and I believe only that one is a valid template one?

Also, `uri-template` also validates `uri`.

ricardozanini:
> I saw: #869 use `format: uri-template` which I didn't even know existed but seems more correct?

That's #879 

The idea is to use `format: uri` for fixed URIs and `format: uri-template` for the ones we need templating, so leveraging directly from the RFC instead of bringing the regex it's safer IMO.

matthias-pichler-warrify:
> @cdavernas I don't think that's desired. JSON Schema already has the `uri-template` format that we can leverage, so only the `endpoint` attribute requires that the URI has a template, not all of them. I went through each URI field and I believe only that one is a valid template one?
> 
> Also, `uri-template` also validates `uri`.

Jap I am going to close this in favor of #879

PR #882, opened 2024-06-03
Remove reliance on key order in maps/objects

### LINKED ISSUE #875: Remove reliance on key order in "do"

I propose to:
  1. only use the `continue` FlowDirective to continue a loop
  2. make `then` required in all tasks
  3. add a `start` field to the workflow that denotes the first task to execute

```yaml
document:
  dsl: '1.0.0-alpha1'
  namespace: test
  name: order-pet
  version: '1.0.0'
  title: Order Pet - 1.0.0
  summary: >
  # Order Pet - 1.0.0
  ## Table of Contents
  - [Description](#description)
  - [Requirements](#requirements)
  ## Description
  A sample workflow used to process an hypothetic pet order using the [PetStore API](https://petstore.swagger.io/)
  ## Requirements
  ### Secrets
   - my-oauth2-secret
use:
  authentications:
    petStoreOAuth2:
      oauth2: my-oauth2-secret
  functions:
    getAvailablePets:
      call: openapi
      with:
        document:
          uri: https://petstore.swagger.io/v2/swagger.json
        operation: findByStatus
        parameters:
          status: available
  secrets:
  - my-oauth2-secret
start: getAvailablePets
do:
  getAvailablePets:
    call: getAvailablePets
    output:
      from: "$input + { availablePets: [.[] | select(.category.name == "dog" and (.tags[] | .breed == $input.order.breed))] }"
    then: submitMatchesByMail
  submitMatchesByMail:
    call: http
    with:
      method: post
      endpoint:
        uri: https://fake.smtp.service.com/email/send
        authentication: petStoreOAuth2
      body:
        from: noreply@fake.petstore.com
        to: ${ .order.client.email }
        subject: Candidates for Adoption
        body: >
        Hello ${ .order.client.preferredDisplayName }!

        Following your interest to adopt a dog, here is a list of candidates that you might be interested in:

        ${ .pets | map("-\(.name)") | join("\n") }

        Please do not hesistate to contact us at info@fake.petstore.com if your have questions.

        Hope to hear from you soon!

        ----------------------------------------------------------------------------------------------
        DO NOT REPLY
        ----------------------------------------------------------------------------------------------
    then: end
```

**Why is this needed**:

Currently the task execution order is determined by the order in which the tasks appear in the "do" field. 
This is very convenient for writing _small_ workflows.
I however see two mayor drawbacks with this approach:

1. similar problems as with a "switch fallthrough". If one forgets a "then" field the execution might resume at a task that is different from the one I expected.

Another interesting case is that the `then` field is allowed top level in `switch` tasks. So The way I understand it, this would execute the `processElectronicOrder` task when the default case is reached because it sets a `continue` clause. 

```yaml
document:
  dsl: '1.0.0-alpha1'
  namespace: test
  name: sample-workflow
  version: '0.1.0'
do:
  processOrder:
    switch:
      case1:
        when: .orderType == "electronic"
        then: processElectronicOrder
      case2:
        when: .orderType == "physical"
        then: processPhysicalOrder
      default:
        then: continue
  processElectronicOrder:
    execute:
      sequentially:
        validatePayment: {...}
        fulfillOrder: {...}
    then: exit
  processPhysicalOrder:
    execute:
      sequentially:
        checkInventory: {...}
        packItems: {...}
        scheduleShipping: {...}
    then: exit
  handleUnknownOrderType:
    execute:
      sequentially:
        logWarning: {...}
        notifyAdmin: {...}
```

4. reliance on object keys forces implementers to alway process/store definitions as text blobs or strings. 
Some problems this causes:
  - One cannot store the definition as a JSON object in SQL or NoSQL databases as they usually do not preserve Object key order. Thus one cannot use built-in filtering in databases
  - When building an API to create workflows the definition has to be accepted as a string or file blob then parsed and validated with the JSON schema and then the parsed object cannot be used because key order is not preserved. This might look something like this in TypeScript:

```typescript
function parse(input: string): IDefinition {
  const obj: Workflow = JSON.parse(input);

  validate(obj);

  // parse document as map
  const map: Map<"do", Map<string, Task>> = yaml.parse(input, {
    mapAsMap: true,
  });
  // get task names in order of appearance
  const taskNames = Array.from(map.get("do")?.keys() ?? []);
  const tasks = new Map<string, Task>();
  taskNames.forEach((name) => {
    // get task as object instead of Map
    const task = obj.do[name];
    if (task) {
      tasks.set(name, task);
    }
  });

  return { ...obj, do: tasks, text: input };
}
```

and this only works because `Map` keys preserve order of insertion, assuming that `yaml` never changes the implementation that would change the insertion order.

3. when providing a UI to build/visualize workflows the logic to create a valid and correct workflow definition might become incredible complex because on has to be precise about the ordering of keys, handle exit and fallthrough cases etc...

matthias-pichler-warrify:

- [x] Specification
- [x] Schema
- [x] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [x] CTK
- [ ] Other

**Discussion or Issue link**:

Fixes #875

**What this PR does**:

**Additional information:**


fjtirado:
I think we should also change do schema for retries and loops

matthias-pichler-warrify:
> I think we should also change do schema for retries and loops

what about `try`?

fjtirado:
I mean changing it here https://github.com/serverlessworkflow/specification/blob/main/schema/workflow.yaml#L526 and here https://github.com/serverlessworkflow/specification/pull/882
Probably it is better to move do: under $ref

matthias-pichler-warrify:
> I mean changing it here https://github.com/serverlessworkflow/specification/blob/main/schema/workflow.yaml#L526 and here https://github.com/serverlessworkflow/specification/blob/main/schema/workflow.yaml#L354 Probably it is better to move do: under $ref

yes already done 👍 but `try` also takes a task reference ... should that become an array as well? 

https://github.com/serverlessworkflow/specification/blob/0a48db935e8b226218ccb9946440126fb43f307f/schema/workflow.yaml#L507

fjtirado:
@matthias-pichler-warrify 
Yes, in my opinion, in any place where a task is allowed, multiple tasks should be allowed too.

schema/workflow.yaml
```diff
@@ -103,6 +103,9 @@ $defs:
   task:
     type: object
     properties:
+      name:
+        type: string
+        description: The task's name. **SHOULD** be unique within the workflow.
```
fjtirado, line 108: I guess should be unique within the array.

matthias-pichler-warrify:
@fjtirado where does the `if` come from? Is this a mistake?

https://github.com/serverlessworkflow/specification/blob/2ac27be0f209cf8d2557ca55194aa5f7e9eea684/dsl-reference.md?plain=1#L560-L565

fjtirado:
@matthias-pichler-warrify Looks like an unintentional mistake. @cdavernas

matthias-pichler-warrify:
> @matthias-pichler-warrify Yes, in my opinion, in any place where a task is allowed, multiple tasks should be allowed too.

I will also update `before` and `after` of extensions then 👍 

By the way ... I guess the order of extensions matters as well? 🤔

cdavernas:
> should that become an array as well?

No it should not and @fjtirado know uts something on which there are diverging opinions

cdavernas:
> I will also update before and after of extensions then 👍

By the way ... I guess the order of extensions matters as well? 🤔

Neither, or the purpose composite task is defeated.

The irder us indeed important, in a pipeline perspective

cdavernas:
If is a runtime expression yhat déterminés whether or not the task should be run ir skipped

matthias-pichler-warrify:
> > I will also update before and after of extensions then 👍
> 
> By the way ... I guess the order of extensions matters as well? 🤔
> 
> Neither, or the purpose composite task is defeated.
> 
> The irder us indeed important, in a pipeline perspective

ok then:

- extensions will be an array 👍 
- `before` & `after` will be a single task and one should use `composite` if needed. understood

what about `try` and `for.do`? should they be an array or single task?

cdavernas:
> but its nowhere defined ... not in the schema or the dsl

Indeed, it seems it left it out. It was a feature already supported in older versions.

ricardozanini:
@cdavernas 
> Indeed, it seems it left it out. It was a feature already supported in older versions.

Do we have an issue for this one?

fjtirado:
In my opinion, 
before and after should be single task
but loops if we are using do:, should be arrays. For two reasons: 
1) consistency, if as a writer I see a do: Im expecting do to behave equal in all places where is used, 
2) because loop iterations, for non trivial workflows, will consisnt on multiple tasks
Anyway, just to find a compromise, I guess that if we are using do, we can include an array of task and if a single task we can put the task just after the for, without the do. 
But to use do: with two different shchemas just because one is on the top and other is under a loop and base of the personal preference (or feeling) that most loops are going to be a single task is unlogical to me.

cdavernas:
@fjtirado That's your take on it, I respect it, we all do, and that's why you opened a PR for your proposal, which I personnaly find excellent.
Regardless, this is not the objective of that PR, which aims at solving ordering **bug**. 
And as said, I've got a different perspective that I also documented multiple times. 
Please let's move that discussion to #869, and take a vote. I'll go happilly with whichever we can get a consensus on.

cdavernas:
For the fifth time, whatever you think your proposal addresses (the main argument being defeated by renaming for/do), this is **NOT** what this PR addresses.

matthias-pichler-warrify:
> @cdavernas Once we move from map to array, the main justification for single task loop (not need to name it) is not there, so keeping the inconsistence in the do: schema is not logical Just for understanding, keeping that incosisntency, is like when defining C Ritchie was forcing the usage of { } (our do: ) in `for` and `while` and allowing just one statement there. In reality, they allow one statement without {} and if you want multiple statement you have to put {}. So either we remove the do: and just use execute: for composition everywhere (include main workflow) or we accept that our do: is indeed equivalent to quick composition (a sequence of sequential task) and in any place where do: appears the user should write a sequence of task. Hope this clarifies this is not a question of personal preference, but internal schema consistency. Now that we are changing the do: schema, this is the PR to be used.

I think your reasoning makes sense. Loops will probably have multiple tasks in them most of the time. I do not have a strong opinion on that so I will go with whatever is decided upon. However I would like to move forward with this PR and I do not think that a timely consensus will be reached here so adding a new issue to 1.0.0-alpha2 for the `for.do` proposal would be my preference 👍

fjtirado:
@matthias-pichler-warrify 
I also have to implement some stuff and for me is quite important that we have a single do: schema, so I cannot, by any means, accept a release cut and start discussing the do thing during ages. 
do:  schema MUST be consistent.

cdavernas:
> I also have to implement some stuff and for me is quite important that we have a single do: schema, so I cannot, by any means, accept a release cut and start discussing the do thing during ages as part of another release (the same you cannot deal with an unsorted map, which, by the way, I can).
do: schema MUST be consistent. If you want single task for loops, use a different word than do;

Sorry this pains you, I really am, but, again, we heard you on all possible channels. Again, whatever you think/need, it's not a bug. 

Please refer to #869 and to https://github.com/serverlessworkflow/specification/discussions/884. Hopefully we can reach a consensus in time for you to stop flooding all issues/prs.

fjtirado:
@cdavernas 
I think the current discussion is a bug is a matter of opinion I do not really want to dicuss. The reality is that PR is named make do: an array and you are asking to only make main flow do: an array, while keeping the other do: a task. The discussion makes fully sense in the context of the current PR. And I think we need to vote now on that.

fjtirado:
In other words, @cdavernas and @ricardozanini, you discuss if do: should have two different schemas depending on the position, take a decission and thats it, we do not need any further discussion on that topic.

fjtirado:
@matthias-pichler-warrify @ricardozanini  @cdavernas @JBBianchi 
It seems we reach consensus to implement https://github.com/serverlessworkflow/specification/discussions/884#discussioncomment-9662801
Thanks for your patience.

cdavernas:
@ricardozanini 

> Do we have an issue for this one?

Just created #885

ctk/features/composite.feature
cdavernas, line 1: Shouldn't it be `do/execute/sequentially`? Otherwise that's a PITA oneOf, am I wrong @fjtirado ?

cdavernas: It's a question, non-blocking imo.

JBBianchi: Not sure I follow. Dropping `execute` altogether has been discussed in https://github.com/serverlessworkflow/specification/discussions/884#discussioncomment-9662801. 
From my point of view, as a user, typing `do/execute/<sequentially|concurrently|compete>` is a bit cumbersome/redundant, I'd rather go for `do/<sequentially|concurrently|compete>` straight away.
As for the `oneOf`, aren't task and composite task already `oneOf`s ? It's "only" merging the two.

fjtirado: Hmmm, not, I think is ok the way it is. 
do: call: 
do: sequentially:
do: try: 
do: switch: 
do: set:

cdavernas: > Dropping execute altogether has been discussed in https://github.com/serverlessworkflow/specification/discussions/884#discussioncomment-9662801.

I overlooked the dropping of execute.

>  It's "only" merging the two.

Then resulting object can be both any task or a sequence/parallel execution of it?

Also, AFAIK, there's only `sequentially|concurrently`,  `compete` is merely a boolean that can be set when `concurrent` strategy has been chosen.

fjtirado: @matthias-pichler-warrify As @cdavernas pointed out, maybe is better to model `compete`as an atribute of `concurrently`

ricardozanini: @cdavernas having `compete` as a "task" makes sense to me since it's simpler than setting a boolean to `concurrently`. In practice, is the same thing and we save one property.

cdavernas: But it's redudant. Is merely a concurrency option. Imo that will be more confusing than anything. But that's just my opinion.

Whatever happens, that change belongs to another issue/PR, as it also requires updating the docs (as for now, it's a boolean). See https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#composite.

cdavernas: Additionally, `compete` is not using the same wording than `sequentially` or `concurrently`. It's a verb whereas we have been forced using adjectives here.

ctk/features/composite.feature
```diff
@@ -40,19 +38,16 @@ Feature: Composite Task
...
-              set:
-                colors: ${ .colors + ["red"] }
-            setGreen:
-              set:
-                colors: ${ .colors + ["green"] }
-            setBlue:
-              set:
-                colors: ${ .colors + ["blue"] }
-          compete: true
+      compete:
```
cdavernas, line 41: That is a shadow change. Please avoid doing so in a single PR, as it hasn't been approved.

cdavernas (changes requested):

ctk/features/composite.feature
JBBianchi: Doesn't having `compete` as a parameter of `concurrently` require an extra level ? As for now, `sequentially|concurrently|compete` are arrays of <key, task>, where would the `compete` parameter fit ?

cdavernas: As for now it's the other way around.

cdavernas: Compete as a strategy has been added here. And as said above, it does not follow the same wording. At best, it should be replaced with `competitively`

matthias-pichler-warrify: https://github.com/serverlessworkflow/specification/discussions/884#discussioncomment-9662801

shows

```yaml
do:
  sequentially: []
```

and mentions

> `composite` loses the `execute` keyword, we use `sequentially` | `concurrently` | `compete` straight away for a list of tasks

cdavernas: My take on it is to keep execute keyword as a task, and to rename compete to competitively or whatever, so we get rid of the additional prop.

ctk/features/composite.feature
matthias-pichler-warrify: I am not sure what you mean by "shadow change"?

We can keep the composite task as one using `execute` but it was agreed upon to remove `execute`. I mean as you said it is uglier with `execute`

ctk/features/composite.feature
cdavernas: so:

```yaml
do:
  execute:
    sequentially: {}
    concurrently: {}
    Competitively: {}
```

JBBianchi: I'm getting confused, why keep `execute` if there is `competitively` ?

ctk/features/composite.feature
cdavernas: I mean it's a change I overlooked and did not think was a part of this PR. It's my bad.

ctk/features/composite.feature
cdavernas: For one, I feel the following is very ugly:

```yaml
do:
  - aTask:
      sequentially:
        - subTask1:
            call: http
        - subTask2:
            competitively:
              ...
```

You see what I mean? `competitively` when used down the tree is odd: unlike all other tasks that are imperative, you now would have an out of context adjective.

And two, I'm concerned that we might need additional properties to configure the composite, as said above. Probably a stupid precaution. 

Anyways, won't change a lot on my end in the end, even with OneOfs.

I'll go with what you guys prefer!

JBBianchi: I didn't think about the ones "nested", it's indeed a bit odd in that case.

fjtirado: Again, oneOf is the same with or without execute, therefore lets get rid of execute.

fjtirado: for the nested case, we can use (although I do not really the need to, but if you insist)
```
do:
  sequentially:
    - subTask1:
        call: http
    - subTask2:
         do:
            competitively:
```

And do: will be the task.

fjtirado: Therefore, if you put do as task,  you can write a flow this way
```
document: 

call: 
```

JBBianchi: > for the nested case, we can use (although I do not really the need to, but if you insist)
> 
> ```
> do:
>   sequentially:
>     - subTask1:
>         call: http
>     - subTask2:
>          do:
>             competitively:
> ```
> 
> And do: will be the task.

But then, with that logic, you should always have `do`:
```yaml
do:
  sequentially:
    - subTask1:
        do:
          call: http
    - subTask2:
         do:
            competitively:
```

fjtirado: @JBBianchi See the examples in the PR, we already have the do:

cdavernas: I think the best way is to keep the execute, and transform compete into a task execution strategy.

```yaml
do:
  execute:
    sequentially: {}
    concurrently: {}
    competitively: {}
```

@matthias-pichler-warrify What's your take on that?

fjtirado: So it is actually the contrary, you only write do: before sequentially: concurently or compettively: for other cases you write call: try: raisea:

fjtirado: do: does not really appear in the schema, it will be just another task.

JBBianchi: To rephrase @fjtirado, `do` is holding a single task anyway, so it might not be necessary to keep it. When declaring composite tasks, we could then use the keyword `do` instead of `execute`.

So a single task would be declared such as (which, besides at the top level, is already the case):
```yaml
document:
  #...
call: http # single call task
with:
  method: get
  endpoint: https://petstore.swagger.io/v2/pet/{petId}
```

And composite tasks:
```yaml
document:
  #...
do: # 'single' composite task
  sequentially: # or concurrently or compete
    - bookHotel:
      call: http  # single call task
      with:
        method: post
        endpoint: 
          uri: https://fake-booking-agency.com/hotels/book
          authentication: fake-booking-agency-oauth2
        body:
          name: Four Seasons
          city: Antwerp
          country: Belgium
    - bookFlight:
      call: http  # single call task
      with:
        method: post
        endpoint: 
          uri: https://fake-booking-agency.com/flights/book
          authentication: fake-booking-agency-oauth2
        body:
          departure:
            date: '01/01/26'
            time: '07:25:00'
            from:
              airport: BRU
              city: Zaventem
              country: Belgium
          arrival:
            date: '01/01/26'
            time: '11:12:00'
            to:
              airport: LIS
              city: Lisbon
              country: Portugal
```

matthias-pichler-warrify: I am fine with:

```yaml
do:
  execute:
    sequentially: {}
    concurrently: {}
    competitively: {}
```

especially replacing `compete` as it's not an adjective 👍  I am also fine with:

```yaml
do:
  sequentially: {}
  concurrently: {}
  competitively: {}
```

Which I think is nicer to read.

I do not like:

```yaml
do:
  sequentially:
    - subTask1:
          call: http
    - subTask2:
         do:
            competitively: {}
```

On the topic of additional properties for `concurrently` does anything prevent us from doing:

```yaml
do:
  compete: true
  concurrently:
    - call1: {}
```

like it used to be in the `composite` task? (i.e. `compete` and `concurrently` being siblings?) ... other options that come to mind (for future additions) would be a `parallelism: 3` option that  limits max concurrency

matthias-pichler-warrify:
Not sure If this was visible enough, but I renamed `compete` to `competitively` according to the @JBBianchi 's reply [here](https://github.com/serverlessworkflow/specification/pull/882#discussion_r1626196634). 

But since the composite task used to have

```yaml
compTask:
  execute:
    compete: true
    concurrently: {}
``` 

I don't see why we couldn't keep this pattern and go with:

```yaml
do:
  compete: true
  parallelism: 3 # hypothetical future prop
  concurrently:
    - call1: {}
```

Previous comment 👇 

> I am fine with:
> 
> ```yaml
> do:
>   execute:
>     sequentially: {}
>     concurrently: {}
>     competitively: {}
> ```
> 
> especially replacing `compete` as it's not an adjective 👍 I am also fine with:
> 
> ```yaml
> do:
>   sequentially: {}
>   concurrently: {}
>   competitively: {}
> ```
> 
> Which I think is nicer to read.
> 
> I do not like:
> 
> ```yaml
> do:
>   sequentially:
>     - subTask1:
>           call: http
>     - subTask2:
>          do:
>             competitively: {}
> ```
> 
> On the topic of additional properties for `concurrently` does anything prevent us from doing:
> 
> ```yaml
> do:
>   compete: true
>   concurrently:
>     - call1: {}
> ```
> 
> like it used to be in the `composite` task? (i.e. `compete` and `concurrently` being siblings?) ... other options that come to mind (for future additions) would be a `parallelism: 3` option that limits max concurrency
> 
> Therefore I would go with:
> 
> ```yaml
> do:
>   compete: true
>   parallelism: 3 # hypothetical future prop
>   concurrently:
>     - call1: {}
> ```
> 
> because `do` is just a single task and a `composite` task allowed siblings to `sequentially` and `concurrently` before so I see no reason why it shouldn't now

fjtirado:
@matthias-pichler-warrify I think you are ruling out 
```
do:
  sequentially:
    - subTask1:
          call: http
    - subTask2:
         do:
            competitively: {}
```
too quickly (it deserver more that one "I do not like it") and I want to be sure that you are aware of the implication in the schema
```
Moving do: as task will allow writing
document:
  #...
call: http # single call task
with:
  method: get
  endpoint: https://petstore.swagger.io/v2/pet/{petId}
```
The schema will be more logically consistent in my opinion, you will have task in try, loop, main flow (basically everywhere and you do: is just synonimous to current execute:
So I need something more than just "I do not like it" ;)

matthias-pichler-warrify:
@fjtirado 
> I think it deserves more that one "I do not like it"

Sure I can elaborate:

What I like about the current workflow structure, is it gives me a clear entry point for each peace of information I want:

```yaml
document: {} # Metadata: name, version, dsl etc
input: {} # what is expected as input
use: {} # reusable components
do: {} # which actions are performed
timeout: {} # how long does it run
output: {} # what is returned
schedule: {} # when is it run
evaluate: {} # which kind of expressions can I expect
```

So if I want to know what task(s) is/are done I can fold everything and just open `do` 👍 

I have the following _issues_ with effectively replacing `execute` with `do` and allowing a top level task:

# Logical Inheritance

Allowing the properties of a task to reside at the top level of the `workflow` would mean that the schema for `workflow` would look something like

```yaml
type: object
$ref: "#/defs/task" # to allow task properties at the top level
properties:
  document: {} 
  input: {} 
  use: {} 
  timeout: {} 
  output: {}
  schedule: {} 
  evaluate: {}
```

Now this is, according to the JSON schema homepage: https://json-schema.org/blog/posts/modelling-inheritance , the de-facto standard way to model Inheritance.

I am not sure how generated types/schemas are handled in (all) OOP programing languages (I believe you are much better informed here) but at least conceptually this would mean `Workflow extends Task` and by the substitution principle an entire workflow could be used anywhere where a task is expected? also in `try`, `catch`, `for.do` ect ? This seems illogical to me.

> The schema will be more logically consistent in my opinion

So I disagree an this one. One is a task and the other a workflow. `try`, `catch`, `loops`, `extensions`  and a workflow all _have_ task(s) to run/execute/do/perform but only some of them are tasks themselves. So for me its a bit of Composition vs Inheritance thing and not everything has to _be_ a task (in my opinion).

# Readability

Assuming I receive a workflow document that looks like:

```yaml
then: exit
output: {} # what is returned
document: {} # Metadata: name, version, dsl etc
do: {} # task of the loop
input: {} # what is expected as input
schedule: {} # when is it run
while: .vet != null
use: {} # reusable components
timeout: {} # how long does it run
for:
  in: .pets
evaluate: {} # which kind of expressions can I expect
```

(I intentionally reordered the keys because this is what this PR is about after all).

Now it is way less clear where to find what. And suddenly `do` is not for sequential or concurrent but the loop task. Which I might completely miss do to the fact that `for` is near the end. And the loop statements `for` and `while` might be spread across the very beginning and end of the workflow ... not very clear to me. 

Furthermore `then` does not make any sense at the top level but suddenly becomes valid.

Also with simpler workflows, such as:

```yaml
document: {} # Metadata: name, version, dsl etc
call: http
input: {} # what is expected as input
use: {} # reusable components
with: {}
```

I find the appearance of `with` and `use` right next to each other less than pretty because it makes me stop every time to think: which one is for components and which one for function calls.

# Conclusion

> I think we should avoid by all means is do: execute: because both express the same thing (its a not needed redundancy)

I agree, which is why I liked the following variant better (having the 2 or 3 tasks):

```yaml
document: {}
do:
  sequentially: [] # or concurrently or competitively
```

and to address the point made by @cdavernas made [here](https://github.com/serverlessworkflow/specification/pull/882#discussion_r1626055636)

> Finally, I'm confident that the additional level will be required for declaring additional options to a task array.

I wanted to demonstrate that dropping `execute` and having (at least) `sequentially` and `concurrently` as tasks would still allow for additional options, namely as siblings of `concurrently` (or `sequentially` respectively), as demonstrated by:

```yaml
document: {}
do:
  compete: true
  parallelism: 3 # hypothetical future prop
  concurrently:
    - call1: {}
```

which incidentally is the exact same way it is already done with execute today:

```yaml
# old example using the dsl version 1.0.0-alpha1
document: {}
do:
  compositeTask:
    execute:
      compete: true
      concurrently: {}
```

Therefore I vote for:

```yaml
document: {}
do:
  sequentially: [] # or concurrently or competitively
```

with the option to remove `competitively` and replace it with `compete: boolean` as a sibling to `concurrently` (as it used to be)

cdavernas:
@matthias-pichler-warrify I totally agree with you.

I'd like however to call your attention on the discussion I had with @JBBianchi yesterday regarding `execute`, and how it looks funny if we get rid of it when nested down. 

@fjtirado I agree that it's not optimal having to repeat do/execute which are similar, but the alternative is worst: having call/wait/set... on the same level than smtg like concurrently is wierd and inconsistent.

I therefore vote for keeping the execute task as is.

matthias-pichler-warrify:
> @matthias-pichler-warrify I totally agree with you.
> 
> I'd like however to call your attention on the discussion I had with @JBBianchi yesterday regarding `execute`, and how it looks funny if we get rid of it when nested down.
> 
> @fjtirado I agree that it's not optimal having to repeat do/execute which are similar, but the alternative is worst: having call/wait/set... on the same level than smtg like concurrently is wierd and inconsistent.
> 
> I therefore vote for keeping the execute task as is.

Yes I saw that. I am ok with keeping `execute` although it is not my preferred variant. I saw your example:

```yaml
do:
  sequentially:
    - subTask1:
        call: http
    - subTask2:
        competitively: []
```

and personally do not think it is ugly. But that is a matter of taste.

cdavernas:
It's not ugly, but is less consistent. It's not a call to action, it introduces an inconsistency and is not a verb using imperative, as mandated by design guidelines. Plus, it doesn't mean much anymore when read out loud, especially of the task name is a bad one.

It's however a solution I can live with, even though I reiterate my strong preference for execute, for both above reasons and cosmetics.

matthias-pichler-warrify:
> It's not ugly, but is less consistent. It's not a call to action, it introduces an inconsistency and is not a verb using imperative, as mandated by design guidelines. Plus, it doesn't mean much anymore when read out loud, especially of the task name is a bad one.
> 
> It's however a solution I can live with, even though I reiterate my strong preference for execute, for both above reasons and cosmetics.

have we considered:

```yaml
document: {}
do:
  execute:
    - task1: {}
    - task2: {}
  in: order # in order or in parallel
```

then you could read it out load as "execute ... in order" or "execute ... in parallel" similar to "call ... with"

cdavernas:
Well, I had not considered it, no, good one. 

I'll give it time to sink and make some trials, then I'll come back to you with a documented opinion 😉

matthias-pichler-warrify:
> Well, I had not considered it, no, good one.
> 
> I'll give it time to sink and make some trials, then I'll come back to you with a documented opinion 😉

I do have to say that with all these renames, splitting tasks etc it feels like we have strayed quite a bit from the original issue this PR is trying to solve 🙈  (I know @cdavernas that you have brought this up multiple time already 😅 )

I am glad about the changes we've come up so far as I feel they made the initial solution much more elegant but I would really like to see the `do/execute` (or whichever variant we end up using) discussion resolved (with well thought through result) and then merge this PR 👍

fjtirado:
@matthias-pichler-warrify 
Thanks for the detailed explanation, but it unfortunately relies on a misunderstanding of the proposal. The schema for worklow I had in mind would actually be
```
document: {} # Metadata: name, version, dsl etc
input: {} # what is expected as input
use: {} # reusable components
timeout: {} # how long does it run
output: {} # what is returned
schedule: {} # when is it run
evaluate: {} # which kind of expressions can I expect
additionalProperties: 
   #/ref/task
```
Which is actually the same as we have by replacing the do: by a direct call to the task ;)
Anyway, I have not issue with 
```
do:
  sequentially:
    - subTask1:
        call: http
    - subTask2:
        competitively: []
```
I was just trying to provide an alternative the really ugly, at least for me, do: execute:

JBBianchi (approved):

fjtirado:
@matthias-pichler-warrify 
Ok, what about, just guessing
```
document: {} # Metadata: name, version, dsl etc
input: {} # what is expected as input
use: {} # reusable components
timeout: {} # how long does it run
output: {} # what is returned
schedule: {} # when is it run
evaluate: {} # which kind of expressions can I expect
$ref: #/$defs/task
```

matthias-pichler-warrify:
> Thats actually the point, you can invole a task without prefixing it with do:. You only need do: when you want multi task.

What about nested loops? The `do` would be needed to define the nested loop in because the properties would collide. But then defining `sequentially` inside the `do` block just for the nested loop?

cdavernas:
@fjtirado Stop and move comments to dedicated PR.

matthias-pichler-warrify:
> @matthias-pichler-warrify Ok, what about, just guessing
> 
> ```
> document: {} # Metadata: name, version, dsl etc
> input: {} # what is expected as input
> use: {} # reusable components
> timeout: {} # how long does it run
> output: {} # what is returned
> schedule: {} # when is it run
> evaluate: {} # which kind of expressions can I expect
> $ref: #/$defs/task
> ```

Still not a valid JSON schema 😂 

the only ways to merge the properties is:

```yaml
type: object
properties: 
  document: {}
  ...
$ref: "#/$defs/task"
```

or


```yaml
type: object
properties: 
  document: {}
  ...
allOf:
 - $ref: "#/$defs/task"
```

or


```yaml
allOf:
  - type: object
     properties: 
      document: {}
      ...
   - $ref: "#/$defs/task"
```

fjtirado:
@matthias-pichler-warrify 
Ok, thanks, so basically there is no smart way of achieving what I had in min (which you do not find appealing anyway)

Can we go for 
```
do:
  sequentially:
    - subTask1:
        call: http
    - subTask2:
        competitively: []
```
as we agreed yesterday?

fjtirado:
> @fjtirado please stop and move comments to dedicated PR.

Which is the dedicated PR?

matthias-pichler-warrify:
> @matthias-pichler-warrify Ok, thanks, so basically there is no smart way of achieving what I had in mind (which you do not find appealing anyway)
> 
> Can we go for
> 
> ```
> do:
>   sequentially:
>     - subTask1:
>         call: http
>     - subTask2:
>         competitively: []
> ```
> 
> as we agreed yesterday?

I am ok with that, however @cdavernas said they wanted to do some further experiments with alternatives to `concurrently`? or are we ready to move forward?

cdavernas:
The one I asked you more than 10 times to create!!!

fjtirado:
> The one I asked you more than 10 times to create!!!

My understanding is that we will be addressing the agreement with this PR. 
Also, it is imposible to create any further PR to restructure the execute without merging this one first.

cdavernas:
@matthias-pichler-warrify Please keep this issue only for the list vs map fix so that we can merge.
Let's open a PR for the renaming, because I feel well need more time to find a truly consensual solution to that one ;)

matthias-pichler-warrify:
> @matthias-pichler-warrify Please keep this issue only for the list vs map fix so that we can merge. Let's open a PR for the renaming, because I feel well need more time to find a truly consensual solution to that one ;)

@cdavernas Understood. So I should bring back composite as it was? But keep `do` a single task so that:

```yaml
document: {}
do:
  execute:
    compete: true
    concurrently: []
```

which is the agreed upon solution without the rename/split of the composite task 👍

cdavernas:
@matthias-pichler-warrify Yes! So we merge this one, find a renaming solution that fits our expectations, and merge that one in time, too.

fjtirado:
@cdavernas 
In the proposal that was agreed, execute; was not there. That was a fundamental part of the agreement that you are rolling back.

matthias-pichler-warrify:
> @cdavernas In the proposal that was agreed, execute; was not there. That was a fundamental part of the agreement that you are rolling back.

@fjtirado  there have been about 5 different solutions we agreed on in the meantime and @cdavernas is refering to the one we had before spliting composite

fjtirado:
@matthias-pichler-warrify @cdavernas 
The proposal that was agreed was this one https://github.com/serverlessworkflow/specification/discussions/884#discussioncomment-9662801

matthias-pichler-warrify:
> > In the proposal that was agreed, execute; was not there. That was a fundamental part of the agreement that you are rolling back.
> 
> * Wasnt agreed. Read again.
> * It's not a fundamental part of it
> * It's not even related to the issue

So I checked and technically @fjtirado is correct that the split was in the agreed upon variant: https://github.com/serverlessworkflow/specification/discussions/884#discussioncomment-9662801

however as it is outside the scope of the core issue and apparently not agreed upon I will proceed with only the first part of the agreed upon soliution (making do a single task)

fjtirado:
@matthias-pichler-warrify 
Ok, so @cdavernas decided that the agreement was not an agreement. 
Point taken for the future

cdavernas:
@matthias-pichler-warrify As said somewhere in the chapter, I had overlooked the removal of execute ;)

fjtirado:
ok, Should I open a new issue for the do: execute:?

matthias-pichler-warrify:
@cdavernas whenever you are ready 👍

JBBianchi:
> ok, Should I open a new issue for the do: execute:?

I'll try to sum up the ideas and open the issue tomorrow

cdavernas:
@matthias-pichler-warrify Thanks! Gonna review it first thing tomorrow morning, I'm off for tonight ;)
Many thanks for your awesome work

cdavernas:
@JBBianchi Yes, please! ❤️

ricardozanini (approved):

cdavernas (approved):
LGTM! Thanks for your work!

JBBianchi:
> > ok, Should I open a new issue for the do: execute:?
> 
> I'll try to sum up the ideas and open the issue tomorrow

Discussion open at https://github.com/serverlessworkflow/specification/discussions/889

Issue #873, opened 2024-05-31
Rename output.from and output.to


fjtirado:
Output.from expression indicates the result of the task (and can be consumed by subsequent task) 
Output.to expression indicates what should be the value of the context once the task is completed. 

I think the names are confusing (it actually takes me a lot of time to understand why do we have two expressions in output), so I think we must change them. 

output.from might be renamed to output.result (the expression indicates the result of the task)
output.to might be renamed to output.setContext (clearly states the purpose is to the set the value of the context), If we found that setContext break the single word paradigm that we are implicitly following in the DSL, then lets just use output.context. 
Also we need to explicilly indicate in the documentation that it is assumed that the expression language should support merging context.  
Related to that, if this assumption turns to be wrong, another possibility is to use
```
output: 
    context: 
        set:
```
Right now, context will only support set: but we leave the door open to add append:, delete: or merge: in future if needed

cdavernas:
@JBBianchi @matthias-pichler-warrify WDYT?

1. Renaming `output.from` to `output.result`
2. Renaming `output.to` to `output.context`
3. Creating two expression properties to describe the performed operation

Personally:

1. Fine by me, though I personally like the actual wording. As demonstrated by #837 and by @fjtirado admitedly late understanding by lack of info, it does however requires much more documentation.
2. Same than 1.
3. As expressed in #837, I personally prefer it the way it is now, meaning that the user has the ability to write the expression he wants (i.e. replace or update the context). I do think however that the proposal makes sense, so I'll rally to the consensus on that one.

JBBianchi:
Globally, more documentation won't hurt, for sure :)

If I can add my grain of salt, we might gain clarity by "splitting" `output.from` and `output.to` into distinct keywords. So my take would be the following:
1. Rename `output.from` to `output`, it shows the intention of the user to alter/transform the default output (but I have no strong opinion)
2. Rename `output.to` to `export`, it shows the intention of the user to export something from the task to a broader context (but  again, I have no strong opinion)
3. Relying on the expression rather than making the spec more complex seems a better solution to me, so I'd rather keep this as they are.

JBBianchi:
> > 1. Rename output.from to output, it shows the intention of the user to alter/transform the default output (but I have no strong opinion)
> > 2. Rename output.to to export, it shows the intention of the user to export something from the task to a broader context (but again, I have no strong opinion)
> 
> You suggestion is awesome! Much clearer and ubiquitous!
> 
[...]

PR #892, opened 2024-06-07
[Fix_#873] Rename output.from and output.to

fjtirado:
Fix https://github.com/serverlessworkflow/specification/issues/873
output.from is renamed as output.as
output.to is renamed as export.as

- [x] Specification
- [x] Schema
- [x] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:

**What this PR does**:

**Additional information:**

### DIFF
--- a/dsl-reference.md
+++ b/dsl-reference.md
@@ -176,7 +176,7 @@ do:
       - getAvailablePets:
           call: getAvailablePets
           output:
-            from: "$input + { availablePets: [.[] | select(.category.name == "dog" and (.tags[] | .breed == $input.order.breed))] }"
+            as: "$input + { availablePets: [.[] | select(.category.name == "dog" and (.tags[] | .breed == $input.order.breed))] }"
       - submitMatchesByMail:
           call: http
           with:
@@ -234,6 +234,7 @@ The Serverless Workflow DSL defines a list of [tasks](#task) that **must be** su
 |:--|:---:|:---:|:---|
 | input | [`input`](#input) | `no` | An object used to customize the task's input and to document its schema, if any. |
 | output | [`output`](#output) | `no` | An object used to customize the task's output and to document its schema, if any. |
+| export | [`export`](#export) | `no` | An object used to customize the content of the workflow context. | 
 | timeout | [`timeout`](#timeout) | `no` | The configuration of the task's timeout, if any. |
 | then | [`flowDirective`](#flow-directive) | `no` | The flow directive to execute next.<br>*If not set, defaults to `continue`.* |
 
@@ -558,7 +559,7 @@ do:
           with:
             type: com.fake.petclinic.pets.checkup.completed.v2
       output:
-        to: '.pets + [{ "id": $pet.id }]'        
+        as: '.pets + [{ "id": $pet.id }]'        
 ```
 
 #### Listen
@@ -1415,6 +1416,36 @@ from:
   petId: '${ .pet.id }'
 to: '.petList += [ . ]'
 ```
+
+### Export
+
+Certain task needs to set the context of the workflow using the task output for later usage. User set the content of the context through a runtime expression. The result of the expression is the new value of the context. If user want to merge the new data into the current context, he might do that using `$context` variable. 
+
+Optionally, the context can have a predefined schema. 
+
+
+#### Properties
+
+| Property | Type | Required | Description |
+|----------|:----:|:--------:|-------------|
+| schema | [`schema`](#schema) | `no` | The [`schema`](#schema) used to describe and validate context.<br>*Included to handle the non frequent case in which the context has a known format.* |
+| as | `string`<br>`object` | `no` | A [runtime expression](#runtime-expressions), if any, used to set the context value. |
+
+#### Examples
+
+Assuming the output of the task is a json object (not primitive, not array)
+Merge into the current context the output of the task. 
+
+```yaml
+as: '$context+=.'
+```
+
+Replace the context with the output of the task. 
+
+```yaml
+as: .
+```
+
 
 ### Schema
 
--- a/dsl-reference.md
+++ b/dsl-reference.md
@@ -44,6 +44,7 @@
   + [Retry](#retry)
   + [Input](#input)
   + [Output](#output)
+  + [Export] (#export)
   + [Timeout](#timeout)
   + [Duration](#duration)
   + [HTTP Response](#http-response)
@@ -1419,33 +1420,30 @@ to: '.petList += [ . ]'
 
 ### Export
 
-Certain task needs to set the context of the workflow using the task output for later usage. User set the content of the context through a runtime expression. The result of the expression is the new value of the context. If user want to merge the new data into the current context, he might do that using `$context` variable. 
+Certain task needs to set the workflow context to save the task output for later usage. Users set the content of the context through a runtime expression. The result of the expression is the new value of the context. The expression is evaluated against the existing context. 
 
-Optionally, the context can have a predefined schema. 
-
+Optionally, the context might have an associated schema. 
 
 #### Properties
 
 | Property | Type | Required | Description |
 |----------|:----:|:--------:|-------------|
 | schema | [`schema`](#schema) | `no` | The [`schema`](#schema) used to describe and validate context.<br>*Included to handle the non frequent case in which the context has a known format.* |
-| as | `string`<br>`object` | `no` | A [runtime expression](#runtime-expressions), if any, used to set the context value. |
+| as | `string`<br>`object` | `no` | A runtime expression, if any, used to export the output data to the context. |
 
 #### Examples
 
-Assuming the output of the task is a json object (not primitive, not array)
-Merge into the current context the output of the task. 
+Merge the task output into the current context. 
 
 ```yaml
-as: '$context+=.'
+as: '.+$output'
 ```
 
-Replace the context with the output of the task. 
+Replace the context with the task output. 
 
 ```yaml
-as: .
+as: $output
 ```
-
 
 ### Schema
 
--- a/schema/workflow.yaml
+++ b/schema/workflow.yaml
@@ -828,7 +828,7 @@ $defs:
         description: The schema used to describe and validate the workflow context.
       as:
         type: string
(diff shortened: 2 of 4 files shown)


dsl-reference.md
```diff
@@ -1416,6 +1417,36 @@ from:
 to: '.petList += [ . ]'
 ```
 
+### Export
+
+Certain task needs to set the context of the workflow using the task output for later usage. User set the content of the context through a runtime expression. The result of the expression is the new value of the context. If user want to merge the new data into the current context, he might do that using `$context` variable. 
+
```
cdavernas, line 1423: This is incorrect. Doing so would go against what was discussed.

The `export/as` expression is evaluated against the context. 
The user can access the input and the output of the task using `input` and `ouput` arguments.

cdavernas (changes requested):

dsl-reference.md
fjtirado: So . is the context, isnt it?

cdavernas: In that case - **AND ONLY THAT CASE** - yes.

So you would have:

```yaml
do:
  call: http
  with: {...}
  export:
    as: . + { responseStatusCode = $output.statusCode }
```

Given the context looked like this before the export:

```yaml
foo:
  bar: baz
```

And given the task output's `statusCode` property being 503;

After evaluating the context it would look like:

```yaml
foo:
  bar: baz
responseStatusCode: 503
```

That is why I was so reticent to rename `export/to` into `export/as`, because the expression itself is doing a `to`.

dsl-reference.md
```diff
@@ -1416,6 +1417,36 @@ from:
 to: '.petList += [ . ]'
 ```
 
+### Export
+
+Certain task needs to set the context of the workflow using the task output for later usage. User set the content of the context through a runtime expression. The result of the expression is the new value of the context. If user want to merge the new data into the current context, the context is accessible within the expression scope: `.` if using jq.
```
cdavernas, line 1422: Speaking of the `.` is irrelevant, because it means the same than anywhere else: its the expression's input.

I think you should leave `jq` out of it. I would remove the export section, and just describe the property with something like:

A [`runtime expression`](#runtime-expressions) evaluated again the workflow/task context data to update or replace it.

dsl-reference.md
fjtirado: I do not think how to access the context value (. being $context rather than  . being $output) is really related with the to/as. 
I think we need to clearly set what . is supposed to be for every expr, probably we should open a PR to do that as result of dicussion https://github.com/serverlessworkflow/specification/discussions/890

dsl-reference.md
```diff
@@ -1416,6 +1417,36 @@ from:
...
+#### Properties
+
+| Property | Type | Required | Description |
+|----------|:----:|:--------:|-------------|
+| schema | [`schema`](#schema) | `no` | The [`schema`](#schema) used to describe and validate context.<br>*Included to handle the non frequent case in which the context has a known format.* |
+| as | `string`<br>`object` | `no` | A [runtime expression](#runtime-expressions), if any, used to set the context value. |
+
+#### Examples
+
+Assuming the output of the task is a json object (not primitive, not array)
```
cdavernas, line 1436: Very good point. The workflow input and the context data are the only dynamic data that are REQUIRED to be a mapping.

However the context could have nothing to do with JSON. You could write/read it from any format. 

I would therefore reformulate the phrase like this, to keep the same linguo than in other places of the document:

```md
Assuming the output of the task is a mapping
```

dsl-reference.md
```diff
@@ -234,6 +234,7 @@ The Serverless Workflow DSL defines a list of [tasks](#task) that **must be** su
 |:--|:---:|:---:|:---|
 | input | [`input`](#input) | `no` | An object used to customize the task's input and to document its schema, if any. |
 | output | [`output`](#output) | `no` | An object used to customize the task's output and to document its schema, if any. |
+| export | [`export`](#export) | `no` | An object used to customize the content of the workflow context. |
```
cdavernas, line 237: Shouldn't we add the schema, as requested by you and @JBBianchi ? It felt like you both wanted it.

schema/workflow.yaml
```diff
@@ -816,13 +816,20 @@ $defs:
...
     description: Configures the output of a workflow or task.
+  export:
+    type: object
+    properties:
+      schema:
+        $ref: '#/$defs/schema'
+        description: The schema used to describe and validate the workflow context.
+      as:
+        type: string
+        description: A runtime expression, if any, used to set the context with output data.
```
cdavernas, line 831: > A runtime expression, if any, used to set the context with output data.

I would reformulate into something like:

```md
A runtime expression, if any, used to export the output data to the context.
```

cdavernas (changes requested):
Don't you feel `export/to` would be better than `export/as`, knowing it actuals transforms the context data?

@JBBianchi @ricardozanini ?

JBBianchi:
> Don't you feel `export/to` would be better than `export/as`, knowing it actuals transforms the context data?

For me `to` implies a destination whereas `as` is more about the form. For instance, "export to file" or "export as PDF". But I don't really have a strong opinion on the matter, both work.

cdavernas:
Allright! @fjtirado ?

dsl-reference.md
fjtirado: The schema is already there

dsl-reference.md
fjtirado: I think is critical to clarify that the result of the expression is the new value of the context. And therefore, if you want to use the existing context, you are supposed to use `.`

ricardozanini:
@cdavernas @JBBianchi @fjtirado given the context, I also believe `export/to` has a stronger impact here since we are modifying the context. But again, a slight difference that in the end, both might work IMO.

dsl-reference.md
cdavernas: Sorry!

dsl-reference.md
cdavernas: > I think is critical to clarify that the result of the expression is the new value of the context. And therefore, if you want to use the existing context, you are supposed to use .

I disagree. `.` is irrelevant because it is tightly bound to `jq`. You should not even mention the latter, all you are going to achieve is confusing users who use another expression language.

The sentence I proposed clearly indicate that the expression should be run AGAINST the context, and that it is used to update or replace it. I can change it so that the word 'output' clearly appears, but more than that...

fjtirado: Please check the repharse. I mention jq as a possible example.

cdavernas: > If user want to merge the new data into the current context, the context is accessible within the expression scope: `.` if using jq. 

The rephrase still is confusing to me. Of course I can merge or whatever if the context becomes whatever I output.

cdavernas:
2v2 draw for `export/as` vs `output/to`. @matthias-pichler-warrify, care to tie break?

dsl-reference.md
fjtirado: yes, but is important to indicate the writer how to do that using jq. I will remove the JQ reference and leave the example

fjtirado: Done, jq is only used in the example

matthias-pichler-warrify:
> 2v2 draw for `export/as` vs `output/to`. @matthias-pichler-warrify, care to tie break?

I'll quickly catch up on the discussion and then get back to you

dsl-reference.md
fjtirado: Ok, another try

cdavernas: Perfect! Thank you man!

matthias-pichler-warrify: So the input to the `export.to/as` expression is the context (makes sense) but then the output would need to be available as an expression argument via `$output` right?

cdavernas: Yes, exactly, we need to add ouput and it should imo be part of that PR indeed

cdavernas: Can you do that too, @fjtirado ?

dsl-reference.md
```diff
@@ -176,7 +176,7 @@ do:
       - getAvailablePets:
```
JBBianchi, line 176: Idk why, I cannot comment the beginning of the time, but `export` is missing in the TOC.

dsl-reference.md
```diff
@@ -1416,6 +1417,33 @@ from:
...
+Certain task needs to set the context of the workflow to save the task output for later usage. Users set the content of the context through a runtime expression. The result of the expression is the new value of the context. The expression is evaluated against the existing context. 
+
+Optionally, the context might have an associated schema. 
+
+#### Properties
+
+| Property | Type | Required | Description |
+|----------|:----:|:--------:|-------------|
+| schema | [`schema`](#schema) | `no` | The [`schema`](#schema) used to describe and validate context.<br>*Included to handle the non frequent case in which the context has a known format.* |
+| as | `string`<br>`object` | `no` | A runtime expression, if any, used to export the output data to the context. |
```
JBBianchi, line 1431: Link is missing on `runtime expression` -> `[runtime expression](#runtime-expressions)`

matthias-pichler-warrify:
> 2v2 draw for `export/as` vs `export/to`. @matthias-pichler-warrify, care to tie break?

Ok so I have no strong opinions here. I agree with

> For me to implies a destination whereas as is more about the shape/format. For instance, "export to file" or "export as PDF". But I don't really have a strong opinion on the matter, both work.

that `as` sounds better in this case. Especially having used AWS StepFunctions a lot which uses a [`ResultPath`](https://docs.aws.amazon.com/step-functions/latest/dg/input-output-resultpath.html) so I initially thought that `output/to` also gave a destination.

therefore I think

```yaml
export:
  schema: {}
  as: "..."
output:
  schema: {}
  as: "..."
```

is my (only slightly) preferred option

cdavernas:
Consensus reached! We keep things as @fjtirado commited them! Cheers everyone!

ricardozanini (approved):

cdavernas (approved):
Many thanks! ❤️ 
Sorry for late review though, forgot about this one 👅

Issue #849, opened 2024-05-21
Making task names optional


fjtirado:
Currently we write

```
execute: 
   tryWith1:
     try: 
         call:
   tryWith2:
      try: 
        call:
```
execute, try and call are reserved words, tryWith1 is a name the user has to came with

If user is lazy and do not want to invent names (which are useful, as far as I know just for `error.instance` feature), he might want to write this instead

```
execute:
  - try: 
    call:
   - try: 
      name: I want to name that one
      call: 

```

cdavernas:
As discussed in PM, this would go against one of the drives of the refactor, which is having proper references, and which you suggested, by the way.

Making names optional would IMO encourage users to malpractice for the sake of lazyness. Malpractice not only because it will make references more technical (using indexes), but also because of maintenance.

Using, as you suggested, maps like we are doing now is a brilliant idea of yours in my opinion:
- We dont have to enforce some kind of loose name uniqueness control
- We dont have to enforce naming conventions. Authors write what they want, upper, lower, whichever case as long as its a valid property name
- It makes references pretty and allows instant identification of a (possibly faulty) component, even for a non-technical person

PR #895, opened 2024-06-11
Make do an array

### LINKED ISSUE #894: Invalid `do:switch`
**What seems off**:
After #875, and because of, https://github.com/serverlessworkflow/specification/discussions/884#discussioncomment-9662801, #882 was merged, making `do` a single task entry. This breaks the possibility to use `do:switch` as there is nothing to switch to. It forces the user to use `do:execute:sequentially:switch`. 

**What you expected to be**:
We might need to rollback to the original idea of #875, which is to have `do` as `map[string, task]`, but not only for top level.

**Anything else we need to know?**:
As mentioned by @matthias-pichler-warrify in #875, making `do` a `map[string, task]`, the `execute.sequentially` directive doesn't really make sense anymore. We might want to refactor `execute` at the same time (or keep that for another issue).

matthias-pichler-warrify:

- [x] Specification
- [x] Schema
- [x] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [x] CTK
- [ ] Other

**Discussion or Issue link**:

Closes #894

**What this PR does**:

- I refactored the schema to make `do`, `try`, `catch.do`, `do` in `for`, `before` and `after` an array of named tasks
- I refactored the schema to only allowed desired properties
- I added test to make sure the schema rejects invalid definitions
- I added tests to check every example in the dsl-reference doc
- I updated the ctk and examples with the new schema   

**Additional information:**

Now the usage of `do` both in `for` Tasks as well as sequential composite tasks lead to the result that `for` tasks now also matched the schema of the composite sequential task. As discussed earlier this is because additional/unevaluated properties were not restricted. This lead to the error that a `for` task was also recognized as a `do` task and hence matched two schemas in `oneOf` ... I therefore refactored the schema to distinguish between these and no allow additional properties


.ci/validation/src/examples.test.ts
```diff
@@ -15,33 +15,35 @@
  */
 
 import { SWSchemaValidator } from "./index";
-import fs from "fs";
-import { join } from "path";
+import fs from "node:fs";
+import path from "node:path";
```
matthias-pichler-warrify, line 19: Best practice is to reference built-in modules with the `node:` prefix

.ci/validation/src/examples.test.ts
```diff
@@ -15,33 +15,35 @@
...
+    .readdirSync(path.join(__dirname, examplePath), {
+      encoding: SWSchemaValidator.defaultEncoding,
+      recursive: false,
+      withFileTypes: true,
+    })
+    .filter((file) => file.isFile())
+    .filter((file) => file.name.endsWith(".yaml"))
+    .map((file) => file.name);
+
+  test.each(examples)("Example %s", (file) => {
```
matthias-pichler-warrify, line 36: jest offers a `test.each` utility to run parameterized tests. So I refactored to use this as it's more idiomatic 👍

.ci/validation/src/index.ts
```diff
@@ -14,10 +14,10 @@
  * limitations under the License.
  */
 
-import fs from "fs";
-import Ajv from "ajv";
+import fs from "node:fs";
+import Ajv from "ajv/dist/2020";
```
matthias-pichler-warrify, line 18: I updated the ajv import to use draft 2020-12 of json schema to support newer features such as `unevaluatedProperties`

.ci/validation/src/index.ts
```diff
@@ -30,38 +30,46 @@ export module SWSchemaValidator {
...
+  export function loadAsJSON(filename: string): any {
+    return yamlToJSON(fs.readFileSync(filename, defaultEncoding));
+  }
+
+  export function yamlToJSON(yamlStr: string): any {
+    const yamlObj = yaml.load(yamlStr, {
       json: true,
     });
-    return JSON.parse(JSON.stringify(yamlObj, null, 2));
+    return structuredClone(yamlObj);
```
matthias-pichler-warrify, line 58: the new built-in `structuredClone` performs a deep clone of an object. I assumed that `JSON.parse(JSON.stringify())` was done to create a clone

.ci/validation/src/index.ts
```diff
@@ -30,38 +30,46 @@ export module SWSchemaValidator {
...
   }
 
-  export function validateSchema(workflow: JSON) {
+  export function validateSchema(workflow: Record<string, unknown>) {
```
matthias-pichler-warrify, line 61: `JSON` is the type of the global `JSON` namespace that holds `JSON.parse` & `JSON.stringify`. It's type does not represent a JSON object. The correct type here is `Record<string, unknown>`

fjtirado:
@matthias-pichler-warrify 
There are merge conflicts

matthias-pichler-warrify:
> @matthias-pichler-warrify There are merge conflicts

yep, @cdavernas just merged the rename from `output.from` to `output.as` 👍  they are resolved now

ctk/features/composite.feature
```diff
@@ -39,18 +39,19 @@ Feature: Composite Task
       namespace: default
       name: composite-sequential
```
JBBianchi, line 40: The name should rather be `composite-concurrent`.

ctk/features/try.feature
```diff
@@ -15,21 +15,24 @@ Feature: Try Task
...
-      catch:
-        errors:
-          with:
-            type: https://serverlessworkflow.io/dsl/errors/types/communication
-            status: 404
-        as: err
-        do:
-          set:
-            error: ${ $err }
+      - tryGetPet:
```
fjtirado, line 18: After watching the examples updated I think it might be worth to reevaluate the need for enforcing names in task, see https://github.com/serverlessworkflow/specification/issues/849 and https://github.com/serverlessworkflow/specification/discussions/886
Compare 
```
do:
   -tryGetPet:
       try: 
           -getPet:
               call: http
                with:
                  method: get
                  endpoint:
                    uri: https://petstore.swagger.io/v2/pet/getPetByName/{petName}
```
with 
```
do:   
   - try:
      name: getPet
      call: http
      with:
         method: get
                  endpoint:
                    uri: https://petstore.swagger.io/v2/pet/getPetByName/{petName} 
```

dsl-reference.md
```diff
@@ -413,9 +425,9 @@ do:
...
-| execute.concurrently | [`map[string, task][]`](#task) | `no` | The tasks to perform concurrently.<br>*Required if `execute.sequentially` has not been set, otherwise ignored.*<br>*If set, must contains **at least** two [`tasks`](#task).* | 
-| execute.compete | `boolean` | `no` | Indicates whether or not the concurrent [`tasks`](#task) are racing against each other, with a single possible winner, which sets the composite task's output.<br>*Ignored if `execute.sequentially` has been set. Defaults to `false`.*<br>*Must **not** be set if the [`tasks`](#task) are executed sequentially.* |
+| do | [`map[string, task][]`](#task) | `no` | The tasks to perform sequentially.<br>*Required if `fork` has not been set, otherwise ignored.*<br>*If set, must contains **at least** two [`tasks`](#task).* | 
+| fork.branches | [`map[string, task][]`](#task) | `no` | The tasks to perform concurrently.<br>*Required if `do` has not been set, otherwise ignored.*<br>*If set, must contains **at least** two [`tasks`](#task).* |
```
cdavernas, line 429: Should we maybe create two different task types? Like `composite` task and `parallel` task, so that we separate the `do` from the `fork` alltogether.

cdavernas (changes requested):

fjtirado (approved):

dsl-reference.md
matthias-pichler-warrify: Yeah I split them at first, and then merged them back together ... so I am ok with splitting them 👍

ctk/features/try.feature
cdavernas: I understand your point of view. However,  it remains the same than discussed before, I'd rather keep names as non-optional.

JBBianchi: Please, leave discussions in the issues. PR are about points agreed upon beforehand.

dsl-reference.md
```diff
@@ -66,7 +66,7 @@ A [workflow](#workflow) serves as a blueprint outlining the series of [tasks](#t
 | document | [`document`](#document) | `yes` | Documents the defined workflow. |
```
JBBianchi, line 67: Side note, there is a little bug from @fjtirado's PR, if you can fix it at the same time. In the TOC:
`[Export] (#export)`
-> `[Export](#export)`

ctk/features/try.feature
fjtirado: Sure, I did not want to reopen a closed discussion just for the sake of discussing endlessly ;). 
I just realize that when using try with just one task, it is difficult to give different names to the try task and the task within the try. I feel most users will end up typing "<taskName>"  for the task and "try<TaskName" for the try. 
When having multiple task in the try, it becomes event more difficult to find a name for the try task, because user has "<taskName1>" and "<taskName2>" (thats ok and mandatory to distinguish the tasks), but to name the try user will have to summarize the two tasks into one meaningful name. 
Not a big deal anyway. Lets move on.

dsl-reference.md
```diff
@@ -411,9 +423,9 @@ do:
...
-| execute.compete | `boolean` | `no` | Indicates whether or not the concurrent [`tasks`](#task) are racing against each other, with a single possible winner, which sets the composite task's output.<br>*Ignored if `execute.sequentially` has been set. Defaults to `false`.*<br>*Must **not** be set if the [`tasks`](#task) are executed sequentially.* |
+| do | [`map[string, task][]`](#task) | `no` | The tasks to perform sequentially.<br>*Required if `fork` has not been set, otherwise ignored.*<br>*If set, must contains **at least** two [`tasks`](#task).* | 
+| fork.branches | [`map[string, task][]`](#task) | `no` | The tasks to perform concurrently.<br>*Required if `do` has not been set, otherwise ignored.*<br>*If set, must contains **at least** two [`tasks`](#task).* | 
+| fork.compete | `boolean` | `no` | Indicates whether or not the concurrent [`tasks`](#task) are racing against each other, with a single possible winner, which sets the composite task's output.<br>*Ignored if `do` has been set. Defaults to `false`.*<br>*Must **not** be set if the [`tasks`](#task) are executed sequentially.* |
```
JBBianchi, line 430: Maybe it should be two separated sections ?

JBBianchi (approved):

schema/workflow.yaml
```diff
@@ -1,5 +1,5 @@
 $id: https://serverlessworkflow.io/schemas/1.0.0-alpha1/workflow.json
```
ricardozanini, line 1:
```suggestion
$id: https://serverlessworkflow.io/schemas/1.0.0-alpha1/workflow.yaml
```

We might need to change it in the `.ci` validation too.

ricardozanini (commented):
LGTM. Great work, many thanks! Once we split the task section as pointed by @cdavernas I'll do another round.

dsl.md
```diff
@@ -101,7 +101,8 @@ Serverless Workflow DSL allows for defining reusable components that can be refe
 The Serverless Workflow DSL defines several default [task](dsl-reference.md#tasks) types that runtimes **must** implement:
 
 - [Call](dsl-reference.md#call), used to call services and/or functions.
-- [Composite](dsl-reference.md#composite), used to define a minimum of two subtasks to perform.
+- [Do](dsl-reference.md#do), used to define a minimum of two subtasks to perform in sequence.
```
JBBianchi, line 104: Shouldn't we be less restrictive for `do` and allow a single task ?

matthias-pichler-warrify: yeah I removed it. It's not currently enforced by the schema anyways

JBBianchi (approved):

ricardozanini (approved):

cdavernas (approved):
Looks great to me! Many thanks ❤️

PR #896, opened 2024-06-13
[Fix #873] Fixing error in export definition

### LINKED ISSUE #873: Rename output.from and output.to

fjtirado:
I forgot to include export definition in the task.

- [ ] Specification
- [X] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:

**What this PR does**:

**Additional information:**

### DIFF
--- a/schema/workflow.yaml
+++ b/schema/workflow.yaml
@@ -110,6 +110,9 @@ $defs:
       output:
         $ref: '#/$defs/output'
         description: Configure the task's output.
+      export:
+        $ref: '#/$defs/export'
+        description: Export task output to context.
       timeout:
         $ref: '#/$defs/timeout'
         description: The task's timeout configuration, if any.


cdavernas (approved):
LGTM! Thanks!

ricardozanini (approved):

Issue #934, opened 2024-07-24
DSL reference Output example


yzhao244:
In the DSL reference, I noticed the examples of Output component, I don't think I can define "from" and "to" for Output properties, right. If I am not mistaken lol, I can submit PR to correct this example. 

https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#output

```
schema:
  format: json
  document:
    type: object
    properties:
      petId:
        type: string
    required: [ petId ]
from:                         ## remove from 
  petId: '${ .pet.id }'   ## remove this line
to: '.petList += [ . ]'   ## replace to with as
```

- Specification version used:
1.0.0-alpha

cdavernas:
@yzhao244 Great catch! You are totally right!

`output.ftom` should actually be `output.as` and `output.to` should be `export.to`. 

I'm not sure how the examples could be validated though, as CICD should have caught those mistakes 😞

matthias-pichler-warrify:
@cdavernas 
The current schema does specify `export.as`

cdavernas:
@matthias-pichler-warrify yes sorry, my bad 😉

JBBianchi:
> @cdavernas The current schema does specify `export.as`

[Doesn't it ? ](https://github.com/serverlessworkflow/specification/blob/main/schema/workflow.yaml#L937)

yzhao244:
@cdavernas @matthias-pichler-warrify Thanks guys for the comments... submitted a PR for updating this example. https://github.com/serverlessworkflow/specification/pull/936

cdavernas:
Thanks @yzhao244! I will let the pleasure to @ricardozanini to merge it once he could review it. He's gonna be back from PTO in a couple of days 😉

PR #936, opened 2024-07-25
Fix output examples from dsl reference and ctk feature

yzhao244:

- [ ] Specification
- [ ] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [x] CTK
- [x] Other

**Discussion or Issue link**:

#934 

**What this PR does**:

Some fixes to specification version : 1.0.0-alpha

- Fixes output property "as" from two of the examples in the call feature in ctk
- Fixes output property "as" and remove "from" against the example of Output in DSL reference. 

**Additional information:**

Please review in case I misinterpreted any examples of Output in either CTK or DSL reference.

### DIFF
--- a/dsl-reference.md
+++ b/dsl-reference.md
@@ -1515,9 +1515,7 @@ schema:
       petId:
         type: string
     required: [ petId ]
-from:
-  petId: '${ .pet.id }'
-to: '.petList += [ . ]'
+as: '.petList += [ . ]'
 ```
 
 ### Export
--- a/dsl-reference.md
+++ b/dsl-reference.md
@@ -1507,15 +1507,20 @@ When set, runtimes must validate output data against the defined schema, unless 
 #### Examples
 
 ```yaml
-schema:
-  format: json
-  document:
-    type: object
-    properties:
-      petId:
-        type: string
-    required: [ petId ]
-as: '.petList += [ . ]'
+output:
+  schema:
+    format: json
+    document:
+      type: object
+      properties:
+        petId:
+          type: string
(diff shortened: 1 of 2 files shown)


dsl-reference.md
```diff
@@ -1515,9 +1515,7 @@ schema:
       petId:
         type: string
     required: [ petId ]
-from:
-  petId: '${ .pet.id }'
-to: '.petList += [ . ]'
+as: '.petList += [ . ]'
 ```
```
JBBianchi, line 1519: There is a little mistake here I think. `output.from` became `output.as` and `output.to` became `export.as`.

yzhao244: @JBBianchi Thanks for the reply.  Do you mean including both Output and Export properties in this example? Sorry, I feel like I misinterpreted something here lol.. Please let me know  

```
output:
  schema:
    format: json
    document:
      type: object
      properties:
        petId:
          type: string
      required: [ petId ]
  as:                        
    petId: '${ .pet.id }'   
export:
  as: 
   '.petList += [ . ]'  
```

JBBianchi: Either that or just the `output` part without `export` but I don't think it hurts to have both.
Thanks a lot ❤️

yzhao244: Got it... just updated the example with both Output and Export properties.  :)

cdavernas (approved):
LGTM! Thanks @yzhao244 ❤️

ricardozanini (approved):

PR #939, opened 2024-07-26
Use allOf to extend taskBase

matthias-pichler-warrify:

- [ ] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:

**What this PR does**:

From the JSON schema spec (since draft 2019) the following declarations are equivalent:

```yaml
type: object
allOf:
  - $ref: #/$defs/taskBase
properties:
  call: {}
```

and

```yaml
type: object
$ref: #/$defs/taskBase
properties:
  call: {}
```

one can even argue that the second declaration is nicer. Using `allOf` however is more widely supported since it was the only way to do it from draft 4-7.
I however noticed that some tools struggle with the second declaration (specifically https://orval.dev/ and https://redocly.com/).

To make type/client and doc generation easier for implementors I propose to use this more widely supported syntax.

**Additional information:**

### DIFF
--- a/schema/workflow.yaml
+++ b/schema/workflow.yaml
@@ -151,7 +151,8 @@ $defs:
   callTask:
     oneOf:
       - title: CallAsyncAPI
-        $ref: '#/$defs/taskBase'
+        allOf:
+          - $ref: '#/$defs/taskBase'
         type: object
         required: [ call, with ]
         unevaluatedProperties: false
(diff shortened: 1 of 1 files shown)


JBBianchi:
The library used in the TS SDK supports none of the suggested formats but
```yaml
type: object
allOf:
  - $ref: #/$defs/taskBase
  - properties:
      call: {}
```

But I'm not sure it's a good idea to change the current schema to better fit this or that tool. I would argue it's the user/implementor's responsibility to format/preprocess the schema to fit his own tool set; not the other way around.

matthias-pichler-warrify:
> But I'm not sure it's a good idea to change the current schema to better fit this or that tool. I would argue it's the user/implementor's responsibility to format/preprocess the schema to fit his own tool set; not the other way around.

that makes sense 👍  then I'll go ahead and close this PR

PR #926, opened 2024-07-18
Fixes inconsistencies between `listen` and `emit`

JBBianchi:
**Please specify parts of this PR update:**

- [x] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [x] CTK
- [ ] Other

**Discussion or Issue link**:
Closes #917 (and other little minor typo)

**What this PR does**:
- Fixes inconsistencies between `listen` and `emit`
  - Added new `event` section in the specification documentation
  - Fixed the sample of `emit` in the specification documentation
  - Added new `event` definition in the JSON schema and referenced it in `emit.event.with` and `eventFilter.with`
- Fixes a typo in the set feature
- Fixed type of an http call `endpoint` by adding `string`
- Added new (missing) `endpoint` section in the specification documentation
- Escaped * character
- Fixed double `https://https://`
- Fixed links to runtime expressions
- Fixed typo in `implictly`

**Additional information:**
\-

PR #943, opened 2024-08-02
Fixed inconsistencies between `listen` and `emit`

### LINKED ISSUE #917: Semantic inconsistencies between `listen` and `emit` tasks
**What seems off**:

The `listen` task allows for user to configure events to listen to in the following way:

```yaml
listen:
  to:
    one:
      with: {}
```

Whereas the `emit` task does not define the `with` keyword:

```yaml
emit:
  event: {}
```

**What you expected to be**:

The `emit` task to define a `with` keyword, use to define event properties:

```yaml
emit:
  event:
    with: {}
```

This would allow for potential future properties such as `format`, and would make it more fluent, ubiquitous and consistent.

JBBianchi:
Redo of #926

- [x] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [x] CTK
- [ ] Other

**Discussion or Issue link**:
Closes #917 (and other little minor typo)

**What this PR does**:
- Fixes inconsistencies between `listen` and `emit`
  - Added new `event` section in the specification documentation
  - Fixed the sample of `emit` in the specification documentation
  - Added new `event` definition in the JSON schema and referenced it in `emit.event.with` and `eventFilter.with`
- Fixes a typo in the set feature
- Fixed type of an http call `endpoint` by adding `string`
- Added new (missing) `endpoint` section in the specification documentation
- Escaped * character
- Fixed double `https://https://`
- Fixed links to runtime expressions
- Fixed typo in `implictly`

**Additional information:**
\-

### DIFF
--- a/dsl-reference.md
+++ b/dsl-reference.md
@@ -25,8 +25,8 @@
         + [Shell](#shell-process)
         + [Script](#script-process)
         + [Workflow](#workflow-process)
-    - [Switch](#switch)
     - [Set](#set)
+    - [Switch](#switch)
     - [Try](#try)
     - [Wait](#wait)
   + [Flow Directive](#flow-directive)
@@ -42,6 +42,7 @@
     - [Standard Error Types](#standard-error-types)
   + [Event Consumption Strategy](#event-consumption-strategy)
   + [Event Filter](#event-filter)
+  + [Event Properties](#event-properties)
   + [Retry](#retry)
   + [Input](#input)
   + [Output](#output)
@@ -229,9 +230,9 @@ The Serverless Workflow DSL defines a list of [tasks](#task) that **must be** su
 - [Call](#call), used to call services and/or functions.
 - [Do](#do), used to define one or more subtasks to perform in sequence.
 - [Fork](#fork), used to define one or more subtasks to perform concurrently.
-- [Emit](#emit), used to emit [events](#event).
+- [Emit](#emit), used to emit [events](#event-properties).
 - [For](#for), used to iterate over a collection of items, and conditionally perform a task for each of them.
-- [Listen](#listen), used to listen for an [event](#event) or more.
+- [Listen](#listen), used to listen for an [event](#event-properties) or more.
 - [Raise](#raise), used to raise an [error](#error) and potentially fault the [workflow](#workflow).
 - [Run](#run), used to run a [container](#container-process), a [script](#script-process) , a [shell](#shell-process) command or even another [workflow](#workflow-process). 
 - [Switch](#switch), used to dynamically select and execute one of multiple alternative paths based on specified conditions
@@ -368,7 +369,7 @@ The [HTTP Call](#http-call) enables workflows to interact with external services
 | Name | Type | Required | Description|
 |:--|:---:|:---:|:---|
 | method | `string` | `yes` | The HTTP request method. |
-| endpoint | [`endpoint`](#endpoint) | `yes` | An URI or an object that describes the HTTP endpoint to call. |
+| endpoint | `string`\|[`endpoint`](#endpoint) | `yes` | An URI or an object that describes the HTTP endpoint to call. |
 | headers | `map` | `no` | A name/value mapping of the HTTP headers to use, if any. |
 | body | `any` | `no` | The HTTP request body, if any. |
 | output | `string` | `no` | The http call's output format.<br>*Supported values are:*<br>*- `raw`, which output's the base-64 encoded [http response](#http-response) content, if any.*<br>*- `content`, which outputs the content of [http response](#http-response), possibly deserialized.*<br>*- `response`, which outputs the [http response](#http-response).*<br>*Defaults to `content`.* |
@@ -494,7 +495,7 @@ Allows workflows to publish events to event brokers or messaging systems, facili
 
 | Name | Type | Required | Description |
 |:--|:---:|:---:|:---|
-| emit.event | [`event`](#event) | `yes` | Defines the event to emit. |
+| emit.event | [`eventProperties`](#event-properties) | `yes` | Defines the event to emit. |
 
 ##### Examples
 
@@ -508,15 +509,16 @@ do:
   - emitEvent:
       emit:
         event:
-          source: https://petstore.com
-          type: com.petstore.order.placed.v1
-          data:
-            client:
-              firstName: Cruella
-              lastName: de Vil
-            items:
-              - breed: dalmatian
-                quantity: 101
+          with:
+            source: https://petstore.com
+            type: com.petstore.order.placed.v1
+            data:
+              client:
+                firstName: Cruella
+                lastName: de Vil
+              items:
+                - breed: dalmatian
+                  quantity: 101
 ```
 
 #### For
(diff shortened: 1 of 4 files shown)


fjtirado (commented):
@JBBianchi I was wondering if it wont be better to have a schema with the data that is shared by emit and listen (which I feel should contain type, source, subject, dataContentType and time?)  and another one that inherits it (and adds id, data and dataschema?) and is used for emit

JBBianchi:
> @JBBianchi I was wondering if it wont be better to have a schema with the data that is shared by emit and listen (which I feel should contain type, source, subject, dataContentType and time?) which will be used for listen filter and another one that inherits it (and adds id, data and dataschema?) and is used for emit Then, you wont have to explain that id is required for emission only (you still have for source and type, which are shared by both), but at least a user wont be able to put id or data in a filter (I have doubts about if dataschema and time can be used for filtering or not, precisely to avoid this kind of doubt I feel we need two different schemas, that can share common properties)

Unless I'm mistaken, the current schema already shares the common properties and validates the ones that are required at their level. The explanation is necessary in the documentation because its describing the "shared" object properties.

As a side note, I'm still waiting for feedbacks ok [@ricardozanini](https://github.com/serverlessworkflow/specification/pull/926#discussion_r1700836924) 's comment. I think it's a broader scope than just this PR though and would affect other object that could be built using runtime expressions. Should we open a dedicated issue ?

fjtirado:
@JBBianchi 
But id and data should not be part of the event filter on listen isnt it?

cdavernas:
@fjtirado It can, but is not required

fjtirado:
@cdavernas Ok, that was my doubt, so we can filter by any cloud event field.

fjtirado:
Which remind me that maybe the right name for the schema, rather that eventproperties should be cloudEvent?

fjtirado (approved):

cdavernas:
@fjtirado that was how JB named it first, but is IMHO opinion confusing. It's the properties/filter of the CloudEvent to produce/listen, not the event itself

dsl-reference.md
```diff
@@ -368,7 +369,7 @@ The [HTTP Call](#http-call) enables workflows to interact with external services
 | Name | Type | Required | Description|
 |:--|:---:|:---:|:---|
 | method | `string` | `yes` | The HTTP request method. |
-| endpoint | [`endpoint`](#endpoint) | `yes` | An URI or an object that describes the HTTP endpoint to call. |
+| endpoint | `string`\|[`endpoint`](#endpoint) | `yes` | An URI or an object that describes the HTTP endpoint to call. |
```
cdavernas, line 372: There is an `/` escape character here that afaik is useless and should therefore be removed

dsl-reference.md
```diff
@@ -1389,6 +1391,28 @@ Represents the configuration of an event consumption strategy.
 | any | [`eventFilter[]`](#event-filter) | `no` | Configures the workflow to wait for any of the defined events before resuming execution.<br>*Required if `all` and `one` have not been set.* |
 | one | [`eventFilter`](#event-filter) | `no` | Configures the workflow to wait for the defined event before resuming execution.<br>*Required if `all` and `any` have not been set.* |
 
+### Event Properties
```
cdavernas, line 1394: While the description is good, afaik it applies to an event, instead of describing the properties object, which might be confusing, no?

dsl.md
```diff
@@ -81,7 +81,7 @@ Workflows in the Serverless Workflow DSL can exist in several phases, each indic
 
 Additionally, the flow of execution within a workflow can be controlled using [directives*](dsl-reference.md#flow-directive), which provide instructions to the workflow engine on how to manage and handle specific aspects of workflow execution.
 
-**To learn more about flow directives and how they can be utilized to control the execution and behavior of workflows, please refer to [Flow Directives](dsl-reference.md#flow-directive).*
+\**To learn more about flow directives and how they can be utilized to control the execution and behavior of workflows, please refer to [Flow Directives](dsl-reference.md#flow-directive).*
```
cdavernas, line 84: AFAIK, starts with a useless escape char

cdavernas (changes requested):
Looks great to me, aside from 3 little comments! Thanks for your work ❤️

dsl-reference.md
JBBianchi: it's useful or otherwise it "brakes" the table.

| Name | Type | Required | Description|
|:--|:---:|:---:|:---|
| endpoint | `string`\|[`endpoint`](#endpoint) | `yes` | An URI or an object that describes the HTTP endpoint to call. |

vs

| Name | Type | Required | Description|
|:--|:---:|:---:|:---|
| endpoint | `string`|[`endpoint`](#endpoint) | `yes` | An URI or an object that describes the HTTP endpoint to call. |

dsl.md
JBBianchi: \* opens formatting for italic, therefore it's better to escape it if you just want to print a star.

dsl-reference.md
cdavernas: Brakes because otherwise the table goes too fast? 😂
Sorry, I couldn't resist, I'm out.

dsl-reference.md
JBBianchi: We are describing the properties of an event, therefore giving the context of what the event is and then having the table with the properties seemed like a good approach. What alternative do you suggest ?

dsl-reference.md
```diff
@@ -1397,7 +1421,7 @@ An event filter is a mechanism used to selectively process or handle events base
 
 | Property | Type | Required | Description |
 |----------|:----:|:--------:|-------------|
-| with | `object` | `yes` | A name/value mapping of the attributes filtered events must define. Supports both regular expressions and runtime expressions.  |
+| with | [`eventProperties`](#event-properties) | `yes` | A name/value mapping of the attributes filtered events must define. Supports both regular expressions and runtime expressions.  |
```
ricardozanini, line 1424: I thought I was getting crazy since I couldn't see my previous comments from the other PR. Now I see you reopened it. 

Regarding this property, since we are now defining that it can be regexp, runtime expressions, or `eventProperties`, this should reflect in the schema.

So this `with` should be an `OneOf` there amongst these types.

JBBianchi: Sorry for the mix-up. When I rebased the branch, GitHub closed my previous PR. I opened a new one and realized afterward that I could have simply reopened the previous one after I readded my commits.

I might be mistaken, but I was under the impression that the statement "Supports both regular expressions and runtime expressions" applies to the properties of `with`, not the object itself. That said, there might be room for discussion about how and where we can use a runtime expression to construct an entire object, but that might be a broader topic than what is addressed in this PR.

@cdavernas @fjtirado @matthias-pichler-warrify, what are your thoughts on this?

matthias-pichler-warrify: I would understand this to mean that the individual properties under `with` (such as `id`, `source`, ...) should each support literals, regex and runtime expressions. These are all strings so we would just produce a union of type `string | string | string` which doesn't seem necessary. 
I would keep the `with` object an object. 

regarding regular expressions: How would an implementor determine if a given string is a literal or a regular expression though?

For example if the filter was specified as

```yaml
with:
  type: com.example  
```

if the string was interpreted as a literal only `com.example` would match. As a regex `comaexample` and `com.example.foo` would also match! Are we planning on adding a regex syntax (JSON schema support the ECMA 262 syntax with `format: regex`) or object? Like

```yaml
with:
  type:
    re: ^com\.example

# or

with:
  type: /com\.example/
```

Maybe we should just stick to runtime expressions and use the `test` function: https://jqlang.github.io/jq/manual/#regular-expressions ?

```yaml
with:
  source: ${ test(., "^com\\.example$") }
```

JBBianchi: That's a good point. It's a very specific pain point considering regular expressions are only supported in `listen` tasks `event filters`.

My understanding was that the value was either a string or a runtime expression. If it's a string, then it's regarded as a regular expression (regexp). But it's not very clear, can introduce undesired matches when not properly escaped, and probably have an bad impact on performance (I would assume string equality to be more performant than regexp, in turn more performant than jq/js expressions.)

For the performance reason I just mentioned (without any proof though, it's only a feeling ^^'), I would rule out relying only on jq/js for processing regexp. Therefore, we need to either consider any string as a regexp or find a way to discriminate the regexp. I must admit I'm not convinced by neither adding a property nor using a specific format but I don't have a better option to suggest. Therefore, I would rather go for the specific format rather than the extra property.

matthias-pichler-warrify: > (I would assume string equality to be more performant than regexp, in turn more performant than jq/js expressions.)

My intuition is that the performance difference between string equality and regex matching is negligible compared to `jq` expressions which are much slower.

> Therefore, I would rather go for the specific format rather than the extra property.

If we support regex I would also say that we should use a format such as `/<regex>/`. We would however not be able to rely on the `format: regex` from json schema since this does not have any delimiters

PR #944, opened 2024-08-02
Adding missing titles for better pojo generation

fjtirado:

- [ ] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:

**What this PR does**:

**Additional information:**


schema/workflow.yaml
```diff
@@ -664,7 +664,8 @@ $defs:
         basic:
           type: object
           oneOf:
-            - properties:
+            - title: BasicAuthInfo
```
cdavernas, line 667: Wouldn't it be better to avoid abbreviations, especially for a descriptive property?

schema/workflow.yaml
```diff
@@ -680,7 +681,8 @@ $defs:
         bearer:
           type: object
           oneOf:
-            - properties:
+            - title: BearerAuthInfo
```
cdavernas, line 684: Same comment than above

schema/workflow.yaml
```diff
@@ -693,7 +695,8 @@ $defs:
         oauth2:
           type: object
           oneOf:
-            - properties:
+            - title: OAuth2Info
```
cdavernas, line 698: To be more consistent, I'd name ot something like `OAuth2AuthenticationInfo`, even though I'm aware that auth would therefore be repeated. Reason is that `OAUTH2` is the scheme's name.

cdavernas (changes requested):
LGTM, aside from a couple if minor comments! Thanks for your work man! ❤️

ricardozanini (approved):

PR #949, opened 2024-08-04
Support to specify auth parameters as secrets

matthias-pichler-warrify:

- [ ] Specification
- [x] Schema
- [x] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:

**What this PR does**:

I updated the schema so that auth parameters such as client_id, password, token etc can be specified via secrets with `use:`

**Additional information:**

The way I understand it is that in principle any literal value can be specified as a secret. Even random things like `grant` or even things like the duration in `wait` ... what should we do about this? 

Should any value that currently uses a literal type (such as `string`, `boolean`, `number`) be substitued for a custom type (e.g. `StringValue`) that is a union of `<literal> | #/$defs/secretReference | #/$defs/runtimeExpression`?

What about complex types such as `oauth2.client`, `basic` or even arbitrary like `timeout`?

### DIFF
--- a/examples/secret-basic-auth.yaml
+++ b/examples/secret-basic-auth.yaml
@@ -1,0 +1,22 @@
+document:
+  dsl: 1.0.0-alpha1
+  namespace: examples
+  name: secret-basic-auth
+  version: 1.0.0-alpha1
+use:
+  secrets:
+    - secretUser
+    - secretPassword
+do:
+  - getPet:
+      call: http
+      with:
+        method: get
+        endpoint:
+          uri: https://petstore.swagger.io/v2/pet/{petId}
+          authentication:
+            basic:
+              username:
+                use: secretUser
+              password:
+                use: secretPassword
--- a/examples/secret-bearer-auth.yaml
+++ b/examples/secret-bearer-auth.yaml
@@ -1,0 +1,19 @@
+document:
+  dsl: 1.0.0-alpha1
+  namespace: examples
+  name: secret-bearer-auth
+  version: 1.0.0-alpha1
+use:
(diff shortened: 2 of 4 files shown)


cdavernas:
I wouldnt do that, as it's extremely challenging to parse and replace. Using secrets should me much more straightforward and should instead allow only direct replacement at top level, or using the `$secrets` expression argument.

matthias-pichler-warrify:
> I wouldnt do that, as it's extremely challenging to parse and replace. Using secrets should me much more straightforward and should instead allow only direct replacement at top level, or using the $secrets expression argument.

Yeah that makes sense. The resulting schema isn't pretty and the `$secrets` argument already gives a good way to access secrets in places where strings are expected.

> In addition, this forces upon all authentication properties (and there are a lot) the ugly and feared union types, which are both confusing and tedious to implement. Also, this would open the Pandora's box, as there would be no reason not to use that mechanism everywhere else.

yep that's what I alluded to in my "Additional Information". I thought that was intended but actually I think it would be super confusing as well 👍 

> As a side note, please refrain from opening a PR that closes an issue/feature request that has not been discussed or approved beforehand 😉

My bad 😅  thought that this was intended by the current spec

cdavernas:
> My bad 😅 thought that this was intended by the current spec

@matthias-pichler-warrify No worries at all mate, come on! That's like the only change out of the ones you proposed and pushed I would personally guard against.

On that note, many thanks for your crazy streak of (awesome) PRs ❤️

Issue #910, opened 2024-06-27
Specifying Query Parameters for call: HTTP 


matthias-pichler-warrify:
**What would you like to be added**:

Would it make sense to add a structured way to add query parameters to uris when using `call: http`?

```yaml
document: {}
do:
  - listPets:
      call: http
      with:
        method: get
        endpoint: 
          uri: https://petstore.swagger.io/v2/pets
        search:
          page[size]: ${ .pageSize }
          str: "Some string with characters & stuff that would need to be uri percent encoded! 👍 "
```

We could use the name "search", "query", "searchParameters", "queryparameters" or "parameters" 

**Why is this needed**:

right now it is somewhat painful to specify these things using a runtime expression because we'd need to use the `@uri` format. And `@uri` also only encodes interpolated strings so the keys would have to manually encoded

```yaml
document: {}
do:
  - listPets:
      call: http
      with:
        method: get
        endpoint: 
          uri: ${ @uri "https://petstore.swagger.io/v2/pets?page%5Bsize%5D=\(.pageSize)&str=\(\"Some string with characters & stuff that would need to be uri percent encoded! 👍 \")" }
```

cdavernas:
@matthias-pichler-warrify you can also use top level args with `{ }`, as demonstrated by https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#http-call.

matthias-pichler-warrify:
I only saw `arguments` under `call: openapi` but didn't find anything for `call: http`

matthias-pichler-warrify:
@cdavernas 
> As for uri encoding, I always assumed that was the responsibility of runtimes.

In principle I am with you on this one but I am thinking of a case where the query parameter keys would also need encoding (e.g. `page[size]`). In this case the author would have to properly encode the square brackets when they include them in the `uri`.

I am also wondering how to elegantly handle array valued query params as there are different representations:

- repetition
- empty square brackets
- comma separated 

Which could be handled with a appropriate field

cdavernas:
@matthias-pichler-warrify fair enough, you make a very good point!

What about using `query` or `queryParameters`? Amongst your proposals, those are the terms that best represent imho the parameters we are trying to describe.

On another note, I believe it would be awesome if the PR addressing the present issue would include documenting the ease of use feature I described in my last comment.

Wdyt?

matthias-pichler-warrify:
[...]

PR #955, opened 2024-08-04
Document limited uri-template support

matthias-pichler-warrify:

- [x] Specification
- [ ] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:

As discussed in #910

**What this PR does**:

Documents the frequently used by undocumented string substitution feature

**Additional information:**


cdavernas:
I'm not sure string substitution should be allowed everywhere, but instead should be restricted to URIs, else it would ask too much checks at runtime, don't you think?

matthias-pichler-warrify:
> I'm not sure string substitution should be allowed everywhere, but instead should be restricted to URIs, else it would ask too much checks at runtime, don't you think?

actually I think it should be available everywhere where a runtime expression could be used (so pretty much everywhere 😅 ).
Furthermore most uris are currently defined as [`uri-template`](https://datatracker.ietf.org/doc/html/rfc6570) in the schema ... are wqe planning on supporting the full uri-template syntax or only the simple substitution?

cdavernas:
> actually I think it should be available everywhere where a runtime expression could be used

That's why I'd personally refrain from doing so, as you can easily do that using runtime expressions, which are much easier to spot for a runtime. 

This would imho become extremely cumbersome to have to check for both `${}` blocks and `{}` blocks, which could in addition conflict. 

Furthermore, it would mean authors would not be able to use those characters without forcing runtimes to evaluate, even when not necessary, therefore needing some escaping feature.

Anyhow, thats just me, let's see what @JBBianchi, @fjtirado and @ricardozanini think about it 😉

cdavernas:
> Furthermore most uris are currently defined as [uri-template](https://datatracker.ietf.org/doc/html/rfc6570) in the schema ... are wqe planning on supporting the full uri-template syntax or only the simple substitution?

IMHO only simple substitution

matthias-pichler-warrify:
> That's why I'd personally refrain from doing so, as you can easily do that using runtime expressions, which are much easier to spot for a runtime.
>
> This would imho become extremely cumbersome to have to check for both ${} blocks and {} blocks, which could in addition conflict.
>
> Furthermore, it would mean authors would not be able to use those characters without forcing runtimes to evaluate, even when not necessary, therefore needing some escaping feature.

No strong opinion from my side. I am fine with only allowing this in URIs. In this case though I would rephrase the feature away from "string substitution" to "limited support for uri-template syntax". This would then allow us to expand to support additional uri-template features in the future (or now?).

cdavernas (changes requested):
LGTM, aside from two little typos ;) Many thanks!

fjtirado:
Although already changed, I agree with @cdavernas. String substitution should apply only to uri templates

fjtirado (approved):

cdavernas (approved):
LGTM! Thanks man!

ricardozanini (approved):

PR #956, opened 2024-08-04
Allow query parameters in call http

### LINKED ISSUE #910: Specifying Query Parameters for call: HTTP

matthias-pichler-warrify:

- [x] Specification
- [x] Schema
- [x] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:

Closes #910

**What this PR does**:

It allows authors to specify query paramters in HTTP requests.

**Additional information:**


cdavernas (approved):
LGTM! Cheers ❤️

ricardozanini (approved):

PR #953, opened 2024-08-04
Specify task and workflow arguments

matthias-pichler-warrify:
Signed-off-by: Matthias Pichler <m.pichler@warrify.com>

**Please specify parts of this PR update:**

- [x] Specification
- [ ] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:

Closes #951

**What this PR does**:

It clarifies the contents of the `$workflow` and `$task` arguments

**Additional information:**

Issue #951, opened 2024-08-04
Document content of $workflow and $task arguments


matthias-pichler-warrify:
**What would you like to be added**:

The `$workflow` and `$task` arguments are mentioned in the DSL but their contents are never specified. We should clarify this:

## `$workflow`

- `$workflow.definition` ... the workflow definition as a parsed object
- `$workflow.input` ... the input of the workflow _BEFORE_ the `input.from` expression was applied. For the result of the `input.from` expression use the `$input` argument
- `$workflow.startedAt`... an object with the properties `iso8601`, `epochMillis` and `epochNanos` to represent the execution start time
- `$workflow.id` ... the unique id of the execution

## `$task`

- `$task.definition` ... the task definition (object under the name) as a parsed object
- `$task.input` ... the input of the workflow _BEFORE_ the `input.from` expression was applied. For the result of the `input.from` expression use the context `.` of the expression
- `$task.startedAt`... an object with the properties `iso8601`, `epochMillis` and `epochNanos` to represent the task start time
- `$task.name` ... the unique name of the task

**Why is this needed**:

**Open Questions**:

Should the `.input` properties be the result of the `input.from` expression or the raw input to avoid duplication?

PR #958, opened 2024-08-04
Document Data Flow and expression context more clearly

### LINKED ISSUE #921: Specify context of task.if expression

I think we should specify what the runtime context (i.e. value of `.` in jq) is of the [`if` expression of a task](https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#properties-1). I understand it to be the task input. 

**Why is this needed**:

This removes ambiguity for both implementors and authors.

matthias-pichler-warrify:

- [x] Specification
- [ ] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:

Closes #921

**What this PR does**:

I extended the "Data Flow" section of the DSL with some clarifying statements and also added a diagram showing the data flow.

**Additional information:**

### DIFF
--- a/dsl.md
+++ b/dsl.md
@@ -154,33 +154,33 @@ Before the workflow starts, the input data provided to the workflow can be trans
 *Example: If the workflow receives a JSON object as input, a transformation can be applied to remove unnecessary fields and retain only those that are required for the workflow's execution.*
 
 2. **Transform First Task Input**
-The input data for the first task can be transformed to match the specific requirements of that task. This ensures that the first task receives only the necessary data it needs to perform its operations. This can be done using the task's `input.from` expression. It evaluates on the transformed workflow input and defaults to the identity expression which leaves the input unchanged. The result of this expression will be set as the `$input` runtime expression argument and be passed to the task. Any runtime expressions used within the task definition will be evaluated using this transformed input.
+The input data for the first task can be transformed to match the specific requirements of that task. This ensures that the first task receives only the data required to perform its operations. This can be done using the task's `input.from` expression. It evaluates the transformed workflow input and defaults to the identity expression, which leaves the input unchanged. The result of this expression will be set as the `$input` runtime expression argument and be passed to the task. This transformed input will be evaluated against any runtime expressions used within the task definition.
 
 *Example: If the first task is a function call that only needs a subset of the workflow input, a transformation can be applied to provide only those fields needed for the function to execute.*
 
 3. **Transform First Task Output**
-After the first task completes, its output can be transformed before passing it to the next task or storing it in the workflow context. Transformations are applied using the `output.as` runtime expression. It evaluates on the raw task output and defaults to the identity expression which leaves the output unchanged. Its result will be the input to the next task. To update the context one uses the `export.as` runtime expression. It evaluates on the raw output and defaults to the expression that returns the existing context. The result of this runtime expression replaces the workflows current context and the content of the `$context`runtime expression argument. This helps in managing the data flow and keeping the context clean by removing any unnecessary data produced by the task.
+After completing the first task, its output can be transformed before passing it to the next task or storing it in the workflow context. Transformations are applied using the `output.as` runtime expression. It evaluates the raw task output and defaults to the identity expression, which leaves the output unchanged. Its result will be input for the next task. To update the context, one uses the `export.as` runtime expression. It evaluates the raw output and defaults to the expression that returns the existing context. The result of this runtime expression replaces the workflow's current context and the content of the `$context` runtime expression argument. This helps manage the data flow and keep the context clean by removing any unnecessary data produced by the task.
 
 *Example: If the first task returns a large dataset, a transformation can be applied to retain only the relevant results needed for subsequent tasks.*
 
 4. **Transform Last Task Input**
-Before the last task in the workflow executes, its input data can be transformed to ensure it receives only the necessary information. This can be done using the task's `input.from` expression. It evaluates on the transformed workflow input and defaults to the identity expression which leaves the input unchanged. The result of this expression will be set as the `$input` runtime expression argument and be passed to the task. Any runtime expressions used within the task definition will be evaluated using this transformed input. This step is crucial for ensuring that the final task has all the required data to complete the workflow successfully.
+Before the last task in the workflow executes, its input data can be transformed to ensure it receives only the necessary information. This can be done using the task's `input.from` expression. It evaluates the transformed workflow input and defaults to the identity expression, which leaves the input unchanged. The result of this expression will be set as the `$input` runtime expression argument and be passed to the task. This transformed input will be evaluated against any runtime expressions used within the task definition. This step is crucial for ensuring the final task has all the required data to complete the workflow successfully.
 
 *Example: If the last task involves generating a report, the input transformation can ensure that only the data required for the report generation is passed to the task.*
 
 5. **Transform Last Task Output**
-After the last task completes, its output can be transformed before it is considered as the workflow output. Transformations are applied using the `output.as` runtime expression. It evaluates on the raw task output and defaults to the identity expression which leaves the output unchanged. Its result will be passes to the workflow `output.as` runtime expression. This ensures that the workflow produces a clean and relevant output, free from any extraneous data that might have been generated during the task execution.
+After the last task completes, its output can be transformed before it is considered the workflow output. Transformations are applied using the `output.as` runtime expression. It evaluates the raw task output and defaults to the identity expression, which leaves the output unchanged. Its result will be passed to the workflow `output.as` runtime expression. This ensures that the workflow produces a clean and relevant output, free from any extraneous data that might have been generated during the task execution.
 
 *Example: If the last task outputs various statistics, a transformation can be applied to retain only the key metrics that are relevant to the stakeholders.*
 
 6. **Transform Workflow Output**
-Finally, the overall workflow output can be transformed before it is returned to the caller or stored. Transformations are applied using the `output.as` runtime expression. It evaluates on the last task's output and defaults to the identity expression which leaves the output unchanged. This step ensures that the final output of the workflow is concise and relevant, containing only the necessary information that needs to be communicated or recorded.
+Finally, the overall workflow output can be transformed before it is returned to the caller or stored. Transformations are applied using the `output.as` runtime expression. It evaluates the last task's output and defaults to the identity expression, which leaves the output unchanged. This step ensures that the final output of the workflow is concise and relevant, containing only the necessary information that needs to be communicated or recorded.
 
 *Example: If the workflow's final output is a summary report, a transformation can ensure that the report contains only the most important summaries and conclusions, excluding any intermediate data.*
 
-By applying transformations at these strategic points, Serverless Workflow DSL ensures that data flows through the workflow in a controlled and efficient manner, maintaining clarity and relevance at each stage of execution. This approach helps in managing complex workflows and ensures that each task operates with the precise data it requires, leading to more predictable and reliable workflow outcomes.
+By applying transformations at these strategic points, Serverless Workflow DSL ensures that data flows through the workflow in a controlled and efficient manner, maintaining clarity and relevance at each execution stage. This approach helps manage complex workflows and ensures that each task operates with the precise data required, leading to more predictable and reliable workflow outcomes.
 
-Visually this can be represented as follows:
+Visually, this can be represented as follows:
 
 ```mermaid
 flowchart TD
@@ -306,7 +306,7 @@ The following table shows which arguments are available for each runtime express
 | Runtime Expression | Evaluated on | Produces | `$context` | `$input` | `$output` | `$secrets` | `$task` | `$workflow` |
 |:-------------------|:---------:|:---------:|:---------:|:---------:|:-------:|:---------:|:-------:|:----------:|
 | Workflow `input.from` | Raw workflow input | Transformed workflow input | | | | ✔ | | ✔ |
-| Task `input.from` | Raw task input (i.e. transformed workflow input for first task, transformed output from previous task otherwise) | Transformed task input | ✔ | | | ✔ | ✔ | ✔ |
+| Task `input.from` | Raw task input (i.e. transformed workflow input for the first task, transformed output from previous task otherwise) | Transformed task input | ✔ | | | ✔ | ✔ | ✔ |
 | Task `if` | Transformed task input | | ✔ | ✔ | | ✔ | ✔ | ✔ |
 | Task definition | Transformed task input | | ✔ | ✔ | | ✔ | ✔ | ✔ |
 | Task `output.as` | Raw task output | Transformed task output | ✔ | ✔ | | ✔ | ✔ | ✔ |
--- a/dsl.md
+++ b/dsl.md
@@ -149,36 +149,92 @@ In Serverless Workflow DSL, data flow management is crucial to ensure that the r
 Here's how data flows through a workflow based on various filtering stages:
 
 1. **Filter Workflow Input**
-Before the workflow starts, the input data provided to the workflow can be filtered to ensure only relevant data is passed into the workflow context. This step allows the workflow to start with a clean and focused dataset, reducing potential overhead and complexity in subsequent tasks.
+Before the workflow starts, the input data provided to the workflow can be filtered to ensure only relevant data is passed into the workflow context. This can be done using the top level `input.from` expression. It evaluates on the raw workflow input and defaults to the identity expression which leaves the input unchanged. This step allows the workflow to start with a clean and focused dataset, reducing potential overhead and complexity in subsequent tasks. The result of this expression will set as the initial value for the `$context` runtime expression argument and be passed to the first task.
 
 *Example: If the workflow receives a JSON object as input, a filter can be applied to remove unnecessary fields and retain only those that are required for the workflow's execution.*
 
 2. **Filter First Task Input**
-The input data for the first task can be filtered to match the specific requirements of that task. This ensures that the first task receives only the necessary data it needs to perform its operations.
+The input data for the first task can be filtered to match the specific requirements of that task. This ensures that the first task receives only the necessary data it needs to perform its operations. This can be done using the task's `input.from` expression. It evaluates on the filtered workflow input and defaults to the identity expression which leaves the input unchanged. The result of this expression will be set as the `$input` runtime expression argument and be passed to the task. Any runtime expressions used within the task definition will be evaluated using this filtered input.
 
 *Example: If the first task is a function call that only needs a subset of the workflow input, a filter can be applied to provide only those fields needed for the function to execute.*
 
 3. **Filter First Task Output**
-After the first task completes, its output can be filtered before passing it to the next task or storing it in the workflow context. This helps in managing the data flow and keeping the context clean by removing any unnecessary data produced by the task.
+After the first task completes, its output can be filtered before passing it to the next task or storing it in the workflow context. Filtering is performed using the `output.as` runtime expression. It evaluates on the raw task output and defaults to the identity expression which leaves the output unchanged. Its result will be the input to the next task. To update the context one uses the `export.as` runtime expression. It evaluates on the raw output and defaults to the expression that returns the existing context. The result of this runtime expression replaces the workflows current context and the content of the `$context`runtime expression argument. This helps in managing the data flow and keeping the context clean by removing any unnecessary data produced by the task.
 
 *Example: If the first task returns a large dataset, a filter can be applied to retain only the relevant results needed for subsequent tasks.*
 
 4. **Filter Last Task Input**
-Before the last task in the workflow executes, its input data can be filtered to ensure it receives only the necessary information. This step is crucial for ensuring that the final task has all the required data to complete the workflow successfully.
+Before the last task in the workflow executes, its input data can be filtered to ensure it receives only the necessary information. This can be done using the task's `input.from` expression. It evaluates on the filtered workflow input and defaults to the identity expression which leaves the input unchanged. The result of this expression will be set as the `$input` runtime expression argument and be passed to the task. Any runtime expressions used within the task definition will be evaluated using this filtered input. This step is crucial for ensuring that the final task has all the required data to complete the workflow successfully.
 
 *Example: If the last task involves generating a report, the input filter can ensure that only the data required for the report generation is passed to the task.*
 
 5. **Filter Last Task Output**
-After the last task completes, its output can be filtered before it is considered as the workflow output. This ensures that the workflow produces a clean and relevant output, free from any extraneous data that might have been generated during the task execution.
+After the last task completes, its output can be filtered before it is considered as the workflow output. Filtering is performed using the `output.as` runtime expression. It evaluates on the raw task output and defaults to the identity expression which leaves the output unchanged. Its result will be passes to the workflow `output.as` runtime expression. This ensures that the workflow produces a clean and relevant output, free from any extraneous data that might have been generated during the task execution.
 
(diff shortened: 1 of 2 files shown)


cdavernas:
Not sure that it's needed, as it's always the filtered input that is used, everywhere but in the `input.from`, which is the only exception.

matthias-pichler-warrify:
> Not sure that it's needed, as it's always the filtered input that is used, everywhere but in the `input.from`, which is the only exception.

@cdavernas should I then close #921 without any additional clarifications? 

I have to admit though that I always have to think a bit what the context of a given expression is so I imagine clarifications could help. Furthermore it is not inherently clear (from just the names) that `if` runs after `input.from` ... maybe this should be added in [the data flow section](https://github.com/serverlessworkflow/specification/blob/main/dsl.md#data-flow) even though it's not data flow related 🤷

cdavernas:
No no, that's just my opinion, let's wait for others to kick in 😉 

Anyways, more info cannot hurt, especially if you feel that's not especially intuitive or lacks clarification 😉

cdavernas:
> maybe this should be added in [the data flow section](https://github.com/serverlessworkflow/specification/blob/main/dsl.md#data-flow) even though it's not data flow related 🤷

Maybe, yes! Along with some explicit graphical/schema representation of what comes/happens when, as @zolero pointed out to me in PMs!

matthias-pichler-warrify:
@cdavernas I removed the extra statement from the `task.if` description and added a general overview for runtime expressions and data flow 👍

JBBianchi:
More explanation cannot hurt, especially on a topic that seems confusing (see https://github.com/serverlessworkflow/specification/discussions/890 for instance). 

If I can add my two cents, I would get rid of the term "filter*" in favor of "transform*". Filter conveys the idea of removing part of the dataset where transform is broader and just implies mutations.

matthias-pichler-warrify:
> If I can add my two cents, I would get rid of the term "filter*" in favor of "transform*". Filter conveys the idea of removing part of the dataset where transform is broader and just implies mutations.

I like it 👍 I also felt "Filtered" wasn't the right term

cdavernas:
@matthias-pichler-warrify I liked the term `filter`, but @JBBianchi made a valid point. I'm totally OK with replacing it with`transform` or whatever you guys feel fits better!

cdavernas (approved):
Looks great to me, many thanks!

matthias-pichler-warrify:
`export.as` sometimes uses the undocumented `$output` variable (e.g. https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#examples-31) 

This argument exists because the `export.as` expressions is evaluated on the existing context. Should I:

a) add the `$output` argument to the docs and evaluate `export.as` on the existing context
b) change `export.as` to be evaluated on the transformed output and remove `$output` from the examples

fjtirado:
> `export.as` sometimes uses the undocumented `$output` variable (e.g. https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#examples-31)
> 
> This argument exists because the `export.as` expressions is evaluated on the existing context and thus there is no other way to access the output. Should I:
> 
> a) add the `$output` argument to the docs and evaluate `export.as` on the existing context b) change `export.as` to be evaluated on the transformed output and remove `$output` from the examples

I prefer option b). @cdavernas prefers option a). We can decide by throwing a coin, ping pong match or voting ;)

cdavernas:
As @fjtirado said, I have a priori a slight preference for `a` because it is a tad more agile imho. I however do not have a strong or irreversible opinion. As long as the user can use both raw and modified output, it's fine by me.

matthias-pichler-warrify:
> As long as the user can use both raw and modified output, it's fine by me.

That is a good point, because as I understand it, neither option allows this. Should I update #953 to include a `$task.output` property that holds the raw output?

fjtirado:
The output is, by default, the input of the next task, isnt it?
Then yes, we should be able to dump the original task output (the result of the rest invocation, for example) to the context. 
Whatever option we chose we need careful documentation and example, since this is as subttle as a wargame manual ;)

JBBianchi:
>a) add the $output argument to the docs and evaluate export.as on the existing context
>b) change export.as to be evaluated on the transformed output and remove $output from the examples

My initial preference was option **a** because it makes sense for `.` to be the context, as it is the affected object. However, the argument for removing the extra `$output` argument, which only exists in `export.as`, is a valid point as well.

> As long as the user can use both raw and modified output, it's fine by me.

From my understanding of the specification, `$output` refers to the raw output, and there is no way to access the "transformed" output because `output` and `export` are processed simultaneously. However, it is reasonable to consider that `export` might be processed after `output`.

If we agree with @matthias-pichler-warrify's proposal to make the raw output accessible to `$task`, we could have:

- `.`: the transformed output
- `$context`: the task context, which will be updated by the current operation
- `$task.output`: the raw output of the task

This approach is acceptable. My only concern is that, in most cases, we would prefer to use the raw output to update the context rather than the transformed output, which is intended specifically for use by the next task.

cdavernas:
> From my understanding of the specification, $output refers to the raw output, and there is no way to access the "transformed" output because output and export are processed simultaneously. However, it is reasonable to consider that export might be processed after output.

That's incorrect imho. The raw ouput was should be suppliedby $output, which is the reason why I created it. The transformed output is accessed with `.`

matthias-pichler-warrify:
> The transformed output is accessed with `.`

Currently it is not because `.` refers to the context object. Making it refer to the transformed output is what option b) proposes. 

As it stands now only one of the raw or transformed output is available ... whatever `$output` refers to currently 🤷‍♂️

matthias-pichler-warrify:
I think both options of simultaneous and in order evaluation are possible. Whichever we define as the intended semantics (independent of implementation). 

Making them simultaneous forces `export.as` to use the raw output either as `.` or `$output`. I definitely think that the raw output should be available to export in some form though

fjtirado:
> order evaluation are possible. Whichever we define as the intended semantics (independent of implementation).
> 
> Making them simultaneous forces `export.as` to use the raw output either as `.` or `$output`. I definitely think that the raw output should be available to export in some form though

I think we should impose an order to avoid ambiguity, so `.` is the transformed output and` $task.output` the original output always, regardless the implementation

cdavernas:
> Currently it is not because . refers to the context object.

That is not as ot was initially intended, because unlike the raw input, the context was accessible thanks to `$context`.

cdavernas:
> I think both options of simultaneous and in order evaluation are possible

Simultaneously makes no sense to me, but that's just my opinion. As a matter if fact, it's counter intuitive, error prone and confusing. Plus, I doubt that implementers will rely on parallel processing to do such  trivial evaluations, therefore they will arbitrarily evaluate one or the other.

cdavernas:
> Historically, output.as and export.as were two operations related to output, to and from. These operations were processed "at the same moment, after the task returned its output" 

Thanks for the clarifications, but the interpretation imho remains the same for an external person.

However, I do not know where you take the "historical" information of those being processed at the same time because, as explained above, when we wrote the refactor, it was very clear, at least to me, that output.from was processed before output.to, thus the need to create an `$output` parameter. Additionally, afaik, before refactor there was no simultaneity either.

cdavernas:
To conclude, my opinion is that we should indicate that the `.` of the `export.as` should be the transformed output, and to keep `$output` parameter (which is more consistent, knowing we already declare an `$input` paramater, too). IMHO, it's perfectly fine that the info might be retrieved using `$task.output`, too: it just is an additional convenience. This would ensure we make no exception to `.` representing the "current" data. Proceeding otherwise might be confusing, at least it is to me.

Whatever we choose to proceed with, we should clarify that output should be transformed BEFORE exporting. This is IMHO more consistent, more intuitive and more logical.

What do you guys think?

matthias-pichler-warrify:
I think the split of `output.from` and `output.to` introduced some drift so the spec is no longer coherent. We should reason from the ground up and I will adapt the spec accordingly. I will try to consolidate the discussion:

1. If we keep `$output` it should most likely refer to the transformed output since `$input` also refers to the transformed input (as documented [here](https://github.com/serverlessworkflow/specification/blob/main/dsl.md#runtime-expression-arguments)). Otherwise it would be inconsistent for authors

1. If we introduce `$task.output` it should probably refer to the _raw_ output since #951 specifies the `$task.input` argument as the raw input

1. `export.as` should **definitely** have access to the raw output since I might want to store something in the context that might not be relevant to the next task/output

1. `export.as` should **probably** have access to the transformed output 🤷 ? If yes then a order is automatically imposed since `output.as` has to produce the transformed output first. Otherwise they still wouldn't be "independent" of each other. After closer inspection order does still matter since `output.as` has access to `$context` which is modified by `export.as` and thus order still matters and `output.as` before `export.as` seems to make more sense/be more intuitve.

1. Other runtime expressions such as `input.from` and `output.as` have their evaluation context be the same as the data they are transforming. This means `input.from` and `output.as` both "default" to the identity expression. In jq this would be `${ . }`. Since the `export.as` expression modifies the value of `$context` keeping this property would mean that the evaluation context of `export.as` should be the context itself. 
I however believe the authors will not think about `export.as` this way and will rather reason that `output.as` and `export.as` will either "be downstream of each other" or "siblings". So I think this point is not very strong.

I think this leaves us with the following options, that we can vote on:

# Option 1

- `export.as` is evaluated on the _transformed_ output. i.e. `export.as` is downstream of `output.as`
- Order is automatically imposed
- `$output` does _NOT_ need to exist
- `$task.output` will refer to the _raw_ output and will be set in both `output.as` (although redundant) and `export.as`

# Option 2

- `export.as` is evaluated on the _raw_ output.
- `$output` does _NOT_ exist and `export.as` does not have access to the transformed output. Thus some processing might need to be duplicated
- Order is still imposed since `output.as` relies on `$context`.
- If we want we can still introduce `task.output` to refer to the _raw_ output but it is no longer needed

# Option 3

- `export.as` is evaluated on the current context
- `$output` needs to exist and will refer to the _transformed_ output to be consistent with `$input`
- Order is automatically imposed
- `$task.output` will refer to the _raw_ output and will be set in both `output.as` (although redundant) and `export.as`

matthias-pichler-warrify:
I prefer either option 1 or option 2. 

> default runtime expressions all are the identity (i.e. ${ . })

I don't think this is a strong pro argument for option 3 and "purely theoretical"

cdavernas:
I prefer option #4 too, which was the initial intent when writing the refactor (as has already been implemented in Synapse's alpha branch more than a month ago :p).

I don't see the problem with having "one more argument", on the contrary, as they are but a convenience.

cdavernas:
> If we want we can still introduce $task.output to refer to the raw output but it is no longer needed

Even though not required, I'd personally still add it for the sake of consistency. Plus, even writing an expression using `$task`, authors might find it clearer.

JBBianchi:
> 5. Other runtime expressions such as `input.from` and `output.as` have their evaluation context be the same as the data they are transforming. This means `input.from` and `output.as` both "default" to the identity expression. In jq this would be `${ . }`. Since the `export.as` expression modifies the value of `$context` keeping this property would mean that the evaluation context of `export.as` should be the context itself.
>  I however believe the authors will not think about `export.as` this way and will rather reason that `output.as` and `export.as` will either "be downstream of each other" or "siblings". So I think this point is not very strong.

I agree with all the points, but the last point needs some nuance IMO. An important aspect to consider is that the evaluation contexts of `input.from` and `output.as` are respectively the raw input and output because those are the results of the previous step in the execution pipeline, not because of the target of the transformation. Therefore, it's a bit of a stretch to suggest that the evaluation context for export.as should be the $context. If we choose to position export.as as the step following output.as, it would be logical for its evaluation context to be the transformed output.

So to keep the consistency of the *pipeline* concept, I'd rather opt for **Options 1** .
Maybe with a twist of using `$output` instead of `$task.output`. The later is a bit useless (`output.as` already have access to the raw output as its evaluation context) and I don't think it brings more clarity than `$output`, it's just makes it longer to type.

matthias-pichler-warrify:
@JBBianchi 
> Therefore, it's a bit of a stretch to suggest that the evaluation context for export.as should be the $context.

I agree. It's a weak argument I just included for completeness

> Maybe with a twist of using $output instead of $task.output.

For me that is honestly a non-starter because it would be SUPER confusing to have `$input` be the _transformed_ input and `$output` be the _raw_ output.

I think that since `export.as` has access to the raw output via `$task.output` and the transformed output using `$output` I am flexible on the evaluation context of `export.as` being either the raw or transformed output

matthias-pichler-warrify:
So I think @cdavernas , @JBBianchi  and I are aligned now 🎉 

@fjtirado @ricardozanini what are your thoughts?

fjtirado:
Im a man of principles, so I still think Option1 is the way to go

fjtirado:
The use case I see for export (context) having access to the transformed output is the same one than when we want to just store the original output of the task. Maybe we have a future task (not the consecutive one) that want to do something with the transformed output. In other words, I think we cannot predict if used would like to store the original output or the transformed one in the context, it will heavily depends on the scenario. I also think forcing an order between output.as and export.as is a good thing.

matthias-pichler-warrify:
@fjtirado 
> Im a man of principles, so I still think Option1 is the way to go

which principle drives your opinion here? The only difference between option 1 and what is currently documented is that we define `$output` additionally

>  I also think forcing an order between output.as and export.as is a good thing.

Absolutely 👍 but as I stated in my summary. Order is always `output.as` before `export.as` (in all options). This is because if `output.as` uses `$context` we need to define that it should refer to the "old" context and not the one `export.as` produces 👍 This is currently already documented this way in this PR

>  I think we cannot predict if users would like to store the original output or the transformed one in the context

I agree. Authors have to have access to both raw & transformed output in the `export.as` expressions. Both Option 1  and the current agreement allow this 👍 

To summarize:
The only difference between Option 1 and what we currently have is that `$output` is defined IN ADDITION (even though redundant). And thus my question is: Are you against defining `$output` additionally?

matthias-pichler-warrify:
I will summarize what we have right now because it does differ from option 4:

# this PR

- `export.as` is evaluated on the _transformed_ output.
- `$output` does exist and contains the transformed output.
- Order is automatically imposed (output.as -> export.as)
- $task.output refers to the _raw_ output

dsl.md
```diff
@@ -146,39 +146,97 @@ Once the task has been executed, different things can happen:
...
+*Example: If the first task returns a large dataset, a transformation can be applied to retain only the relevant results needed for subsequent tasks.*
 
-4. **Filter Last Task Input**
-Before the last task in the workflow executes, its input data can be filtered to ensure it receives only the necessary information. This step is crucial for ensuring that the final task has all the required data to complete the workflow successfully.
+4. **Transform Last Task Input**
+Before the last task in the workflow executes, its input data can be transformed to ensure it receives only the necessary information. This can be done using the task's `input.from` expression. It evaluates on the transformed workflow input and defaults to the identity expression which leaves the input unchanged. The result of this expression will be set as the `$input` runtime expression argument and be passed to the task. Any runtime expressions used within the task definition will be evaluated using this transformed input. This step is crucial for ensuring that the final task has all the required data to complete the workflow successfully.
```
ricardozanini, line 167:
```suggestion
Before the last task in the workflow executes, its input data can be transformed to ensure it receives only the necessary information. This can be done using the task's `input.from` expression. It evaluates the transformed workflow input and defaults to the identity expression, which leaves the input unchanged. The result of this expression will be set as the `$input` runtime expression argument and be passed to the task. This transformed input will evaluate any runtime expressions used within the task definition. This step is crucial for ensuring the final task has all the required data to complete the workflow successfully.
```

dsl.md
```diff
@@ -146,39 +146,97 @@ Once the task has been executed, different things can happen:
...
 
-*Example: If the last task involves generating a report, the input filter can ensure that only the data required for the report generation is passed to the task.*
+*Example: If the last task involves generating a report, the input transformation can ensure that only the data required for the report generation is passed to the task.*
 
-5. **Filter Last Task Output**
-After the last task completes, its output can be filtered before it is considered as the workflow output. This ensures that the workflow produces a clean and relevant output, free from any extraneous data that might have been generated during the task execution.
+5. **Transform Last Task Output**
+After the last task completes, its output can be transformed before it is considered as the workflow output. Transformations are applied using the `output.as` runtime expression. It evaluates on the raw task output and defaults to the identity expression which leaves the output unchanged. Its result will be passes to the workflow `output.as` runtime expression. This ensures that the workflow produces a clean and relevant output, free from any extraneous data that might have been generated during the task execution.
```
ricardozanini, line 172:
```suggestion
After the last task completes, its output can be transformed before it is considered the workflow output. Transformations are applied using the `output.as` runtime expression. It evaluates the raw task output and defaults to the identity expression, which leaves the output unchanged. Its result will be passed to the workflow `output.as` runtime expression. This ensures that the workflow produces a clean and relevant output, free from any extraneous data that might have been generated during the task execution.
```

dsl.md
```diff
@@ -146,39 +146,97 @@ Once the task has been executed, different things can happen:
...
 
-*Example: If the last task outputs various statistics, a filter can be applied to retain only the key metrics that are relevant to the stakeholders.*
+*Example: If the last task outputs various statistics, a transformation can be applied to retain only the key metrics that are relevant to the stakeholders.*
 
-6. **Filter Workflow Output**
-Finally, the overall workflow output can be filtered before it is returned to the caller or stored. This step ensures that the final output of the workflow is concise and relevant, containing only the necessary information that needs to be communicated or recorded.
+6. **Transform Workflow Output**
+Finally, the overall workflow output can be transformed before it is returned to the caller or stored. Transformations are applied using the `output.as` runtime expression. It evaluates on the last task's output and defaults to the identity expression which leaves the output unchanged. This step ensures that the final output of the workflow is concise and relevant, containing only the necessary information that needs to be communicated or recorded.
```
ricardozanini, line 177:
```suggestion
Finally, the overall workflow output can be transformed before it is returned to the caller or stored. Transformations are applied using the `output.as` runtime expression. It evaluates the last task's output and defaults to the identity expression, which leaves the output unchanged. This step ensures that the final output of the workflow is concise and relevant, containing only the necessary information that needs to be communicated or recorded.
```

dsl.md
```diff
@@ -146,39 +146,97 @@ Once the task has been executed, different things can happen:
...
-*Example: If the workflow's final output is a summary report, a filter can ensure that the report contains only the most important summaries and conclusions, excluding any intermediate data.*
+*Example: If the workflow's final output is a summary report, a transformation can ensure that the report contains only the most important summaries and conclusions, excluding any intermediate data.*
 
-By applying filters at these strategic points, Serverless Workflow DSL ensures that data flows through the workflow in a controlled and efficient manner, maintaining clarity and relevance at each stage of execution. This approach helps in managing complex workflows and ensures that each task operates with the precise data it requires, leading to more predictable and reliable workflow outcomes.
+By applying transformations at these strategic points, Serverless Workflow DSL ensures that data flows through the workflow in a controlled and efficient manner, maintaining clarity and relevance at each stage of execution. This approach helps in managing complex workflows and ensures that each task operates with the precise data it requires, leading to more predictable and reliable workflow outcomes.
```
ricardozanini, line 181:
```suggestion
By applying transformations at these strategic points, Serverless Workflow DSL ensures that data flows through the workflow in a controlled and efficient manner, maintaining clarity and relevance at each execution stage. This approach helps manage complex workflows and ensures that each task operates with the precise data required, leading to more predictable and reliable workflow outcomes.
```

dsl.md
```diff
@@ -146,39 +146,97 @@ Once the task has been executed, different things can happen:
...
+
+Visually this can be represented as follows:
```
ricardozanini, line 183:
```suggestion
Visually, this can be represented as follows:
```

dsl.md
```diff
@@ -236,11 +296,23 @@ This argument contains information about the runtime executing the workflow.
...
 | startedAt.epochMillis | `integer` | The start time of the execution as a integer value of milliseconds since midnight of 1970-01-01 UTC | `1641024000123` (="2022-01-01T08:00:00.123Z") |
 | startedAt.epochNanos | `integer` | The start time of the execution as a integer value of nanoseconds since midnight of 1970-01-01 UTC | `1641024000123456` (="2022-01-01T08:00:00.123456Z") |
 
+The following table shows which arguments are available for each runtime expression:
+
+| Runtime Expression | Evaluated on | Produces | `$context` | `$input` | `$output` | `$secrets` | `$task` | `$workflow` |
+|:-------------------|:---------:|:---------:|:---------:|:---------:|:-------:|:---------:|:-------:|:----------:|
+| Workflow `input.from` | Raw workflow input | Transformed workflow input | | | | ✔ | | ✔ |
+| Task `input.from` | Raw task input (i.e. transformed workflow input for first task, transformed output from previous task otherwise) | Transformed task input | ✔ | | | ✔ | ✔ | ✔ |
```
ricardozanini, line 309:
```suggestion
| Task `input.from` | Raw task input (i.e. transformed workflow input for the first task, transformed output from previous task otherwise) | Transformed task input | ✔ | | | ✔ | ✔ | ✔ |
```

dsl.md
```diff
@@ -146,39 +146,97 @@ Once the task has been executed, different things can happen:
...
+*Example: If the workflow receives a JSON object as input, a transformation can be applied to remove unnecessary fields and retain only those that are required for the workflow's execution.*
 
-2. **Filter First Task Input**
-The input data for the first task can be filtered to match the specific requirements of that task. This ensures that the first task receives only the necessary data it needs to perform its operations.
+2. **Transform First Task Input**
+The input data for the first task can be transformed to match the specific requirements of that task. This ensures that the first task receives only the necessary data it needs to perform its operations. This can be done using the task's `input.from` expression. It evaluates on the transformed workflow input and defaults to the identity expression which leaves the input unchanged. The result of this expression will be set as the `$input` runtime expression argument and be passed to the task. Any runtime expressions used within the task definition will be evaluated using this transformed input.
```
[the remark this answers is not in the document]
cdavernas: > This transformed input will evaluate any...

Better yet imho:

This transformed input will be evaluated against any...

dsl.md
```diff
@@ -146,39 +146,97 @@ Once the task has been executed, different things can happen:
...
 
-3. **Filter First Task Output**
-After the first task completes, its output can be filtered before passing it to the next task or storing it in the workflow context. This helps in managing the data flow and keeping the context clean by removing any unnecessary data produced by the task.
+3. **Transform First Task Output**
+After the first task completes, its output can be transformed before passing it to the next task or storing it in the workflow context. Transformations are applied using the `output.as` runtime expression. It evaluates on the raw task output and defaults to the identity expression which leaves the output unchanged. Its result will be the input to the next task. To update the context one uses the `export.as` runtime expression. It evaluates on the raw output and defaults to the expression that returns the existing context. The result of this runtime expression replaces the workflows current context and the content of the `$context`runtime expression argument. This helps in managing the data flow and keeping the context clean by removing any unnecessary data produced by the task.
```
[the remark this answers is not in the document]
cdavernas: > Its result will be input for the next task.

Shouldn't it be:

Its result will be input to the next task.

ricardozanini:
I didn't go through our examples, but if we need something that documents the transformation, we should add it to this PR.

dsl.md
cdavernas: Same comment as above: input does not evaluate expression, it uis evaluated against expressions, right?

dsl.md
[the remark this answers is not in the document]
ricardozanini: I think `for` fits here better. And that's what the AI is saying. 🤷

matthias-pichler-warrify:
> I didn't go through our examples, but if we need something that documents the transformation, we should add it to this PR.

I went through all examples and ctk cases and didn't find anything that needed updating. The one example that used `export.as` was added by me and already used the current spec 😅  

But thanks to your reminder I found one example in the dsl-reference that needed updating 🙏

ricardozanini (approved):

PR #969, opened 2024-08-09
Fix custom function documentation

### LINKED ISSUE #962: Fix the custom function documentation

Fix the documentation for custom functions.

Namely, we need to:

- [ ] Remove top level name in `function.yaml`, as the name is defined by the function's directory
- [ ] Indicate that custom functions can be referenced using the URI pointing at the repository directory that defines the `function.yaml` OR using the concatenation `{name}:{version}`. This is however only be supported for cataloged functions.
- [ ] Indicate that runtime *may* define mechanisms to register additional function catalogs, in which case the above concatenation can also be used for functions defined in those.

**Why is this needed**:

The documentation drifted away from latest changes made to the SW Catalog.

cdavernas:

- [x] Specification
- [ ] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:
#962

**What this PR does**:

Fixes the custom function documentation

Closes #962


dsl.md
```diff
@@ -466,20 +474,21 @@ document:
   namespace: default
   name: customFunctionWorkflow
   version: '0.1.0'
-
 do:
   - validateEmail:
-      call: https://github.com/myorg/functions/validateEmailAddress@v1
+      call: https://raw.githubusercontent.com/myorg/main/validateEmailAddress/1.0.0
```
ricardozanini, line 479: FWIW, `https://github.com/myorg/functions/validateEmailAddress@v1` also works nicely. Users can tag their functions on their private repos instead of a direct link to the raw file. It's elegant.

ricardozanini (approved):

dsl.md
cdavernas: I'l replace it then, so that we have both examples!

PR #987, opened 2024-08-21
Document the difference between an event-driven schedule and a startup listen task

### LINKED ISSUE #972: Exhaustively document difference between `schedule.on` and a start `listen` task

Exhaustively document difference between `schedule.on` and a start `listen` task.

**Why is this needed**:

Ensure users understand the (subtle) difference between the two approaches.

### LINKED ISSUE #977: Document workflow input when relying on event based schedule

Document workflow input when relying on event based schedule.

**Why is this needed**:

So far, nothing in the DSL documents what should be done with the event(s) consumed by the event-based schedule of a workflow, which are therefore lost to authors.

cdavernas:

- [x] Specification
- [ ] Schema
- [x] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:

Fixes #972
Fixes #977

**What this PR does**:

- Documents the difference between an event-driven schedule and a startup listen task
- Documents the input of a workflow that has been started by an event-driven schedule
- Adds examples to document event-driven and cron schedules


dsl.md
```diff
@@ -150,6 +151,24 @@ Workflow scheduling in ServerlessWorkflow allows developers to specify when and
 
 See the [DSL reference](dsl-reference.md#schedule) for more details about workflow scheduling.
 
+##### Event-driven scheduling
+
+###### Input of event-driven scheduled workflows
+
+In event-driven scheduled workflows, the input is structured as an array containing the events that trigger the execution of the workflow. This array serves as a vital resource, providing workflow authors with access to all relevant data associated with each triggering event. When an event activates the workflow, it populates this array with one or more occurrences, allowing authors to process multiple events simultaneously as needed.
```
ricardozanini, line 158:
```suggestion
In event-driven scheduled workflows, the input is structured as an array containing the events that trigger the execution of the workflow. This array serves as a vital resource, providing workflow authors access to all relevant data associated with each triggering event. When an event activates the workflow, it populates this array with one or more occurrences, allowing authors to process multiple events simultaneously as needed.
```

We could simplify `This array servers as a vital resource` to `This array is a vital resource`. It's your call, but the simpler the phrases, the better, IMO.

dsl.md
```diff
@@ -150,6 +151,24 @@ Workflow scheduling in ServerlessWorkflow allows developers to specify when and
...
+
+Authors can reference individual events within the array using syntax such as $workflow.input[index], where index indicates the position of the event, starting from 0. For instance, $workflow.input[0] refers to the first event, while $workflow.input[1] refers to the second. This structure allows for easy access to specific event details, and if multiple events are received at once, authors can iterate through the array to handle each one appropriately. This flexibility ensures that workflows can respond effectively to various conditions and triggers, enhancing their overall responsiveness and functionality.
```
ricardozanini, line 160:
```suggestion
Authors can reference individual events within the array using syntax such as $workflow.input[index], where index indicates the event's position, starting from 0. For instance, $workflow.input[0] refers to the first event, while $workflow.input[1] refers to the second. This structure allows for easy access to specific event details, and if multiple events are received at once, authors can iterate through the array to handle each one appropriately. This flexibility ensures that workflows can respond effectively to various conditions and triggers, enhancing their overall responsiveness and functionality.
```

dsl.md
```diff
@@ -150,6 +151,24 @@ Workflow scheduling in ServerlessWorkflow allows developers to specify when and
...
+
+###### Distinguishing event-driven scheduling from start `listen` Tasks
+
+While both `schedule.on` and a start listener task enable event-driven execution of workflows, they serve distinct purposes and have different implications:
+
+- **`schedule.on`**: This property defines when a new instance of a workflow should be created based on an external event. When an event occurs that matches the criteria specified in `schedule.on`, a new workflow instance is initiated. The key point here is that `schedule.on` solely manages the creation of new workflow instances. Any faults or timeouts related to the scheduling process itself are typically invisible to the user and do not impact the workflow instance.
```
ricardozanini, line 166:
```suggestion
- **`schedule.on`**: This property defines when a new workflow instance should be created based on an external event. When an event matches the criteria specified in `schedule.on`, a new workflow instance is initiated. The critical point here is that `schedule.on` solely manages the creation of new workflow instances. Any faults or timeouts related to the scheduling process are typically invisible to the user and do not impact the workflow instance.
```

dsl.md
```diff
@@ -150,6 +151,24 @@ Workflow scheduling in ServerlessWorkflow allows developers to specify when and
...
+
+- **Start `listen` task**: A start listener task defines a task that must be undertaken after a new workflow instance has been created. This task listens for specific events and begins processing once the instance is active. The critical difference lies in the fact that a start listener task operates within the context of an already instantiated workflow. If a start listener task experiences a timeout or fault, it can cause the entire workflow instance to fail or behave unexpectedly, directly impacting the flow's execution and outcome.
```
ricardozanini, line 168:
```suggestion
- **Start `listen` task**: A start listener task defines a task that must be undertaken after a new workflow instance has been created. This task listens for specific events and begins processing once the instance is active. The critical difference is that a start listener task operates within an already instantiated workflow. If a start listener task experiences a timeout or fault, it can cause the entire workflow instance to fail or behave unexpectedly, directly impacting the flow's execution and outcome.
```

dsl.md
```diff
@@ -150,6 +151,24 @@ Workflow scheduling in ServerlessWorkflow allows developers to specify when and
...
+
+In essence, while `schedule.on` is concerned with *when* a new workflow instance should be initiated, a start listener task deals with *what* should happen once the instance is active. This distinction is important because it influences how errors and timeouts are handled—`schedule.on` faults are typically invisible and do not affect the workflow, whereas start listener task failures can have a direct and potentially severe impact on the workflow instance they belong to.
```
ricardozanini, line 170:
```suggestion
While `schedule.on` is concerned with *when* a new workflow instance should be initiated, a start listener task deals with *what* should happen once the instance is active. This distinction is crucial because it influences how errors and timeouts are handled—`schedule.on` faults are typically invisible and do not affect the workflow, whereas start listener task failures can directly and potentially severely impact the workflow instance they belong to.
```

examples/event-driven-schedule.yaml
```diff
@@ -0,0 +1,24 @@
...
+  dsl: 1.0.0-alpha1
+  namespace: examples
+  name: event-driven-schedule
+  version: 1.0.0-alpha1
+schedule:
+  on:
+    one:
+      with:
+        type: com.example.hospital.events.patients.heartbeat.low
+do:
```
ricardozanini, line 11: Don't we have to use a listener task instead in this example?

cdavernas: Depends on what you want to achieve.

The objective here is to trigger a new workflow when and only when the event is received.

Using a listen task would mean you'd have started a workflow out of the blue, then wait potentially forever for an event to be consumed.

ricardozanini: Based on our private conversation, we agreed to follow up with an issue to review the `schedule.on` naming since it may confuse users. Some may expect that `schedule` enforces a timestamp definition.

ricardozanini (approved):

PR #995, opened 2024-08-28
Add a new property used to determine whether or not to await the execution of a process 

### LINKED ISSUE #991: Add a new property used to determine whether or not to await the execution of a process

Add a new property used to determine whether or not to await the execution of a process

**Why is this needed**:

Allows authors to fire-and-forget a process, such as a workflow, just like it was possible in previous versions of the DSL.

My proposal is to add a new `await` (`boolean`) property to the `run` concept, thus making that option available to `container`, `script`, `shell` and `workflow` processes.

cdavernas:

- [x] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:

Closes #991 

**What this PR does**:

- Adds a new property used to determine whether or not to await the execution of a process

**Additional information:**

- Removes obsolete `.markdownlint.yaml`,  `.typo-ci.yml` and `Makefile` obsolete files


Makefile
```diff
@@ -1,5 +0,0 @@
-# will add the license headers to all of our files
-# requires go installed in your system: https://golang.org/doc/install
-addheaders:
```
ricardozanini, line 3: This is actually pretty useful. We can add the install command to install this binary and make it OS-independent.

cdavernas: Must it be at the repository's root, though? Can we remove it for now and add the necessary command with a dedicated PR?

ricardozanini: Can you leave this file here so I can address this in a new PR?

cdavernas: Sure! Will re-add!

ricardozanini (approved):

Issue #923, opened 2024-07-16
Default case in switch


matthias-pichler:
**What would you like to be added**:

Currently the DSL states for the [switch case](https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#switch-case) that: `when`

> A runtime expression used to determine whether or not the case matches.
> If not set, the case will be matched by default if no other case match.
> Note that there can be only one default case, all others MUST set a condition.

This wording implies that the default case does not follow ordering because it is only `matched by default if no other case match`

therefore these two would be equivalent:

```yaml
document:
  dsl: 1.0.0-alpha1
  namespace: test
  name: sample-workflow
  version: 0.1.0
do:
  - processOrder:
      switch:
        - case1:
            when: .orderType == "electronic"
            then: processElectronicOrder
        - case2:
            when: .orderType == "physical"
            then: processPhysicalOrder
        - default:
            then: handleUnknownOrderType
``` 

and 

```yaml
document:
  dsl: 1.0.0-alpha1
  namespace: test
  name: sample-workflow
  version: 0.1.0
do:
  - processOrder:
      switch:
        - case1:
            when: .orderType == "electronic"
            then: processElectronicOrder
        - default:
            then: handleUnknownOrderType
        - case2:
            when: .orderType == "physical"
            then: processPhysicalOrder
``` 

I think this is a bit ambiguous and would change the wording to something like:

> If `when` is not specified it always matches.

This would then mean that the only logical place to put a default case would be at the end. Interested in what you folks think.

> Note that there can be only one default case, all others MUST set a condition.

This is currently not validated by the schema. ~I think it could be~ It is doable with [`contains`](https://json-schema.org/understanding-json-schema/reference/array#contains). This schema ensures only one case has no `when`

```yaml
type: object
required: [ switch ]
unevaluatedProperties: false
properties:
  switch:
    type: array
    minItems: 1
    contains:
      type: object
      minProperties: 1
      maxProperties: 1
      additionalProperties:
        type: object
        properties:
          when: false
    minContains: 0
    maxContains: 1
    items:
      type: object
      minProperties: 1
      maxProperties: 1
      title: SwitchItem
      additionalProperties:
        type: object
        title: SwitchCase
        required: [ then ]
        properties:
          when:
            type: string
            description: A runtime expression used to determine whether or not the case matches.
          then:
            $ref: '#/$defs/flowDirective'
            description: The flow directive to execute when the case matches.
```

[...]

PR #999, opened 2024-08-29
Remove default case in switch case list

matthias-pichler:

- [x] Specification
- [x] Schema
- [x] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:

#923

**What this PR does**:

I updated the schema to make the `when` property required. A defualt case now has to be defined using the task level `then` property

**Additional information:**

### DIFF
--- a/.ci/validation/test/fixtures/invalid/switch-missing-when.yaml
+++ b/.ci/validation/test/fixtures/invalid/switch-missing-when.yaml
@@ -1,0 +1,45 @@
+document:
+  dsl: 1.0.0-alpha2
+  namespace: test
+  name: switch-missing-when
+  version: 0.1.0
+do:
+  - processOrder:
+      switch:
+        - case1:
+            when: .orderType == "electronic"
+            then: processElectronicOrder
+        - case2:
+            when: .orderType == "physical"
+            then: processPhysicalOrder
+        - default:
+            then: handleUnknownOrderType
+  - processElectronicOrder:
+      do:
+        - validatePayment:
+            set:
+              validate: true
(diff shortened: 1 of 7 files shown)


.ci/validation/test/fixtures/invalid/switch-missing-when.yaml
```diff
@@ -0,0 +1,45 @@
...
+  - processOrder:
+      switch:
+        - case1:
+            when: .orderType == "electronic"
+            then: processElectronicOrder
+        - case2:
+            when: .orderType == "physical"
+            then: processPhysicalOrder
+        - default:
+            then: handleUnknownOrderType
```
ricardozanini, line 16:
```suggestion
          then: handleUnknownOrderType
```

cdavernas (changes requested):
As  I commented in the issue, I'm not for removing the default, whenless case, but rather to properly document ordering and the possibility to instead use `then`.

I wasn't aware a consensus was reached on that aspect, which breaks graphical representation in many contexts, or forces authors to define another useless case. 

As said in the issue, it's also a regression from what was done in 0.8/0.9, namely giving the possibility to actually name a default case.

.ci/validation/test/fixtures/invalid/switch-missing-when.yaml
matthias-pichler: this is a intentionally failing example ... see file path

flow6979 (approved):

PR #1041, opened 2024-12-16
[Fix #1040] Set behaviour clarification

### LINKED ISSUE #1040: Set behaviour clarification
It is unclear if Set task should replace the output of the workflow with the object result of the set task or it should add the properties to the existing input (so the output is actually merge of the input plus the set object)

**Why is this needed**:
One of the [examples](https://github.com/serverlessworkflow/specification/blob/main/examples/switch-then-string.yaml#L26-L37) seems to take the merge approach (otherwise the first two sets are dead code), while one of the [features](https://github.com/serverlessworkflow/specification/blob/main/ctk/features/set.feature#L16-L20) seems to take the replacement approach (otherwise it would have just set the shape)

After some discussion (there are different view on that) , the feature wins over the example. The example needs to be corrected and the specification should clarify the behaviour of the task.

fjtirado:
Fix https://github.com/serverlessworkflow/specification/issues/1040

- [X] Specification
- [ ] Schema
- [X] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:

**What this PR does**:

**Additional information:**


cdavernas (approved):
LGTM! Cheers ❤️

ricardozanini (approved):

Issue #1013, opened 2024-09-13
Add a property to configure the lifetime of spawned containers


cdavernas:
**What would you like to be added**:

Add a property to configure the lifetime of spawned containers.

**Why is this needed**:

Gives authors the ability to decide whether or not to delete containers after executing them, and to eventually define their Time To Live.

PR #1049, opened 2025-01-08
Add a new `lifetime` property to the `container` process

cdavernas:

- [x] Specification
- [x] Schema
- [x] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:
#1013

**What this PR does**:
- Adds a new `lifetime` property to the `container` process

### DIFF
--- a/dsl-reference.md
+++ b/dsl-reference.md
@@ -55,6 +55,7 @@
   + [HTTP Response](#http-response)
   + [HTTP Request](#http-request)
   + [URI Template](#uri-template)
+  + [Container Lifetime](#container-lifetime)
 
 ## Abstract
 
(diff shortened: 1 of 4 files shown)


ricardozanini (commented):
Should this control not be on the platform instead? In which cases would a user like to control the infra behavior from the workflow? I can see plenty since users are creative, but it is a dangerous mix of functional and non-functional requirements.

cdavernas:
> Should this control not be on the platform instead?

Not really, no, as this should be configurable on a per-container basis. Some containers will just run dumb work which no one cares about double checking, whereas others might be critical and require auditing. 
The PR addresses an issue that was long opened, and which addresses a couple of professional requirements.

ricardozanini:
>  Not really, no, as this should be configurable on a per-container basis

I agree. The platform should know the containers a given workflow would spawn, and an operator/admin could manage it. 

My point is that in the long term, we will add new things like `maxMemory`, `cpu`, etc. Replicating what platforms like Kubernetes or even Docker Compose have already declared.

ricardozanini:
No, I think it's valuable. I am just raising a concern about what can grow more complex if we open Pandora's box. We can keep this one as long as we don't start transforming the `container` object into a Kubernete's container/pod mimic.

ricardozanini (approved):

Issue #1030, opened 2024-11-20
Cloud events to be publlished when workflow status change 


fjtirado:
Possible workflow [status](https://github.com/serverlessworkflow/specification/blob/main/dsl.md#status-phases) are defined in the specifcation. However, there is not information of the content of the cloud event that very likely any implementation might publish when status is changed. 

**What would you like to be added**:
Common content of the cloud events to be published when a status is changed

**Why is this needed**:
To make it possible users of any implementation to rely on the common content when listening to a status phase change event.

fjtirado:
The first thing (my guess) that we should decide is between: 

1. Publishing a different event type per status phase: started, aborted, complete...
2. Publishing the same event type (StatusChangeEvent) and include the old and the new status
3. Other ;)

cdavernas:
I'd definitly go for your first suggestion, which makes routing/processing easier and more compliant with the Cloud Event Spec: the payload of the events should be opaque to those concerns.

As for the events types, I'd recommend using the following:

| Type | Description |
|:------|:-------------|
| `io.serverlessworkflow.workflows.started.v1` | Notifies about the start of a workflow instance. |
| `io.serverlessworkflow.workflows.suspended.v1` | Notifies about suspending a workflow instance execution. |
| `io.serverlessworkflow.workflows.resumed.v1` | Notifies about resuming a workflow instance execution. |
| `io.serverlessworkflow.workflows.cancelled.v1` | Notifies about the cancellation of a workflow instance execution. |
| `io.serverlessworkflow.workflows.faulted.v1` | Notifies about a workflow instance being faulted. |
| `io.serverlessworkflow.workflows.completed.v1` | Notifies about the completion of a workflow instance execution. |
| `io.serverlessworkflow.workflows.correlation-started.v1` | Notifies about a workflow instance starting to correlate events. |
| `io.serverlessworkflow.workflows.correlation-completed.v1` | Notifies about a workflow instance completing an event correlation. |
| `io.serverlessworkflow.tasks.started.v1` | Notifies about the start of a task. |
| `io.serverlessworkflow.tasks.suspended.v1` | Notifies about suspending a task's execution. |
| `io.serverlessworkflow.tasks.resumed.v1` | Notifies about resuming a task's execution. |
| `io.serverlessworkflow.tasks.cancelled.v1` | Notifies about the cancellation of a task's execution. |
| `io.serverlessworkflow.tasks.faulted.v1` | Notifies about a task being faulted. |
[...]

Issue #1024, opened 2024-10-25
Add a new `suspended` workflow status phase


cdavernas:
**What would you like to be added**:

Reintroduce the `suspended` workflow status phase

**Why is this needed**:

Adds a status phase that clearly describes that the workflow has been manually and explicitly suspended by a user, as opposed to `waiting`, which describes that the workflow is waiting for event(s) to be correlated.

**Additional notes**:

The `suspended` status phase was initialy part of the DSL, but was removed following (very valid) observations from @fjtirado, who considered that `waiting` could actually represent both phases.

However, following the implementation of the `suspend` feature in Synapse, we realized that users could not differenciate flows that were explicitly suspended from ones that were being correlated, thus the feature request.

fjtirado:
ok, so waiting is a "natural" status, the workflow reach it as part of its execution while suspended is "forced" one, explicitly set by the user. 
My only concern is that suspending a workflow should not necessarily be implemented by all implementors. I mean, I can foresee an implementation that does not provide ways to suspend workflow execution and its still a compliant one, wdyt?

cdavernas:
> ok, so waiting is a "natural" status, the workflow reach it as part of its execution while suspended is "forced" one, explicitly set by the user.

That's my understanding of it, yes.

> I mean, I can foresee an implementation that does not provide ways to suspend workflow execution and its still a compliant one, wdyt?

@fjtirado Very valid point indeed. I personally don't see the requirement of enforcing such features. That status phase is IMO a nice to have in case one implements the related feature. Furthermore, adding it would allows us to exhaustively describe all the possible phases of a workflow, regardless of implemented features.

fjtirado:
As an alternative, I think we can broadly understand `waiting` as waiting for something (not just events), so it will also included the state "waiting for the user to remove the workflow he temporalily stopped" and then let implementos provides "substates" or "states additional description" freely. Becuase there might be implementations that might want to differentiate between "waiting for event" or "waiting for rest response" and this will allow them to do that.

fjtirado:
Or following a similar idea, add an additional `actor` field, which will include the "initiator", human or machine ;), that let the workflow to reach the current state
So in the "suspend" case, state will be waiting and actor user
In the "waiting for event" case., state will be waiting and actor event

fjtirado:
[...]

PR #1054, opened 2025-01-09
Add documentation about workflow and task lifecycle events

cdavernas:

- [x] Specification
- [ ] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:
#1024
#1030

**What this PR does**:
- Adds documentation about workflow and task lifecycle events


dsl-reference.md
```diff
@@ -1106,6 +1128,389 @@ Flow Directives are commands within a workflow that dictate its progression.
...
+##### Workflow Started Event
+
+The data carried by the cloud event that notifies that a workflow has started.
+
+###### Properties
+
+| Name | Type | Required | Description|
+|:--|:---:|:---:|:---|
+| name | `string` | `yes` | The qualified name of the workflow that has started. |
+| definition | [`workflowDefinitionReference`](#workflow-definition-reference) | `yes` | An an object that describes the definition of the workflow that has started. |
```
ricardozanini, line 1146:
```suggestion
| definition | [`workflowDefinitionReference`](#workflow-definition-reference) | `yes` | An object that describes the definition of the workflow that has started. |
```

dsl-reference.md
```diff
@@ -30,6 +30,27 @@
...
+      + [Workflow Correlation Started](#workflow-correlation-started-event)
+      + [Workflow Correlation Completed](#workflow-correlation-completed-event)
+      + [Workflow Cancelled](#workflow-cancelled-event)
+      + [Workflow Faulted](#workflow-faulted-event)
+      + [Workflow Completed](#workflow-completed-event)
+      + [Workflow Status Changed](#workflow-status-changed-event)
+    - [Task Lifecycle Events](#task-lifecycle-events)
+      + [Task Created](#task-created-event)
+      + [Task Started](#task-started-event)
+      + [Task Suspended](#task-suspended-event)
```
fjtirado, line 47: Now that I took a closer look, and after my yesterday comment related with workflow cancelation, Im not 100% sure a task can be really suspended. 
A cancelation of a workflow can be understood as "let the current task to finish and hold execution of the next one" rather than "Interrupt current task and let it in a limbo"

dsl-reference.md
```diff
@@ -1106,6 +1128,389 @@ Flow Directives are commands within a workflow that dictate its progression.
...
+faultedAt: '2024-07-26T16:59:57-05:00'
+error:
+  type: https://serverlessworkflow.io/spec/1.0.0/errors/communication
+  title: Service Not Available
+  status: 503
+```
+
+##### Workflow Completed Event
+
+The data carried by the cloud event that notifies that notifies that a workflow ran to completion.
```
ricardozanini, line 1279:
```suggestion
The data carried by the cloud event that notifies that a workflow ran to completion.
```

dsl-reference.md
```diff
@@ -1106,6 +1128,389 @@ Flow Directives are commands within a workflow that dictate its progression.
...
+completedAt: '2024-07-26T16:59:57-05:00'
+output:
+  orderId: '0a7f8581-acb9-4133-a378-0460c98ea60c'
+  petId: xt84hj202q14s
+  status: placed
+```
+
+##### Workflow Status Changed Event
+
+The data carried by the cloud event that notifies that notifies that the status phase of a workflow has changed.
```
ricardozanini, line 1302:
```suggestion
The data carried by the cloud event that notifies that the status phase of a workflow has changed.
```

dsl-reference.md
```diff
@@ -1106,6 +1128,389 @@ Flow Directives are commands within a workflow that dictate its progression.
...
+name: orderPetWorkflow-ix7iryakiem8j.samples
+completedAt: '2024-07-26T16:59:57-05:00'
+correlationKeys:
+  petId: xt84hj202q14s
+  orderId: '0a7f8581-acb9-4133-a378-0460c98ea60c'
+```
+
+##### Workflow Cancelled Event
+
+The data carried by the cloud event that notifies that notifies that a workflow has been cancelled.
```
ricardozanini, line 1238:
```suggestion
The data carried by the cloud event that notifies that a workflow has been canceled.
```

dsl-reference.md
```diff
@@ -1106,6 +1128,389 @@ Flow Directives are commands within a workflow that dictate its progression.
...
+###### Examples
+
+```yaml
+name: orderPetWorkflow-ix7iryakiem8j.samples
+cancelledAt: '2024-07-26T16:59:57-05:00'
+```
+
+##### Workflow Faulted Event
+
+The data carried by the cloud event that notifies that notifies that a workflow has faulted.
```
ricardozanini, line 1256:
```suggestion
The data carried by the cloud event that notifies that a workflow has faulted.
```

dsl-reference.md
```diff
@@ -1106,6 +1128,389 @@ Flow Directives are commands within a workflow that dictate its progression.
...
+
+```yaml
+workflow: orderPetWorkflow-ix7iryakiem8j.samples
+task: '/do/1/initialize'
+suspendedAt: '2024-07-26T16:59:57-05:00'
+```
+
+##### Task Resumed Event
+
+The data carried by the cloud event that notifies that notifies that a task has been resumed.
```
ricardozanini, line 1384:
```suggestion
The data carried by the cloud event that notifies that a task has been resumed.
```

dsl-reference.md
```diff
@@ -1106,6 +1128,389 @@ Flow Directives are commands within a workflow that dictate its progression.
...
+
+```yaml
+workflow: orderPetWorkflow-ix7iryakiem8j.samples
+task: '/do/1/initialize'
+resumedAt: '2024-07-26T16:59:57-05:00'
+```
+
+##### Task Retried Event
+
+The data carried by the cloud event that notifies that notifies that a task is being retried.
```
ricardozanini, line 1404:
```suggestion
The data carried by the cloud event that notifies that a task is being retried.
```

dsl-reference.md
```diff
@@ -1106,6 +1128,389 @@ Flow Directives are commands within a workflow that dictate its progression.
...
+
+```yaml
+workflow: orderPetWorkflow-ix7iryakiem8j.samples
+task: '/do/1/initialize'
+retriedAt: '2024-07-26T16:59:57-05:00'
+```
+
+##### Task Cancelled Event
+
+The data carried by the cloud event that notifies that notifies that a task has been cancelled.
```
ricardozanini, line 1424:
```suggestion
The data carried by the cloud event that notifies that a task has been canceled.
```

dsl-reference.md
```diff
@@ -1106,6 +1128,389 @@ Flow Directives are commands within a workflow that dictate its progression.
...
+
+```yaml
+workflow: orderPetWorkflow-ix7iryakiem8j.samples
+task: '/do/1/initialize'
+cancelledAt: '2024-07-26T16:59:57-05:00'
+```
+
+##### Task Faulted Event
+
+The data carried by the cloud event that notifies that notifies that a task has faulted.
```
ricardozanini, line 1444:
```suggestion
The data carried by the cloud event that notifies that a task has been faulted.
```

dsl-reference.md
```diff
@@ -1106,6 +1128,389 @@ Flow Directives are commands within a workflow that dictate its progression.
...
+faultedAt: '2024-07-26T16:59:57-05:00'
+error:
+  type: https://serverlessworkflow.io/spec/1.0.0/errors/communication
+  title: Service Not Available
+  status: 503
+```
+
+##### Task Completed Event
+
+The data carried by the cloud event that notifies that notifies that a task ran to completion.
```
ricardozanini, line 1469:
```suggestion
The data carried by the cloud event that notifies that a task ran to completion.
```

dsl-reference.md
```diff
@@ -30,6 +30,27 @@
...
+    - [Task Lifecycle Events](#task-lifecycle-events)
+      + [Task Created](#task-created-event)
+      + [Task Started](#task-started-event)
+      + [Task Suspended](#task-suspended-event)
+      + [Task Resumed](#task-resumed-event)
+      + [Task Retried](#task-retried-event)
+      + [Task Cancelled](#task-cancelled-event)
+      + [Task Faulted](#task-faulted-event)
+      + [Task Completed](#task-completed-event)
+      + [Task Status Changed](#task-status-changed-event)
```
fjtirado, line 53: Status, as I understood it, only applies to the worklow as a whole
https://github.com/serverlessworkflow/specification/blob/main/dsl.md#status-phases
I

dsl-reference.md
```diff
@@ -30,6 +30,27 @@
...
+      + [Workflow Suspended](#workflow-suspended-event)
+      + [Workflow Resumed](#workflow-resumed-event)
+      + [Workflow Correlation Started](#workflow-correlation-started-event)
+      + [Workflow Correlation Completed](#workflow-correlation-completed-event)
+      + [Workflow Cancelled](#workflow-cancelled-event)
+      + [Workflow Faulted](#workflow-faulted-event)
+      + [Workflow Completed](#workflow-completed-event)
+      + [Workflow Status Changed](#workflow-status-changed-event)
+    - [Task Lifecycle Events](#task-lifecycle-events)
+      + [Task Created](#task-created-event)
```
fjtirado, line 45: I do not see how a task can be created and not started

dsl-reference.md
```diff
@@ -1106,6 +1128,389 @@ Flow Directives are commands within a workflow that dictate its progression.
...
+completedAt: '2024-07-26T16:59:57-05:00'
+output:
+  orderId: '0a7f8581-acb9-4133-a378-0460c98ea60c'
+  petId: xt84hj202q14s
+  status: placed
+```
+
+##### Task Status Changed Event
+
+The data carried by the cloud event that notifies that notifies that the status phase of a task has changed.
```
ricardozanini, line 1494:
```suggestion
The data carried by the cloud event that notifies that the status phase of a task has changed.
```

fjtirado (commented):
I have several doubts about the concept of task life cycle. 
Its doable, buy pretty tricky and Im not sure we have to force implementors to keep track of every task status. 
Also, for some task, for example, wait, which should be the status? suspended?

dsl.md
```diff
@@ -91,21 +94,56 @@ Additionally, workflows may optionally accept inputs and produce outputs, allowi
...
-| `running` |	The workflow is currently in progress. |
-| `waiting` |	The workflow execution has been paused or halted temporarily and is waiting for something to happen. |
-| `cancelled` |	The workflow execution has been terminated before completion. |
-| `faulted` |	The workflow execution has encountered an error. |
-| `completed` |	The workflow execution has successfully finished all tasks. |
+| `pending` |	The workflow/task has been initiated and is pending execution. |
+| `running` |	The workflow/task is currently in progress. |
+| `waiting` |	The workflow/task execution is temporarily paused, awaiting either inbound event(s) or a specified time interval as defined by a [`wait`](https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#wait) task. |
+| `suspended` |	The workflow/task execution has been manually paused by a user and will remain halted until explicitly resumed. |
+| `cancelled` |	The workflow/task execution has been terminated before completion. |
```
ricardozanini, line 105:
```suggestion
| `cancelled` |	The workflow/task execution has been terminated before completion. |
```

> While both canceled and cancelled are acceptable for the past tense of cancel, the version with one L is more common in American English, and the version with two L's is more common in British English.

🤷

ricardozanini: Maybe we should use the single L?

ricardozanini (approved):

dsl-reference.md
fjtirado: If a task is a `composite` one (for, do, try) it can be of course be stopped, but at the end, it will executing another `non composite` task (basically a call), which is kind of atomic. 
I think we can avoid that conundrum by remaining silent about the whole task cycle thing and just define events at workflow level (and task started and completed)

cdavernas: > Now that I took a closer look, and after my yesterday comment related with workflow cancelation, Im not 100% sure a task can be really suspended.

Why not? Let's say you do a `wait` task, which sleeps for 10secs. You suspend it after 5secs, when resumed it still sleeps for another 5secs. Same can apply to consuming events, foreach enumerations, etc?

dsl-reference.md
cdavernas: Not in my opinion. Task status is extremely important in most, if not all workflow execution scenarios. Think of a long running flow which you represent in a UI. Just having flow events will let your users know the flow has started or ended, but whatever is in the middle will be unknown to them, potentially for hours or even days at a time.

dsl-reference.md
cdavernas: Well, a task can be instantiated, meaning initialized and provisioned with its input, and not have yet started. That's what we do in Synapse. Unlike workflows, tasks do not exist beforehand, you therefore need to tell the user about them being created: you cannot start something that has not been created.

dsl-reference.md
fjtirado: About the wait example, if we interrupt the wait (or stop listening to an event), what we should do when resuming, execute the next task after the wait or execute the wait again? I think we can avoid that question by not letting suspeding a workflow that is already waiting. 
In the case of a For, what I suggested is to let the call that the loop is currently exeucting finish (the definition of finish might vary for a task, if a call, let the call finish, if a set, let the assignment finish, if a do, hold after the current task being executed finish, and so on )

cdavernas:
> Also, for some task, for example, wait, which should be the status? suspended?

That's specified in the `dsl.md` file:

> | `waiting` |	The workflow/task execution is temporarily paused, awaiting either inbound event(s) or a specified time interval as defined by a [`wait`](https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#wait) task. |

dsl-reference.md
cdavernas: > About the wait example, if we interrupt the wait (or stop listening to an event), what we should do when resuming

I think you missed my point there: 

> Let's say you do a wait task, which sleeps for 10secs. You suspend it after 5secs, when resumed it still sleeps for another 5secs.

> In the case of a For, what I suggested is to let the call that the loop is currently executing finish

If what you mean is that the iteration should be completed before suspending, I think it's a possibility, but not my personal preference, as the iteration can be a succession of perfectly suspendable tasks.

dsl.md
cdavernas: Well, UK vs USA. I have no particular preference, so yeah, as you wish 😉

cdavernas: On a side note, maybe we should add that we should always prefer US english vs UK one in [here](https://github.com/serverlessworkflow/specification/blob/main/dsl.md#design).

cdavernas:
> I have several doubts about the concept of task life cycle.
Its doable, buy pretty tricky

I must be missing something, because I do not see the problem with it. As a matter of fact, we are already doing it in Synapse, as you can see [here](https://github.com/serverlessworkflow/synapse/blob/main/src/runner/Synapse.Runner/Services/ConnectedWorkflowExecutionContext.cs#L151), for example.

dsl-reference.md
fjtirado: Lets use an example
do
   - firstTask
   - secondTask
   - thirdTask

I guess the sequence of events will be firstTask created, first task started, ..., first task completed, second task created, second task started, ..., second task completed

so inmediatealy after firstTask instance has been created with some input, it will start execution. (obviously you have to create something before executing it, which Im discussing is the fact that a task can be created without starting it just inmediately)

fjtirado: What Im trying to say is the same we do not have an event called "setting input to task instance before actually starting it" we do not need an event "taks created before starting it"

dsl-reference.md
fjtirado: ok, then we also need to edit that section to make it clear phase status applies both for workflow and task (I implemented it just for workflows because there was not hint on that section)
Note that I also implemented task created and task completed, because I concur on the usefulness for users, which is unclert to me is that they need more task created and task completed ;)

dsl-reference.md
fjtirado: I think we should not allow suspending a workflow that is waiting (so implementors do not need that resuming logic for event/wait, which might be pretty tricky in some cases)
And for other task, I think suspending should let the operation finish (the call or the jq expression)  and then freeze, so you resume in the next task.

dsl-reference.md
cdavernas: > ok, then we also need to edit that section to make it clear phase status applies both for workflow and task

I edited it in the PR, because as you said it only applied to workflows beforehand. Have you checked the updated section in the PR? I think it's clear, but I'd be happy to update it as you see fit!

cdavernas: > is unclert to me is that they need more than task created and task completed

Well, I see what you mean, but having at least `canceled` and `faulted` seemed reasonable. Let's discuss it in today's daily!

dsl-reference.md
cdavernas: > I think we should not allow suspending a workflow that is waiting

Your opinion makes sense, it can be tricky indeed! Maybe we should let that up to the implementers, and add a couple of lines about those specifics?

dsl-reference.md
cdavernas: You make a point, those are redundant in a way! I guess I'm a bit biased because of my EDA tunnel vision 👅
 What do you guys think @JBBianchi @ricardozanini?

dsl-reference.md
JBBianchi: Resuming tasks can be challenging. Ideally, as @cdavernas suggested, allowing the runtime to directly resume a suspended task would be the best approach. However, this is often not feasible. For a passive task, such as a `wait` operation, it might work. But for active tasks, like an HTTP `call` request, the situation becomes more complex. Even if the runtime cancels the request, the server may already be processing it, and retrying the request upon resuming could lead to unintended behaviors.

As a result, there isn't a universal solution. The suspension mechanism must depend on the nature of the task. Passive tasks, such as `wait` or `listen`, can typically be suspended and resumed. Active tasks, like `call` or `run`, would need to complete before the suspension can occur, deferring it to the subsequent task.

dsl.md
fjtirado: > Well, UK vs USA. I have no particular preference, so yeah, as you wish 😉

Which is the accepted form in Irish English? ;)

fjtirado: Answering myself https://prowritingaid.com/cancelled-vs-canceled. 
Also, English is named english because is originally from England, so, in case of doubt, lets always use the original.  I vote for double L.

dsl-reference.md
ricardozanini: I'd rather have a simplified way and go with only `started`.

JBBianchi: > Lets use an example do
> 
>     * firstTask
> 
>     * secondTask
> 
>     * thirdTask
> 
> 
> I guess the sequence of events will be firstTask created, first task started, ..., first task completed, second task created, second task started, ..., second task completed
> 
> so inmediatealy after firstTask instance has been created with some input, it will start execution. (obviously you have to create something before executing it, which Im discussing is the fact that a task can be created without starting it just inmediately)

I believe this depends on the runtime implementation. It’s possible to provision tasks beforehand and start them later. For example:
- firstTask created
- secondTask created
- thirdTask created
- firstTask started
- firstTask completed
- secondTask started
- started completed

From a human perspective, creating a task is like adding it to your to-do list, while starting a task corresponds to actually executing it.

That being said, I’m not entirely convinced that having a `created` state for tasks is mandatory. One could even argue that its counterpart should also exist at the workflow level if it were the case. Additionally, I see no issue if a runtime generates more events than those listed here, as long as it still produces at least the required events.

With that in mind, perhaps the `created` state could be omitted.

fjtirado: @JBBianchi Thats it. For me the moment in which is a task instance is created is an implementation detail the user is not really interested on, specially when, as far as I know, we do not really have the concept of an schedule task and users cannot arbitrarily exeucte task out of order (the task are executed according to the workflow definition)  or do not explicitly intervene in the transitions (they might indirectly intervene by sending events for listen task or in custom ones, but there is not a task phase transition API the user might invoke. In other words, task are not created waiting for the user to start them manually. If that was the situation, the event will make perferct sense, but that being not the case, this state creates some ambiguity that I will prefer to avoid)

dsl-reference.md
fjtirado: @cdavernas @JBBianchi as dicussed in Slack, I think that if we have specific event for every phase change (both for workflow an task) we do not really need the status change one (or the other way around, if we have the status change, we do not need the specific ones)
Having only the specific ones (created, started, faulted, completed) covers all the status phase changes, so a user just interested in the fact that the status has changed, can consume them and ignore the cloud event data (thats the beauty of the CloudEvent structure, the same event can be used in different ways depending on the consumer needs)

cdavernas: The status change event is a generic event which is in my personal experience critically useful. As a matter of fact, users who want to have lifecycle notifications but that are not interested in event-specific data are usually going for those kind of events. While not necessary per say, they are an absolute plus for many users. They are usually fired after any other events, to communicate in a generic fashion with the outside world.

dsl-reference.md
cdavernas: @fjtirado The problem that I see is the data carried by the created event, which I do not want to carry over to the started one, for semantic reasons. While I do not care getting rid of the created event, I am strongly opposed to moving its payload to the started one. However, not doing so, we are forbidding the external world to know intrinsics about the task being run.

cdavernas: > In other words, task are not created waiting for the user to start them manually.

No, but as JB mentionned, you could have provisionned them beforehand, before running them in the order defined by the workflow, in a warmup procedure or whatever.

dsl-reference.md
fjtirado: I think users interested just in status change can subscribe to all event types and just check the type CE attribute to find out which phase changed. They can completely ignore (and with CE they do not need to deserialize) the data field (thats why I mention the beauty of CE)
That way we avoid publishing two different types of events for the same ocurrence (that the task changed the status from one state to another, with the current approach, we will be sending two events: the specific and the generic one, Im claiming that the specific can be consumed generically)

dsl-reference.md
ricardozanini: Can't we define the required ones in the spec and allow implementations to emit others as they find specific use cases? Later, if the proposal is compelling and the community finds it useful, we can add it to the spec.

This discussion is endless.

cdavernas: Actually, that's exactly what @fjtirado and I concluded hours ago in Slack PMs 😉

fjtirado (approved):
Lets perform the agreed changed and move forward
@cdavernas Thanks a lot!

fjtirado (approved):

PR #1067, opened 2025-01-14
Add data to EventProperties

fjtirado:

- [x] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:

**What this PR does**:

**Additional information:**

### DIFF
--- a/schema/workflow.yaml
+++ b/schema/workflow.yaml
@@ -1294,10 +1294,10 @@ $defs:
             description: An expression based event data schema.
       data:
         title: EventData
-        description: Allow filtering by data content
-        oneOf:
-          - type: string
-          - type: object 
+        description: The event's payload data
+        anyOf:
+          - $ref: '#/$defs/runtimeExpression'
+          - {}
     additionalProperties: true
   eventConsumptionStrategy:
     type: object
--- a/schema/workflow.yaml
+++ b/schema/workflow.yaml
@@ -1292,6 +1292,12 @@ $defs:
           - title: ExpressionDataSchema
             $ref: '#/$defs/runtimeExpression'
             description: An expression based event data schema.
+      data:
+        title: EventData
+        description: Allow filtering by data content
+        oneOf:
+          - type: string
+          - type: object 
     additionalProperties: true
   eventConsumptionStrategy:
     type: object
--- a/dsl-reference.md
+++ b/dsl-reference.md
@@ -1597,7 +1597,7 @@ An event object typically includes details such as the event type, source, times
 | subject | `string` | `no` | Describes the subject of the event in the context of the event producer. |
 | datacontenttype | `string` | `no` | Content type of `data` value. If omitted, it implies the `data` is a JSON value conforming to the "application/json" media type. |
 | dataschema | `string` | `no` | An URI formatted string, or [runtime expression](dsl.md#runtime-expressions), that identifies the schema that `data` adheres to. |
-| data | `object` | `no` | The event payload. |
+| data | `any` | `no` | The event payload. |
 
 *Additional properties can be supplied, see the Cloud Events specification [documentation](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#extension-context-attributes) for more info.*
 


schema/workflow.yaml
```diff
@@ -1292,6 +1292,9 @@ $defs:
           - title: ExpressionDataSchema
             $ref: '#/$defs/runtimeExpression'
             description: An expression based event data schema.
+      data:
+        type: Object
+        description: Allow filtering by data content
```
JBBianchi, line 1297: I would be consistent and add a title:

```suggestion
      data:
        type: Object
        title: EventData
        description: Allow filtering by data content 
```

cdavernas (changes requested):

schema/workflow.yaml
```diff
@@ -1292,6 +1292,12 @@ $defs:
           - title: ExpressionDataSchema
             $ref: '#/$defs/runtimeExpression'
             description: An expression based event data schema.
+      data:
+        title: EventData
+        description: The event's payload data
+        oneOf:
+          - $ref: '#/$defs/runtimeExpression'
+          - type: object
```
cdavernas, line 1300: We must add the following property, or else it won't work on some validators:

```suggestion
          - type: object 
            additionalProperties: true
```

cdavernas (changes requested):

schema/workflow.yaml
matthias-pichler, line 1300: Actually data can be anything according to the spec: https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#event-data

```suggestion
        oneOf:
          - $ref: '#/$defs/runtimeExpression'
          - {}
```

fjtirado: I think this comments overlaps with mathias one, lets dicuss it there

matthias-pichler: Which we should also update in the dsl-reference.md

fjtirado: hmm, {} its not working

fjtirado: It appears object is more inclusive than {}
@cdavernas once you put object, additional properties is implicitly true for that object

cdavernas: You should remove the `oneOf` and should not specify the `type` at all, like so:

```
data:
  title: EventData
  description: The event's payload data
```

fjtirado: @cdavernas you are right. See my latest proposal

cdavernas: > Which we should also update in the dsl-reference.md

@fjtirado Can you also update the  `dsl-reference.md` accordingly, to specify that data can be `any`?

cdavernas: > @cdavernas you are right. See my latest proposal. Its equivalent to admit anything, but we stress the possibility of using a runtime expression in the schema itself.

Great! And it's probably more explicit! Well done!

schema/workflow.yaml
```diff
@@ -1292,6 +1292,12 @@ $defs:
           - title: ExpressionDataSchema
             $ref: '#/$defs/runtimeExpression'
             description: An expression based event data schema.
+      data:
+        title: EventData
+        description: The event's payload data
+        anyOf:
+          - $ref: '#/$defs/runtimeExpression'
+          - type: {}
```
matthias-pichler: I don't think this is a valid JSON schema 🤔 

I think to allow anything in JSON schema you can either use `true` or the empty object `{}`

```suggestion
          - {} 
```

matthias-pichler (changes requested):

matthias-pichler (approved):
Not putting any restrictions on data would also work

fjtirado:
> Not putting any restrictions on data would also work

I was trying different approaches. I think 
```
        anyOf:
          - $ref: '#/$defs/runtimeExpression'
          - {}
```
Basically {} is similar to former `type:any`

PR #1072, opened 2025-01-22
[Fix #1071] Until option if any is empty

### LINKED ISSUE #1071: Any should allow empty list of filters without having to specify until clause

In case we want to subsribe to all event types and move forward, we have to write
```
listen:
        to:
          any: []
          until: '${ true }'
```
because current schema makes `until` required if any is empty. 

It will be better to just write
```
listen:
        to:
          any: []
```

Note that with current schema, until is not required if you wrote

```
listen:
        to:
          any: [eventType1, eventType2]
```

### Proposal(s):

Change the schema from 
```
  - title: AnyEventConsumptionStrategy
        oneOf:
          - properties:
              any:
                type: array
                title: AnyEventConsumptionStrategyConfiguration
                description: A list containing any of the events to consume.
                items:
                  $ref: '#/$defs/eventFilter'
                minItems: 1
              until:
                oneOf:
                  - type: string
                    title: AnyEventUntilCondition
                    description: A runtime expression condition evaluated after consuming an event and which determines whether or not to continue listening.
                  - allOf:
                      - $ref: '#/$defs/eventConsumptionStrategy'
                        title: AnyEventUntilConsumed
                        description: The strategy that defines the event(s) to consume to stop listening.
                      - properties:
                          until: false
            required: [ any ]
          - properties:
              any:
                type: array
                title: AnyEventConsumptionStrategyConfiguration
                description: A list containing any of the events to consume.
                items:
                  $ref: '#/$defs/eventFilter'
                maxItems: 0
              until:
                oneOf:
                  - type: string
                    title: AnyEventUntilCondition
                    description: A runtime expression condition evaluated after consuming an event and which determines whether or not to continue listening.
                  - allOf:
                      - $ref: '#/$defs/eventConsumptionStrategy'
                        title: AnyEventUntilConsumed
                        description: The strategy that defines the event(s) to consume to stop listening.
                      - properties:
                          until: false
            required: [ any, until ]
```
 to 
```
   - title: AnyEventConsumptionStrategy
        properties:
          any:
              type: array
              title: AnyEventConsumptionStrategyConfiguration
              description: A list containing any of the events to consume.
              items:
                $ref: '#/$defs/eventFilter'
          until:
            oneOf:
              - type: string
                title: AnyEventUntilCondition
                description: A runtime expression condition evaluated after consuming an event and which determines whether or not to continue listening.
              - allOf:
                  - $ref: '#/$defs/eventConsumptionStrategy'
                    description: The strategy that defines the event(s) to consume to stop listening.
                  - properties:
                      until: false
                title: AnyEventUntilConsumed
        required: [ any ]
```
This also has the benefit of making code generation from schema far easier  (for those SDK that prefer to take this approach) 

### Alternative(s):

_No response_

### Additional info:

_No response_

### Community Notes

fjtirado:
Fix https://github.com/serverlessworkflow/specification/issues/1071

- [ ] Specification
- [x] Schema
- [x] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:

**What this PR does**:

**Additional information:**


cdavernas (changes requested):
Can I ask you to update [this sample](https://github.com/serverlessworkflow/specification/blob/main/examples/listen-to-any-forever-foreach.yaml) by removing the now useless `until: '${ false }'` condition?

fjtirado:
@cdavernas If no until, I guess the assumption is to move forward (in other words, no until is equivalent to` until: ${true},` so the until you are requesting to remove is actually needed so is forever listening (aligned with the example file name)

cdavernas:
@fjtirado Not as far as I understand it. If you define an empty `any` and do not define `until`, the task will listen for events forever.
Unless, of course, you'd want to say that defining an empty `any` will listen to the first consumed event, whatever it is, and then keep on. If so, which is a feature that I'm obviously not opposed against, I think that we need to add a new info block to the `listen` task, explaining the result of the different strategies. Also, adding a new example listening to one single 'any' event would then be required.

fjtirado:
@cdavernas I prefer this because is aligned with the behaviour of other examples where there is at least one filter in the any array and there is no until. In those examples, once the expected event arrive, it moves forwards to next task. I think it should be the same when it is listening to all events. 
Therefore, I will add the example and the explanation.

cdavernas (changes requested):

cdavernas (approved):
LGTM! Cheers man! ❤️

PR #1077, opened 2025-02-05
Fix #1076 - Make detail and title a runtime expression in the schema

### LINKED ISSUE #1076: Error title and detail in `workflow.yaml` schema should be a Runtime Expression and not a String
### I tried this:

The current `title` and `detail` in the `Error` definition are strings. Still, they should be runtime expressions so users can manipulate the error structure when raising an error, for example.

### This happened:

The schema is not enforcing what we have in examples:

```
document:
  dsl: '1.0.0-alpha5'
  namespace: test
  name: raise-not-implemented
  version: '0.1.0'
do: 
  - notImplemented:
      raise:
        error:
          type: https://serverlessworkflow.io/errors/not-implemented
          status: 500
          title: Not Implemented
          detail: ${ "The workflow '\( $workflow.definition.document.name ):\( $workflow.definition.document.version )' is a work in progress and cannot be run yet" }
```

https://github.com/serverlessworkflow/specification/blob/v1.0.0/examples/raise-inline.yaml

### I expected this:

_No response_

### Is there a workaround?

_No response_

### Anything else?

See reference:
https://github.com/serverlessworkflow/specification/blob/v1.0.0/schema/workflow.yaml#L1190-L1197

### Area(s)

Schema

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.name: Bug Report 🐞

ricardozanini:

- [ ] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:

Fix #1076 

**What this PR does**:

**Additional information:**


schema/workflow.yaml
```diff
@@ -1188,13 +1188,21 @@ $defs:
             title: ExpressionErrorInstance
             description: An expression based error instance.
       title:
-        type: string
-        title: ErrorTitle
         description: A short, human-readable summary of the error.
+        title: ErrorTitle
+        anyOf:
```
cdavernas, line 1193: Should be `oneOf` AFAIK ;)

schema/workflow.yaml
```diff
@@ -1188,13 +1188,21 @@ $defs:
...
+          - $ref: '#/$defs/runtimeExpression'
+            title: ExpressionErrorTitle
+          - type: string
+            title: LiteralErrorTitle
       detail:
-        type: string
         title: ErrorDetails
         description: A human-readable explanation specific to this occurrence of the error.
+        anyOf:
```
cdavernas, line 1201: Same than above

cdavernas (changes requested):
Aside from two little changes, LGTM! Cheers ❤️

schema/workflow.yaml
ricardozanini: @cdavernas that was my first try, see the first commit. But the validation was failing. And once I get there: https://stackoverflow.com/a/34038277/1354502. I figured `anyOf` would be the right approach. 

Also, we have this in the schema too: https://github.com/serverlessworkflow/specification/blob/main/schema/workflow.yaml#L1220-L1230

Which is a similar context.

If you have any ideas why, we can change the type no problem.

cdavernas: @ricardozanini I guess that's because of a multi match. Second alternative also matches the first. My bad!

cdavernas (approved):
LGTM!

PR #1099, opened 2025-04-30
Improve JSON Schema compatibility

JBBianchi:
**Notes**
The schema in JSON format has been temporarily added. This allows some tools to use the JSON version as it's not accessible elsewhere otherwise.

- [ ] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:

https://github.com/serverlessworkflow/specification/discussions/1093

**What this PR does**:

- use `allOf` instead of a combination of `$ref` and `properties` for inheritance.
- bump to version 1.0.1

**Additional information:**

### DIFF
--- a/.ci/validation/src/index.ts
+++ b/.ci/validation/src/index.ts
@@ -24,7 +24,7 @@ export module SWSchemaValidator {
   const ajv = new Ajv({ strict: false, allowUnionTypes: true });
   addFormats(ajv);
 
-  const workflowSchemaId = "https://serverlessworkflow.io/schemas/1.0.0/workflow.yaml";
+  const workflowSchemaId = "https://serverlessworkflow.io/schemas/1.0.1/workflow.yaml";
   const schemaPath = "../../../schema";
   export const defaultEncoding = "utf-8";
 
--- a/.ci/validation/test/fixtures/invalid/extra-property-in-call.yaml
+++ b/.ci/validation/test/fixtures/invalid/extra-property-in-call.yaml
@@ -1,5 +1,5 @@
 document:
-  dsl: '1.0.0'
+  dsl: '1.0.1'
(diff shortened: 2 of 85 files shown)


.ci/validation/src/index.ts
```diff
@@ -24,7 +24,7 @@ export module SWSchemaValidator {
   const ajv = new Ajv({ strict: false, allowUnionTypes: true });
   addFormats(ajv);
 
-  const workflowSchemaId = "https://serverlessworkflow.io/schemas/1.0.0/workflow.yaml";
+  const workflowSchemaId = "https://serverlessworkflow.io/schemas/1.0.1/workflow.yaml";
```
ricardozanini, line 27: Are we merging this as part of the 1.0.1 release?

JBBianchi: I'm not sure what else is planned for v1.0.1, but since this introduces a schema change, I think we should bump the version. Otherwise, anyone relying on v1.0.0 might be caught off guard when this gets merged.
For example, if @fjtirado wasn’t aware of the change and we published an override of v1.0.0, it could break the Java SDK.

JBBianchi:
> @JBBianchi can we have a script to translate from YAML to JSON as part of this PR? We can create one under the `hack` or `scripts` directory. This way we keep the JSON generation from YAML idempotent.

I forgot to remove the JSON file altogether. I would rather keep only the YAML file in this repo, and build a pipeline to update the website with both YAML and JSON. WDYT ?

cdavernas (approved):
LGTM! Cheers bro!

ricardozanini (approved):

Issue #1102, opened 2025-05-01
Add a new `A2A` call task


cdavernas:
### What would you like to be added?

Add a new call task for the [**Agent-to-Agent Protocol (A2A)**](https://github.com/google/a2a), enabling secure and structured communication between agentic systems within workflows. This task would make it possible to invoke A2A-compliant endpoints and pass structured messages following the A2A specification.

The task should support:

- Discovering agents using the A2A discovery endpoint
- Sending `A2ARequest` messages to a remote agent or service,
- Receiving and handling the corresponding `A2AResponse`,
- Optional streaming for multi-part or progressive responses,
- Support for headers, metadata, and error codes defined by A2A.

This addition would enable interoperable, agent-native operations within declarative workflows and facilitate integration with a growing ecosystem of A2A-aware systems.

### Proposal(s):

Introduce a new task type, e.g. `a2a`, to handle outbound A2A messages. This task would be modeled around the A2A JSON-based protocol and support the following configuration:

- A `server` field specifying the target A2A server,
- A `method` field containing the name of the A2A request to perform
- A `parameters` field containing the request's parameters
- An optional `protocol` (version) to allow future-proofing.

On execution, the workflow runtime should:
- Format and send the message according to the A2A protocol over HTTP(S) (JSON RPC),
- Await and parse the `A2AResponse`,
- Handle errors (e.g., `error.code`, `error.message`) per A2A spec,
- Optionally process streaming responses if supported by the agent.

This proposal enables workflows to natively speak A2A, paving the way for standardized, composable agent orchestration across services.

### Community Notes

* Please vote by adding a 👍 reaction to the feature to help us prioritize.
* If you are interested to work on this feature, please leave a comment.

emmanuelallen:
@amin-nikanjam and I have been using HTTP call tasks to begin to work with A2A servers, but it leads to verbose task definitions even without comprehensive error handling. I think an A2A call task makes sense but some of the problems can be fixed with a catalog function. 

### A2A call task

This is an example of what the A2A call task could be

| Name | Type | Required | Description|
|:--|:---:|:---:|:---|
| method | `string` | `yes` | The A2A rpc method. One of `message/send`, `message/stream`, `tasks/get`, or `tasks/cancel`.  |
| agentCard | `any`\|[`externalResource`](#external-resource) | `yes` | The resource that describes the agent. |
[...]

PR #1114, opened 2025-08-11
Add new `a2a` call task to the specification and schema

emmanuelallen:

- [x] Specification
- [x] Schema
- [ ] Examples
- [ ] Extensions
- [ ] Use Cases
- [ ] Community
- [ ] CTK
- [ ] Other

**Discussion or Issue link**:

#1102 

**What this PR does**:

Adds A2A call task to the specification and schema

**Additional information:**

***Push Notifications***
A2A push notification currently only support webhooks as a way to send task update notifications. That said external services can convert those webhook invocations into cloud events that the workflow can listen for. Hopefully at some point A2A can support natively or through an extension sending out push notifications as cloud events directly.

***Streaming***
I am suggesting streaming be handled by collecting up all the objects until the stream is complete and returning them as an ordered array. In the future we can add a `foreach` argument similar to AsyncAPI that when set will run a set of tasks for each object received.

### DIFF
--- a/dsl-reference.md
+++ b/dsl-reference.md
@@ -496,7 +496,7 @@ The [A2A Call](#a2a-call) enables workflows to interact with AI agents described
 | method | `string` | `yes` | The A2A JSON-RPC method to send.<br>*Supported values are:  `message/send`, `message/stream`, `tasks/get`, `tasks/list`, `tasks/cancel`, `tasks/resubscribe`, `tasks/pushNotificationConfig/set`, `tasks/pushNotificationConfig/get`, `tasks/pushNotificationConfig/list`, `tasks/pushNotificationConfig/delete`, and `agent/getAuthenticatedExtendedCard`* |
 | agentCard | [`externalResource`](#external-resource) | `no` | The AgentCard resource that describes the agent to call.<br>*Required if `server` has not been set.* |
 | server | `string`\|[`endpoint`](#endpoint) | `no` | An URI or an object that describes the A2A server to call.<br>*Required if `agentCard` has not been set, otherwise ignored* |
-| parameters | `map` <br> `string` | `no` | The parameters for the A2A RPC method. For the `message/send` and `message/stream` methods, runtimes must set the parameters `message.messageId` and `message.role` if missing in the definition.<br>*Can be an object or a direct runtime expression.* |
+| parameters | `map` <br> `string` | `no` | The parameters for the A2A RPC method. For the `message/send` and `message/stream` methods, runtimes must default `message.messageId` to a uuid and `message.role` to `user`.<br>*Can be an object or a direct runtime expression.* |
 
 > [!NOTE]
 > The `security` and `securitySchemes` fields of the AgentCard contain authentication requirements and schemes for when communicating with the agent.
@@ -518,7 +518,8 @@ do:
       call: a2a
       with:
         method: message/send
-        agentCard: https://example.com/.well-known/agent-card.json
+        agentCard:
+          endpoint: https://example.com/.well-known/agent-card.json
         parameters:
           message:
             parts:


dsl-reference.md
```diff
@@ -483,6 +485,48 @@ do:
...
+##### A2A Call
+
+The [A2A Call](#a2a-call) enables workflows to interact with AI agents described by [A2A](https://a2a-protocol.org/).
+
+###### Properties
+
+| Name | Type | Required | Description|
+|:--|:---:|:---:|:---|
+| method | `string` | `yes` | The A2A JSON-RPC method to send. |
+| agentCard | [`externalResource`](#external-resource) | `yes` | The AgentCard resource that describes the agent to call. |
```
ricardozanini, line 497: I don't think this one should be required. I remember in our discussions @cdavernas mentioning that the vcards are not required between the a2a communication.

dsl-reference.md
```diff
@@ -483,6 +485,48 @@ do:
...
+| parameters | `any` | `no` | The parameters for the A2A rpc method. For the `message/send` and `message/stream` methods the parameters `message.messageId` and `message.role` must be automatically set if missing. |
```
ricardozanini, line 498:
```suggestion
| parameters | `any` | `no` | The parameters for the A2A RPC method. For the `message/send` and `message/stream` methods, runtimes must set the parameters `message.messageId` and `message.role` if missing in the definition. |
```

ricardozanini: Don't we want parameters in the format `map[string, any]` instead?

Are both valid?

Map:

```yaml
parameters:
     param1: value1
     param2: value2
```

Any:
```yaml
parameters: myparam
```

dsl-reference.md
```diff
@@ -483,6 +485,48 @@ do:
...
+
+> [!NOTE]
+> The `security` and `securitySchemes` fields of the AgentCard contain authentication requirements and schemes for when communicating with the agent.
+
+> [!NOTE]
+> On success the output is the JSON-RPC result. On failure runtimes must raise an error.
+
+> [!NOTE]
+> For `message/stream` and `tasks/resubscribe` methods the output is a sequentially ordered array of all the result objects.
+
```
ricardozanini, line 508: I'd avoid adding three consecutive note boxes like this. We should rewrite as regular paragraphs.

dsl-reference.md
```diff
@@ -483,6 +485,48 @@ do:
...
+###### Properties
+
+| Name | Type | Required | Description|
+|:--|:---:|:---:|:---|
+| method | `string` | `yes` | The A2A JSON-RPC method to send. |
+| agentCard | [`externalResource`](#external-resource) | `yes` | The AgentCard resource that describes the agent to call. |
+| parameters | `any` | `no` | The parameters for the A2A rpc method. For the `message/send` and `message/stream` methods the parameters `message.messageId` and `message.role` must be automatically set if missing. |
+
+> [!NOTE]
+> The `security` and `securitySchemes` fields of the AgentCard contain authentication requirements and schemes for when communicating with the agent.
```
ricardozanini, line 501: I don't see this definition in the table above. The `agentCard` is an external resource, so a file in a remote location? Won't be able to set it in the workflow definition?

dsl-reference.md
```diff
@@ -483,6 +485,48 @@ do:
...
+
+> [!NOTE]
+> On success the output is the JSON-RPC result. On failure runtimes must raise an error.
```
ricardozanini, line 504: We have error definitions. I believe we should indicate which errors.

schema/workflow.yaml
```diff
@@ -430,6 +430,36 @@ $defs:
...
+          with:
+            type: object
+            title: A2AArguments
+            description: The A2A call arguments.
+            properties:
+              agentCard:
+                $ref: '#/$defs/externalResource'
+                title: WithA2AAgentCard
+                description: The Agent Card that defines the agent to call.
+              method:
```
ricardozanini, line 452: I'm questioning myself if we need to import the a2a schema here to restrict this method as an enum.
Leaving as a string the a2a specification can grow, and we won't need to release a patch version for a single enum.

Maybe just importing their schema here and reusing the type would solve this.

schema/workflow.yaml
```diff
@@ -430,6 +430,36 @@ $defs:
...
+        properties:
+          call:
+            type: string
+            const: a2a
+          with:
+            type: object
+            title: A2AArguments
+            description: The A2A call arguments.
+            properties:
+              agentCard:
```
ricardozanini, line 448: I think this type must be defined or imported from the a2a.

ricardozanini (commented):
Many thanks!! Just a few comments so we can start the conversation.

dsl-reference.md
cdavernas: @ricardozanini it's actually needed, as was pointed out in the weekly, to resolve potential authentication.

dsl-reference.md
emmanuelallen: @ricardozanini Only the first one is valid but I would like to support directly setting parameters from a jq script the same as Set tasks. Which looking at Set task in the spec would make it `map | string`.

dsl-reference.md
cdavernas: @emmanuelallen we could put it as status code, because even if it is logically related to http status codes, it is not restricted to them.
I however think your proposal is cleaner. The title could be the code and or a high level description like you proposed, that decision is purely esthetic and should therefore be up to runtimes.

dsl-reference.md
ricardozanini: But is potential authentication required? I'm inquiring because I'm integrating with a few agents and in dev, it's not. I don't think this field should be required.

dsl-reference.md
ricardozanini: Perfect!

schema/workflow.yaml
emmanuelallen: I turned `method` into an enum.

schema/workflow.yaml
ricardozanini: Yes indeed, this discussion can be closed. The `agentCard` is better well-defined as an endpoint or expression.

dsl-reference.md
cdavernas: It shouldn't be required, indeed

schema/workflow.yaml
cdavernas: The enumerate is very lean, and it would be painless to adapt imho

dsl-reference.md
emmanuelallen: I added the server field and made the agentCard not required.

@ricardozanini The `server` field is ignored if the `agentCard` is set. I did this because it matches other parts of the spec but we could clarify it in another way.

emmanuelallen: @ricardozanini For us so far when we want to have separate environments we have been creating separate Agent Cards so haven't needed an override. Since it's just as easy to change the `agentCard` field as it is to change the `server` field I can't think of any situation where an override would be necessary.

ricardozanini:
@emmanuelallen can you take a look on the red CI?

ricardozanini:
I'll take a look later.

dsl-reference.md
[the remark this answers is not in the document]
emmanuelallen: I have changed it to say 

> For the `message/send` and `message/stream` methods, runtimes must default `message.messageId` to a uuid and `message.role` to `user`.

emmanuelallen:
@yzhao244 @emmanuelallen 

Yeah, I'm happy to add a few a2a examples as a separate PR tomorrow.

yzhao244:
@ricardozanini @cdavernas @emmanuelallen  Thanks guys for the review and valuable comments. Now we have A2A in our spec  👍

ricardozanini:
Let's finish MCP and wrap up to release 1.1.0.

PR #1161, opened 2026-06-30
Fix further dsl-reference.md / schema/workflow.yaml mismatches

### LINKED ISSUE #1156: Mismatches between dsl-reference.md and schema/workflow.yaml
**Disclaimer**: This issue was generated by Claude Code.

## Summary

While cross-checking [`dsl-reference.md`](https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md) against the authoritative JSON Schema in [`schema/workflow.yaml`](https://github.com/serverlessworkflow/specification/blob/main/schema/workflow.yaml), I found a number of divergences between the prose reference and the schema. They range from outright bugs in the schema to documented features that don't exist in the schema, plus several minor wording/typing inconsistencies.

References below are given as `dsl-reference.md:line` vs `schema/workflow.yaml:line` (line numbers as of `main`).

### 1. `evaluate` / `Evaluate` workflow property is not in the schema
- **Doc:** workflow property `evaluate` (`dsl-reference.md:110`) plus a full `Evaluate` table with `language` + `mode` (`dsl-reference.md:153-160`).
- **Schema:** root `properties` are only `document, input, use, do, timeout, output, schedule` (`workflow.yaml:6-154`). There is no `evaluate`.

### 2. Certificate authentication is documented but absent from the schema
- **Doc:** lists `certificate` as an authentication scheme (`dsl-reference.md:1711`) and has a "Certificate Authentication" section (`dsl-reference.md:1808`, body empty).
- **Schema:** `authenticationPolicy.oneOf` only contains `basic, bearer, digest, oauth2, oidc` (`workflow.yaml:1094-1219`). No `certificate`.

### 3. MCP Client — `version` is required but never defined (schema bug)
- **Doc:** MCP Client = `name` (yes) + `version` (yes) (`dsl-reference.md:2857-2860`).
- **Schema:** `McpClient` defines `name` and `description` (the `description` property is even mis-titled `McpClientVersion`), but declares `required: [ name, version ]` — `version` is not a defined property (`workflow.yaml:563-576`). So the schema requires a nonexistent property **and** exposes an undocumented `description` property.

### 4. Run → Workflow Process: required `namespace` missing from the doc table
- **Doc:** property table lists only `name`, `version`, `input` (`dsl-reference.md:1062-1066`) — no `namespace`, even though the examples use it.
- **Schema:** `SubflowConfiguration` requires `[ namespace, name, version ]` (`workflow.yaml:919-937`).

### 5. Extension `extend` enum has an extra value in the doc
- **Doc:** `call, composite, emit, extension, for, listen, raise, run, set, switch, try, wait, all` (`dsl-reference.md:2009`).
- **Schema:** `call, composite, emit, for, listen, raise, run, set, switch, try, wait, all` — no `extension` (`workflow.yaml:1587`).

### 6. MCP `protocolVersion` required flag
- **Doc:** `protocolVersion` marked required `yes` (`dsl-reference.md:539`).
- **Schema:** `with.required` is `[ method, transport ]`; `protocolVersion` is optional with default `2025-06-18` (`workflow.yaml:492-577`). Given the default, it should be `no`.

## Minor / representational

7. **OAuth2 & OIDC `authority`/`grant` required flag** — doc marks both `yes` (`dsl-reference.md:1855,1859,1920,1921`), but `oauth2AuthenticationProperties` declares no `required` (`workflow.yaml:1220-1301`).
8. **OAuth2/OIDC `issuers` type** — doc says `uri-template[]` (`dsl-reference.md:1865,1927`); schema is `string[]` (`workflow.yaml:1268-1273`).
9. **AsyncAPI Call `channel` vs `operation`** — doc marks both required `yes` (`dsl-reference.md:336-337`); schema makes them mutually-alternative via `with.oneOf` (`workflow.yaml:276-280`), so neither is unconditionally required.
10. **Catch `errors` shape** — doc types it as `errorFilter` directly (`dsl-reference.md:1249`); schema models it as an object `{ with: errorFilter }` (`workflow.yaml:1009-1015`), matching the example at `dsl-reference.md:1226`.
11. **A2A `agentCard`/`server` mutual requirement** — doc states each is required when the other is unset (`dsl-reference.md:498-499`); schema enforces no such constraint (only `required: [ method ]`, `workflow.yaml:474`).
12. **Script `language` constraint** — doc restricts to `js`/`python` (ES2024 / 3.13) (`dsl-reference.md:978,990-993`); schema has `language: type: string` with no enum (`workflow.yaml:842-845`).
13. **Duration property casing** — doc table uses `Days, Hours, Minutes, Seconds, Milliseconds` (`dsl-reference.md:2392-2396`); schema keys are lowercase `days, hours, …` (`workflow.yaml:1323-1342`). Examples use lowercase, so the table headers are misleading.
14. **Container `pullPolicy`** — doc omits the allowed values (`dsl-reference.md:934`); schema enumerates `[ ifNotPresent, always, never ]` (`workflow.yaml:826-830`).

## Also worth flagging (schema-internal)

The schema's `errorFilter` uses `details` (`workflow.yaml:1418`) while `error` and the doc use `detail` (`workflow.yaml:1391`, `dsl-reference.md:2093`).

## Suggested priority

Items **1 (`evaluate`)**, **2 (certificate auth)**, **3 (MCP client `version`)**, and **5 (extension enum)** are the substantive divergences. **3 is an outright bug in the schema** regardless of the doc.

Note that several fixes are direction-dependent (e.g. should `evaluate` and `certificate` be *added to the schema* or *removed from the doc*?). Maintainer guidance on which artifact is authoritative for each would help before opening a PR.

dimastbk:
Fix #1156

Follow-up to the mismatches reported in #1156:

Doc alignments (dsl-reference.md):
- Add the required `namespace` property to the Run -> Workflow Process table.
- Remove the spurious `extension` value from the extension `extend` enum (there is no "extension" task type).
- Mark AsyncAPI `channel`/`operation` as optional; they are mutually alternative depending on the AsyncAPI version, not both required.
- Document catch `errors.with` as the static error filter, matching the schema and examples.
- Lowercase the Duration property names (days, hours, ...) to match the schema keys.
- Document the `pullPolicy` enum values (ifNotPresent, always, never).
- Change `issuers` from `uri-template[]` to `string[]` to match the schema and RFC 7519 (StringOrURI).

Script language token (js):
- Standardize the script language token on `js` (as used by the Java SDK), replacing `javascript` in dsl-reference.md, dsl.md and the example.

Schema (schema/workflow.yaml):
- Rename the errorFilter `details` property to `detail`, matching the `error` definition and RFC 7807.
- Require `authority` and `grant` for inline OAuth2/OIDC authentication (oauth2AuthenticationProperties); `grant_type` is REQUIRED per RFC 6749 and neither field has a default.

### DIFF
--- a/dsl-reference.md
+++ b/dsl-reference.md
@@ -332,13 +332,16 @@ The [AsyncAPI Call](#asyncapi-call) enables workflows to interact with external 
 | Name | Type | Required | Description |
 |:-----|:----:|:--------:|:------------|
 | document | [`externalResource`](#external-resource) | `yes` | The AsyncAPI document that defines the [operation](https://www.asyncapi.com/docs/reference/specification/v3.0.0#operationObject) to call. |
-| channel | `string` | `yes` | The name of the channel on which to perform the operation. The operation to perform is defined by declaring either `message`, in which case the [channel](https://v2.asyncapi.com/docs/reference/specification/v2.6.0#channelItemObject)'s `publish` operation will be executed, or `subscription`, in which case the [channel](https://v2.asyncapi.com/docs/reference/specification/v2.6.0#channelItemObject)'s `subscribe` operation will be executed.<br>*Used only in case the referenced document uses AsyncAPI `v2.6.0`.*  |
-| operation | `string` | `yes` | A reference to the AsyncAPI [operation](https://www.asyncapi.com/docs/reference/specification/v3.0.0#operationObject) to call.<br>*Used only in case the referenced document uses AsyncAPI `v3.0.0`.*  |
+| channel | `string` | `no` | The name of the channel on which to perform the operation. The operation to perform is defined by declaring either `message`, in which case the [channel](https://v2.asyncapi.com/docs/reference/specification/v2.6.0#channelItemObject)'s `publish` operation will be executed, or `subscription`, in which case the [channel](https://v2.asyncapi.com/docs/reference/specification/v2.6.0#channelItemObject)'s `subscribe` operation will be executed.<br>*Used only in case the referenced document uses AsyncAPI `v2.6.0`.*  |
+| operation | `string` | `no` | A reference to the AsyncAPI [operation](https://www.asyncapi.com/docs/reference/specification/v3.0.0#operationObject) to call.<br>*Used only in case the referenced document uses AsyncAPI `v3.0.0`.*  |
 | server | [`asyncApiServer`](#asyncapi-server) | `no` | An object used to configure to the [server](https://www.asyncapi.com/docs/reference/specification/v3.0.0#serverObject) to call the specified AsyncAPI [operation](https://www.asyncapi.com/docs/reference/specification/v3.0.0#operationObject) on.<br>If not set, default to the first [server](https://www.asyncapi.com/docs/reference/specification/v3.0.0#serverObject) matching the operation's channel. |
 | protocol | `string` | `no` | The [protocol](https://www.asyncapi.com/docs/reference/specification/v3.0.0#definitionsProtocol) to use to select the target [server](https://www.asyncapi.com/docs/reference/specification/v3.0.0#serverObject). <br>Ignored if `server` has been set.<br>*Supported values are:  `amqp`, `amqp1`, `anypointmq`, `googlepubsub`, `http`, `ibmmq`, `jms`, `kafka`, `mercure`, `mqtt`, `mqtt5`, `nats`, `pulsar`, `redis`, `sns`, `solace`, `sqs`, `stomp` and `ws`* |
 | message  | [`asyncApiMessage`](#asyncapi-outbound-message) | `no` | An object used to configure the message to publish using the target operation.<br>*Required if `subscription` has not been set.* |
 | subscription | [`asyncApiSubscription`](#asyncapi-subscription) | `no` | An object used to configure the subscription to messages consumed using the target operation.<br>*Required if `message` has not been set.*  |
 | authentication | `string`<br>[`authentication`](#authentication) | `no` | The authentication policy, or the name of the authentication policy, to use when calling the AsyncAPI operation. | 
+
+> [!NOTE]
+> Exactly one of `channel` (AsyncAPI `v2.6.0`) or `operation` (AsyncAPI `v3.0.0`) **must** be set, paired with exactly one of `message` or `subscription`.
 
 ###### Examples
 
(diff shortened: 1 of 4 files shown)


ricardozanini:
Regarding this:

> Standardize the script language token on js (as used by the Java SDK), replacing javascript in dsl-reference.md, dsl.md and the example.

I don't think `js` change should be supported by the usage in the Java SDK, but by the Mozilla standard.

Unfortunately, what I have found is that `JavaScript` is a trademark owned by Oracle, and the "official" name is supposed to be ECMAScript (es). I don't think this name is as popular as javascript thou. 

Having said that, I think `js` or `javascript` is fine. The SDK should actually follow the spec, not the other way around. But I'm fine changing this to `js` for short. I don't have a strong opinion about it.

cdavernas:
@ricardozanini Same here. Whatever you guys fancy, both are ubiquitous enough anyways.

dsl-reference.md
```diff
@@ -332,14 +332,17 @@ The [AsyncAPI Call](#asyncapi-call) enables workflows to interact with external
...
 | message  | [`asyncApiMessage`](#asyncapi-outbound-message) | `no` | An object used to configure the message to publish using the target operation.<br>*Required if `subscription` has not been set.* |
 | subscription | [`asyncApiSubscription`](#asyncapi-subscription) | `no` | An object used to configure the subscription to messages consumed using the target operation.<br>*Required if `message` has not been set.*  |
 | authentication | `string`<br>[`authentication`](#authentication) | `no` | The authentication policy, or the name of the authentication policy, to use when calling the AsyncAPI operation. | 
 
+> [!NOTE]
+> Exactly one of `channel` (AsyncAPI `v2.6.0`) or `operation` (AsyncAPI `v3.0.0`) **must** be set, paired with exactly one of `message` or `subscription`.
+
```
ricardozanini, line 345: This should be enforced by the schema. Do you think you can do it? 

@fjtirado since you are familiar with this implementation, does this change resonate with you?

dimastbk: Hi. Version is not presented in the schema. But basic constraints already here: https://github.com/serverlessworkflow/specification/blob/0fa90c063f50d2e99999d6188a7904f757b469f5/schema/workflow.yaml#L293-L297

ricardozanini (approved):

ricardozanini:
@cdavernas can you take a look?

cdavernas (approved):
LGTM! Cheers ❤️

PR #1168, opened 2026-07-23
[Fix #1167] Clarify event id is auto-generated by runtime

### LINKED ISSUE #1167: Event id should not be required in emit.event.with — runtime should auto-generate it
## Issue

The [event properties](https://github.com/open-workflow-specification/specification/blob/main/dsl-reference.md#event-properties) table says:

> `id` — *Required when emitting an event using `emit.event.with`.*

But `id` must be **unique per event emission**. Requiring it in the DSL forces users to either hardcode a static value (defeating uniqueness) or write a runtime expression. In practice, runtimes auto-generate a UUID at emission time — the sdk-java reference implementation does exactly this.

The `id` should be **guaranteed to exist at emission time** (by the runtime), not **required in the DSL definition**.

## Proposal

Clarify in the spec text that `id` is auto-generated by the runtime when not provided, and remove the "required when emitting" constraint. The `source` and `type` requirements for emit are correct and should stay.

ricardozanini:
## Summary

- The DSL reference text stated `id` is required when emitting events via `emit.event.with`
- The schema already only requires `[ source, type ]` — `id` was never enforced
- Runtimes (e.g., sdk-java) auto-generate a UUID at emission time
- Update the text to clarify this behavior instead of implying a DSL-level requirement

Fixes #1167


cdavernas (approved):
LGTM! Cheers ❤️

PR #1171, opened 2026-07-23
[Fix #1162] Add security note against evaluating expressions in user input

### LINKED ISSUE #1162: Add a note to the specification for implementations to NOT evaluate `jq` expressions on user's input
Make it explicitly that `jq` expressions MUST be evaluated ONLY on the workflows definitions and NOT on user's input.

Having user's to submit workflow data to a workflow runtime that contains `jq` expressions on it, open a huge security breach on their systems. Regular Expressions, for instance, can be used to exploit system data and resources.

Workflows should evaluate `jq` expressions ONLY on the definitions itself, not on user's input.

ricardozanini:
## Summary

- Adds a caution note in the Runtime Expressions section of `dsl.md` making it explicit that implementations **must not** evaluate runtime expressions contained in user-provided input data
- This prevents injection attacks where user input containing expressions (e.g., jq, regex) could be executed against system data and resources

Fixes #1162


cdavernas (approved):
LGTM! Cheets ❤️

PR #1172, opened 2026-07-23
[Fix #1151] Replace output with outputUri in lifecycle completed events

### LINKED ISSUE #1151: Consider removing `output` from Workflow Completed Lifecycle event
## Background
The current [Serverless Workflow DSL specification for Lifecycle Events](https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#lifecycle-events) defines an `output` attribute for both **Workflow Completed** and **Task Completed** events. This attribute is intended to carry the resulting data payload of the workflow or task execution.

## The Problem
The [CloudEvents specification (Size Limits section)](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#size-limits) explicitly states:
> *"Intermediaries MUST forward events of a size of 64 KiB or less."*

Workflow and task outputs in production environments frequently exceed this 64KB threshold. Because lifecycle events act as the control plane for workflow orchestration, embedding large data payloads (the data plane) directly into these events introduces a significant risk. Generic event intermediaries, brokers, or API gateways routing these CloudEvents are well within their rights to drop or truncate events exceeding 64KB. This would lead to silent failures and broken integrations for downstream consumers waiting for state transitions.

## Proposed Solution
To ensure strict interoperability with the CloudEvents spec and prevent systemic routing failures, I propose we remove the raw `output` attribute from the lifecycle events.

Instead of embedding the payload, we should recommend the **Claim Check pattern**. 

**Suggested Changes:**
1. **Remove**: The `output` (map) property from the `Workflow Completed Event` and `Task Completed Event`.
2. **Introduce**: An optional `outputReference` or `outputUri` (URI) property (same for `input`-related events). This attribute would provide a standard URI pointing to the workflow engine's storage where the full output/input payload can be securely fetched by interested consumers.

## Alternatives Considered
* **Opt-in embedding**: Allow users to explicitly configure whether the output is embedded via a workflow configuration flag, accompanied by strong documentation warnings about the 64KB limit.
* **Truncated Summary**: Introduce a strictly size-capped `outputSummary` for lightweight status codes, while still removing the bulk data field.

## Impact
This is a breaking change to the DSL schema. It will require updates to runtime implementations to ensure they emit the reference URI instead of the raw map. However, it is a necessary adjustment to guarantee architectural stability and adherence to standard eventing patterns.

## References
* [Serverless Workflow - Lifecycle Events Spec](https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#lifecycle-events)
* [CloudEvents - Size Limits Spec](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#size-limits)

ricardozanini:
## Summary

- Replace `output` (`map`) with `outputUri` (`uri`) in both **Workflow Completed** and **Task Completed** lifecycle events
- Adopts the [Claim Check](https://www.enterpriseintegrationpatterns.com/patterns/messaging/StoreInLibrary.html) pattern: instead of embedding potentially large payloads, events carry a URI pointing to where the output can be fetched
- This prevents exceeding the [CloudEvents 64KB size limit](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#size-limits) which could cause intermediaries to drop or truncate events

### Breaking change

This removes the `output` property from completed lifecycle events. Runtime implementations must emit `outputUri` instead.

Fixes #1151


cdavernas (approved):
LGTM! Cheers ❤️

fjtirado:
I know I late but this one, as innocent as it might look, has important design implications. 
It assumes that the implementer has means to store the output and expose it though an URI (so it basically make imposible to have  embedded spec implementations) and for small outputs it implies killing flies with guns (store all ouput result and make them avaialble externally, forcing all listeners to write retrieveal ouput code)
On the other hand, the problem is real, we might not be able to include the output in the CloudEvent if the output is too large. 
I think we need to compromise here. In my opinion output should be a complex structure. When the output Map fits into the CE, then it should be sent.  When the output map does not fit into the CE, then implementors should either send a truncated map or an URI (if the information makes sense to be make an URI). Therefore we neeed a flag that indicated if the output is the whole raw info as it is (because it fits the model), it is an URI or it is a truncated one. Listeners will be aware of that structure and act in consequence (yes, the listener will be complex, but with the merged change already is and actually if it is up to the listener writer to retrieve the URI for large outputs, while, with the merged change it has to write the code anyway even for retrieving small outputs)

cdavernas:
@fjtirado While I agree with your observations and reservations, I believe the primary intent of these events is to communicate lifecycle information, not to carry task outputs.

Consequently, I do not think we should introduce an output structure supporting inline or truncated payloads. We should retain the URI reference, as it provides a consistent, payload-agnostic way to locate additional information when available, without requiring the event itself to model task output data.

This is especially appropriate given that, as @ricardozanini pointed out, the specification currently defines neither workflow input data nor task input/output data.

cdavernas:
@geomagilles Care to join the discussion? Given that Lemline is, as far as I know, a fully event-driven, event-native workflow engine, I’d be interested to hear how you approached this.

fjtirado:
If URI is just a way to describe how to manually find out which is the output data of the workflow/task (without really forcing implementors to provide a way to downalod it thorugh such uri), for example, a hint of how to query a DB where that data might have been stored for historical purposes, Im perfectly fine with that. Initially I understood the URI as somethign which CloudEvent consumers can copy paste in your browser or in Postman (or your preferred client to invoke Rest APIS) and download the output data (and that I consider too heavyweight requirement)

geomagilles:
@cdavernas sorry for joining lately. Lemline is indeed event-driven, but use CE exports for observability only, not for actual processing. That being said, I agree with @fjtirado 's initial comment: imposing an URI for accessing outputs seems overkill. After reading @fjtirado's last comment above, i'm a bit puzzled - I'd probably appreciate an example to better understand how to use an URI here. In particular how to make sure the interpretation is not implementation specific?

ricardozanini:
@geomagilles ideally, your lifecycle event will emit an `outputURI` that is a pointer where users can find the output value. As per the spec, we **can't impose** a physical location on implementations, but what we do **recommend** is a pointer where users can find it.

For instance, you can emit `outputURI: /<namespace>/<name>/<version>/<task json path>`. This way, users and systems know the origin and can create a complete URL later on to find it. The **how**, though, should be per implementation. 

This value is per runtime implementation, since it may vary. Nevertheless, the structure must conform so users will expect that any implementation would emit lifecycle events. 

The way each implementation/runtime provides the platform for querying the results, or the URI itself, is up to them. It won't interfere with cross-compatibility.

geomagilles:
I'm a bit confused- I thought the promise of cross-compatibility was that I could change an engine by another. This is not the case if the engine has the choice of the outputURI structure. Why not adding the data as explicit parameters then?

cdavernas:
@geomagilles Why would `outputUri` break that promise?

In addition, we can't add the whole data because, as said by @ricardozanini, that would potentially break the CE's recommandation about payload size. 

Also, big payloads would break on most transports (ex: max 2MB over NATS, for example). 

Finally, those events are intended to communicate lifecycle info, not data.

ricardozanini:
@geomagilles please have a look: https://learn.microsoft.com/en-us/azure/architecture/patterns/claim-check

The URI can be anything designed by the system/runtime. Users can then migrate their workflows from one platform to another. In the new platform, the lifecycle events will target the new system/domain. This is not a cross-compatibility problem, AFAIU.

PR #1173, opened 2026-07-23
[Fix #1148] Add then flow directive to catch clause

### LINKED ISSUE #1148: feat: Add `then` directive in `catch` clause for `try` tasks.

Support for the `then` directive (or a similar control-flow transition) directly within the `catch` clause of a `try` task in the Serverless Workflow DSL.

### Proposal(s):

I propose adding a `then` property to the `catch` object definition. When an execution error is caught by matching the errors criteria, the workflow should immediately transition to the task specified by the `then` directive, bypassing the default sequential flow.

Here is an example of how this would look in the DSL:

```yaml
document:
  dsl: '1.0.3'
  namespace: test
  name: try-example
  version: '0.1.0'
do:
  - trySomething:
      try:
        - invalidHttpCall:
            call: http
            with:
              method: get
              endpoint: https://
      catch:
        errors:
          with:
            type: https://serverlessworkflow.io/dsl/errors/types/communication
            status: 503
        then: sharedErrorHandlerTask #  <-- Transitions to a reusable task defined elsewhere
```

### Alternative(s):

1. **Using a subsequent switch task**: Catching the error, mapping it to a state variable, and following the `try` task with a `switch` statement to evaluate the error variable and branch accordingly. However, this is much more verbose, spreads the error-handling logic across multiple tasks, and hurts readability.

2. **Using the do directive within catch**: We can currently define tasks inline using `do`. However, if multiple tasks across the workflow need to trigger the same compensation or error-handling sequence, using `do` forces us to duplicate that logic inline every single time, violating DRY principles.

### Additional info:

The primary motivation for this addition is task reusability and control flow flexibility.

Currently, if we want to handle an `error`, we are forced to define the reaction inline (via do). However, in complex workflows or Saga patterns, it is common for multiple different tasks to fail in ways that require the exact same compensation sequence (e.g., a shared `rollbackTransaction` or `notifyAdmin` task).

By introducing the `then` directive into the `catch` clause, we can transition the control flow to a `task` that is called by another control flow or sits in the default control flow of the same workflow. This allows multiple `try` tasks to cleanly route to a single, shared error-handling task, significantly reducing duplication and making the workflow much easier to maintain.

### Community Notes

ricardozanini:
## Summary

- Add `then` property (type [`flowDirective`](#flow-directive)) to the `catch` clause of `try` tasks
- Enables control flow transitions after catching errors (e.g., `then: handleError` to jump to a shared error-handling task)
- Avoids duplicating inline `do` blocks when multiple `try` tasks need the same error-handling sequence

### Changes

- `schema/workflow.yaml`: Added `then` property referencing `flowDirective` to `TryTaskCatch`
- `dsl-reference.md`: Added `then` row to the Catch properties table
- `examples/try-catch-then-directive.yaml`: New example demonstrating the `then` directive in a catch clause

### Test plan

- [x] All 130 validation tests pass (including the new example)

Fixes #1148


examples/try-catch-then-directive.yaml
```diff
@@ -0,0 +1,26 @@
...
+        errors:
+          with:
+            type: https://serverlessworkflow.io/dsl/errors/types/communication
+            status: 503
+        then: handleError
+  - continueProcessing:
+      set:
+        status: ok
+      then: end
+  - handleError:
```
cdavernas: I think `handleError` is a misleading name, as the handling should be performed in the `do`, the `then` is what happens **after**, regardless of whether an error was handled.

cdavernas (changes requested):

examples/try-catch-then-directive.yaml
ricardozanini: Good point! Renamed `handleError` to `recordFailure` — the `do` block handles the error, this task just records the outcome after catch processing.

ricardozanini:
@cdavernas can you have another look please?

cdavernas (approved):

PR #1175, opened 2026-07-24
[Fix #1174] Event source auto-generated by runtime implementations

### LINKED ISSUE #1174: Event source should be auto-generated by runtime implementations
### Problem

The `source` property on emitted events is currently documented as:

> An URI formatted string, or runtime expression, that identifies the context in which an event happened. `source` + `id` is unique for each distinct event.
> *Required when emitting an event using `emit.event.with`.*

This forces workflow authors to always provide a `source` value (typically as a runtime expression) when emitting events, even though the runtime already has all the information needed to construct a meaningful source URI from the workflow's identity.

### Proposal

Apply the same rationale as #1167 / #1168 (event `id` auto-generation): make `source` optional and document that runtime implementations **must** generate it at emission time when not explicitly provided.

A natural auto-generated value would be a URI derived from the workflow identity, e.g.:

```
/{namespace}/{name}/{version}
```

For example, a workflow with `namespace: payments`, `name: order-processing`, `version: 1.0.0` would produce:

```
source: /payments/order-processing/1.0.0
```

When the author explicitly sets `source`, the provided value takes precedence.

### Changes needed

- Update the `source` description in `dsl-reference.md` to clarify it is auto-generated by runtime implementations (e.g., from the workflow identity) at emission time, similar to the `id` field
- Keep `source` as not required in the schema (it already is optional)
- Update the "Required when emitting" note to reflect that runtimes provide a default

### Benefits

- Reduces boilerplate in workflow definitions
- Consistent with the `id` auto-generation pattern (#1168)
- The runtime-generated value is more reliable than expecting authors to keep source URIs in sync with workflow identity

### Community Notes

ricardozanini:
## Summary

- Clarify that runtime implementations generate the event `source` at emission time (e.g., from the workflow identity `/{namespace}/{name}/{version}`) when not explicitly provided
- When the workflow author supplies a value, it takes precedence over the auto-generated one
- Consistent with the `id` auto-generation pattern from #1168

Fixes #1174

## Test plan

- [x] Verify `dsl-reference.md` event properties table reflects auto-generation and override behavior
- [x] Verify `schema/workflow.yaml` EventSource description is updated


cdavernas (approved):
LGTM! Cheers ❤️

PR #1178, opened 2026-07-29
[Fix #1177] Add read property to schedule.on

### LINKED ISSUE #1177: Schedule.on does not allow selecting between Cloud data or envelope

Unlike listen task, which allows workflow definition writer to choose between the whole cloud event or just the data property, `schedule.on` remains silent. 
This forces the runtime implementation to use envelope, so user can filter it in the workflow input filter. 

### Proposal(s):

The proposal is to add to the schema the possibility to chose between envelope and data for schedule.on (basically copying the already existing [listen](https://github.com/open-workflow-specification/specification/blob/main/schema/workflow.yaml#L733-L738) section, or, better extracting it to a shared schema object)

### Alternative(s):

Force users to always put a dummy filter if they are interested on data

### Additional info:

_No response_

### Community Notes

ricardozanini:
## Summary

- Adds a `read` property to the `schedule` definition as a sibling of `on`
- Allows users to choose how consumed events are read when a workflow is triggered by events (`data`, `envelope`, or `raw`)
- Defaults to `data`, making this a non-breaking change
- Mirrors the existing `listen.read` property for consistency

### Example

```yaml
schedule:
  on:
    one:
      with:
        type: com.example.order-placed
  read: envelope
```

Fixes #1177


dsl-reference.md
```diff
@@ -148,6 +148,7 @@ Configures the schedule of a workflow.
 | cron | `string` | `no` | Specifies the schedule using a CRON expression, e.g., '0 0 * * *' for daily at midnight.<br>*Required when no other property has been set.* |
 | after | [`duration`](#duration) | `no` | Specifies a delay duration that the workflow must wait before starting again after it completes. In other words, when this workflow completes, it should run again after the specified amount of time.<br>*Required when no other property has been set.*  |
 | on | [`eventConsumptionStrategy`](#event-consumption-strategy) | `no` | Specifies the events that trigger the workflow execution.<br>*Required when no other property has been set.* |
+| read | `string` | `no` | Specifies how [events](https://cloudevents.io/) are read when the workflow is triggered by events.<br>*Supported values are:*<br>*- `data`: Reads the [event's](https://cloudevents.io/) data.*<br>*- `envelope`: Reads the [event's](https://cloudevents.io/) envelope, including its [context attributes](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#context-attributes).*<br>*- `raw`: Reads the [event's](https://cloudevents.io/) raw data.*<br>*Defaults to `data`.* |
```
cdavernas: Maybe worth saying that it can only be set when `on` is set, otherwise MUST NOT be set, otherwise runtime must consider the workflow invalid

cdavernas (changes requested):
LGTM! Cheers ❤️

cdavernas (approved):
LGTM!

PR #1187, opened 2026-08-11
[Fix #1186] Allow any type in for.in inline array

### LINKED ISSUE #1186: Using any as type of hard coded array for For task
This is a follow up of https://github.com/open-workflow-specification/specification/issues/1163

Probably the type of the array should be any rather than object, to allow arrays of numbers and string, without having to wrap it into object

Lets illsutrate with an example

This is the workflow definition using a expression string 
[Workflow definition usign expression](https://github.com/open-workflow-specification/sdk-java/blob/main/impl/test/src/test/resources/workflows-samples/for-sum.yaml)

Lets suppose that the workflow input is 
```
{
 "input": [1,2,3]
}
```
It is not possible to genreate an equivalent workflow using hard coded array if the array type is object, the closer we might is this one

[Workflow definition using hardcoded array ](https://github.com/open-workflow-specification/sdk-java/pull/1614/changes#diff-88cb74ab544cda98cb4bac056cb5b663ed010352385c8ffcbdf2b0624852185bR1-R22)

ricardozanini:
## Summary
- Removes the `type: object` constraint from inline array items in `for.in`, allowing iteration over arrays of any type (strings, numbers, booleans, objects, etc.)
- Updates the DSL reference to reflect the type change from `array<object>` to `array`
- Adds a new example (`for-inline-primitive-array.yaml`) demonstrating iteration over an inline array of numbers

Fixes #1186

## Test plan
- [x] All 185 schema validation tests pass
- [ ] Verify the new example validates correctly in CI

### DIFF
--- a/dsl-reference.md
+++ b/dsl-reference.md
@@ -689,7 +689,7 @@ Allows workflows to iterate over a collection of items, executing a defined set 
 | Name | Type | Required | Description|
 |:--|:---:|:---:|:---|
 | for.each | `string` | `no` | The name of the variable used to store the current item being enumerated.<br>Defaults to `item`. |
-| for.in | `string` \| `array<object>` | `yes` | A [runtime expression](dsl.md#runtime-expressions) or an inline array of objects used to get the collection to enumerate. |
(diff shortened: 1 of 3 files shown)


cdavernas (approved):
LGTM! Cheers ❤️

